PR #53, opened 2023-12-20
iox2-51 shm long name support

elfenpiff:
## Notes for Reviewer
Do not review before #52 is merged, since it is based on it.

MacOS has the restrictions that only 31 characters are allowed for the shared memory name but iceoryx2 requires much more. The trick is that a file is created to:
* be able to list all shared memorys in mac os (already present)
* store the actual shm name, the file name corresponds to the user given shared memory name

In the file is then the triple stored (`$PID$_$TIMESTAMP_SECS$_$TIMESTAMP_NSEC$`). In the worst case scenario where PID + SEC is larger than 1.000.000.000 it would hit the max name length of mac os (32 > 31), but to our luck NSEC is always smaller than 1.000.000.000 giving us the extra character we need.

1. [x] Relevant issues are linked
1. [x] Assign PR to reviewer

## References

Use either 'Closes #123' or 'Relates to #123' to reference the corresponding issue.

Closes #**ISSUE-NUMBER**

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -269,19 +269,4 @@ macos_aarch64_stable_debug_task:
     - rustup component add clippy rustfmt
   set_toolchain_script: rustup default stable
   <<: *IOX2_CARGO_FMT_AND_CLIPPY
-  <<: *IOX2_COMMON_BUILD_DEBUG
-  test_script:
-    - cargo test -p iceoryx2-pal-concurrency-sync
-                 -p iceoryx2-pal-posix
-                 -p iceoryx2-bb-container
-                 -p iceoryx2-bb-elementary
-                 -p iceoryx2-bb-lock-free
-                 -p iceoryx2-bb-log
-                 -p iceoryx2-bb-memory
-                 -p iceoryx2-bb-posix
-                 -p iceoryx2-bb-system-types
-                 -p iceoryx2-bb-testing
-                 -p iceoryx2-bb-threadsafe
-                 -p iceoryx2-bb-trait-tests
-                 -p iceoryx2-cal
-                 --lib --bins --tests --no-fail-fast
+  <<: *IOX2_COMMON_BUILD_AND_TEST_DEBUG
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -278,6 +278,9 @@ macos_aarch64_stable_debug_task:
                  -p iceoryx2-bb-lock-free
                  -p iceoryx2-bb-log
                  -p iceoryx2-bb-memory
+                 -p iceoryx2-bb-posix
                  -p iceoryx2-bb-system-types
                  -p iceoryx2-bb-testing
+                 -p iceoryx2-bb-threadsafe
+                 -p iceoryx2-bb-trait-tests
                  --lib --bins --tests --no-fail-fast
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -283,4 +283,5 @@ macos_aarch64_stable_debug_task:
                  -p iceoryx2-bb-testing
                  -p iceoryx2-bb-threadsafe
                  -p iceoryx2-bb-trait-tests
+                 -p iceoryx2-cal
                  --lib --bins --tests --no-fail-fast
--- a/doc/release-notes/iceoryx2-unreleased.md
+++ b/doc/release-notes/iceoryx2-unreleased.md
@@ -1,28 +1,28 @@
 # iceoryx2 v?.?.?
 
-## [vx.x.x](https://github.com/larry-robotics/iceoryx2/tree/vx.x.x) (xxxx-xx-xx) <!--NOLINT remove this when tag is set-->
+## [vx.x.x](https://github.com/eclipse-iceoryx/iceoryx2/tree/vx.x.x) (xxxx-xx-xx) <!--NOLINT remove this when tag is set-->
 
-[Full Changelog](https://github.com/larry-robotics/iceoryx2/compare/vx.x.x...vx.x.x) <!--NOLINT remove this when tag is set-->
+[Full Changelog](https://github.com/eclipse-iceoryx/iceoryx2/compare/vx.x.x...vx.x.x) <!--NOLINT remove this when tag is set-->
 
 ### Features
 
- * Example [#1](https://github.com/larry-robotics/iceoryx2/issues/1)
+ * MacOS Platform support [#51](https://github.com/eclipse-iceoryx/iceoryx2/issues/51)
 
 ### Bugfixes
 
- * Example [#1](https://github.com/larry-robotics/iceoryx2/issues/1)
+ * Example [#1](https://github.com/eclipse-iceoryx/iceoryx2/issues/1)
 
 ### Refactoring
 
- * Example [#1](https://github.com/larry-robotics/iceoryx2/issues/1)
+ * Example [#1](https://github.com/eclipse-iceoryx/iceoryx2/issues/1)
 
 ### Workflow
 
- * Example [#1](https://github.com/larry-robotics/iceoryx2/issues/1)
+ * Example [#1](https://github.com/eclipse-iceoryx/iceoryx2/issues/1)
 
 ### New API features
 
- * Example [#1](https://github.com/larry-robotics/iceoryx2/issues/1)
+ * Example [#1](https://github.com/eclipse-iceoryx/iceoryx2/issues/1)
 
 ### API Breaking Changes
 
--- a/iceoryx2-pal/posix/src/macos/mman.rs
+++ b/iceoryx2-pal/posix/src/macos/mman.rs
@@ -141,6 +141,10 @@ pub unsafe fn shm_open(name: *const char, oflag: int, mode: mode_t) -> int {
     let real_name = match real_name {
         Some(name) => name,
         None => {
+            if oflag & O_CREAT == 0 {
+                Errno::set(Errno::ENOENT);
+                return -1;
+            }
             let real_name = generate_real_shm_name();
             if !write_real_shm_name(name, &real_name) {
                 return -1;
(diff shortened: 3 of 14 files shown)


doc/release-notes/iceoryx2-unreleased.md
```diff
@@ -1,28 +1,28 @@
 # iceoryx2 v?.?.?
 
-## [vx.x.x](https://github.com/larry-robotics/iceoryx2/tree/vx.x.x) (xxxx-xx-xx) <!--NOLINT remove this when tag is set-->
+## [vx.x.x](https://github.com/eclipse-iceoryx/iceoryx2/tree/vx.x.x) (xxxx-xx-xx) <!--NOLINT remove this when tag is set-->
```
elBoberido, line 3: Hahaha, we need a link checker :)

iceoryx2/tests/service_publish_subscribe_tests.rs
```diff
@@ -864,7 +864,7 @@ mod service_publish_subscribe {
         const BUFFER_SIZE: usize = 1;
         const HISTORY_SIZE: usize = 0;
         const MAX_BORROW: usize = 1;
-        const MAX_SUBSCRIBERS: usize = 100;
+        const MAX_SUBSCRIBERS: usize = 50;
```
elBoberido, line 867: I'm thinking whether it would make sense to keep the old value on Linux as canary check that something is broken?

.cirrus.yml
```diff
@@ -278,6 +278,10 @@ macos_aarch64_stable_debug_task:
                  -p iceoryx2-bb-lock-free
                  -p iceoryx2-bb-log
                  -p iceoryx2-bb-memory
+                 -p iceoryx2-bb-posix
                  -p iceoryx2-bb-system-types
                  -p iceoryx2-bb-testing
+                 -p iceoryx2-bb-threadsafe
+                 -p iceoryx2-bb-trait-tests
+                 -p iceoryx2-cal
                  --lib --bins --tests --no-fail-fast
```
elBoberido, line 287: If all tests are running, these lines can be removed and `IOX2_COMMON_BUILD_DEBUG` can be replaced by `IOX2_COMMON_BUILD_AND_TEST_DEBUG`

iceoryx2-pal/posix/src/macos/mman.rs
```diff
@@ -74,23 +78,94 @@ unsafe fn shm_file_path(name: *const char, suffix: &[u8]) -> [u8; MAX_PATH_LENGT
...
+    if read(shm_state_fd, buffer.as_mut_ptr().cast(), SHM_MAX_NAME_LEN) <= 0 {
+        close(shm_state_fd);
+        return None;
+    }
+
+    close(shm_state_fd);
+    Some(buffer)
+}
+
+unsafe fn write_real_shm_name(name: *const char, buffer: &[u8]) -> bool {
```
elBoberido, line 99: This `char` is not the `std::char` but an alias to a `u8`, right? I would urge to rename it to `c_char` or something similar since it is quite confusing when reading the code.

iceoryx2-pal/posix/src/macos/mman.rs
```diff
@@ -74,23 +78,94 @@ unsafe fn shm_file_path(name: *const char, suffix: &[u8]) -> [u8; MAX_PATH_LENGT
...
+
+    let mut buffer = [0u8; SHM_MAX_NAME_LEN];
+
+    if read(shm_state_fd, buffer.as_mut_ptr().cast(), SHM_MAX_NAME_LEN) <= 0 {
+        close(shm_state_fd);
+        return None;
+    }
+
+    close(shm_state_fd);
+    Some(buffer)
```
elBoberido, line 96: Does the buffer need to have a null-termination? I mean, is it possible that read write a non-null character at the last index of the buffer?

iceoryx2-pal/posix/src/macos/mman.rs
```diff
@@ -74,23 +78,94 @@ unsafe fn shm_file_path(name: *const char, suffix: &[u8]) -> [u8; MAX_PATH_LENGT
     state_file_path
 }
 
+unsafe fn get_real_shm_name(name: *const char) -> Option<[u8; SHM_MAX_NAME_LEN]> {
+    let shm_file_path = shm_file_path(name, SHM_STATE_SUFFIX);
+    let shm_state_fd = open_with_mode(shm_file_path.as_ptr().cast(), O_RDONLY, 0);
```
elBoberido, line 83: Just to get it right. It's possible to creates files with more than 33 characters in /dev/shm, e.g. text files but it is not possible to create a shared memory file with more than 33 characters, right?

iceoryx2-pal/posix/src/macos/mman.rs
```diff
@@ -74,23 +78,94 @@ unsafe fn shm_file_path(name: *const char, suffix: &[u8]) -> [u8; MAX_PATH_LENGT
...
-    }
+    };
 
-    ret_val
+    crate::internal::shm_open(real_name.as_ptr().cast(), oflag, mode as uint)
 }
 
 pub unsafe fn shm_unlink(name: *const char) -> int {
+    let real_name = get_real_shm_name(name);
     remove(shm_file_path(name, SHM_STATE_SUFFIX).as_ptr().cast());
```
elBoberido, line 161: Wouldn't it be better to delete this file only after the shm is closed. Just in case there is an error on closing the shm and one needs to retry?

iceoryx2-pal/posix/src/macos/mman.rs
elfenpiff: Correct, but the limit is 31. The buffer here has a size of 33 since this is the theoretical maximum that will be never reached.

Here the link to the shm name length: https://stackoverflow.com/questions/38049068/osx-shm-open-returns-enametoolong

iceoryx2-pal/posix/src/macos/mman.rs
elfenpiff: No, since a shared memory is persistent and it is useful to keep this file until the shared memory is completely removed otherwise we may leak shared memory since we no longer know their names.
There is also no possibility in mac os to list all shared memories.

iceoryx2/tests/service_publish_subscribe_tests.rs
elfenpiff: At this level I would like to have the same tests, behavior, limits etc. and say with confidence that every platform supports at least 50 subscribers.

I completely understand your rational, but when we go this route we may have a wild mixture of verified upper limits for every platform. I would rather go the path that we say 50 subscribers are supported for every platform, but an individual higher limit for a specific platform is maybe working.

Maybe we can introduce platform specific system limit tests where exactly those differences can be verified. What do you think?

iceoryx2-pal/posix/src/macos/mman.rs
elfenpiff: nope `char` is an alias to `core::ffi::c_char` but I get your point, I will create an issue

elfenpiff: See: https://github.com/eclipse-iceoryx/iceoryx2/issues/54

iceoryx2-pal/posix/src/macos/mman.rs
elfenpiff: Yes, since it is used as input argument for `shm_open` which requires a null terminated c string.

The theoretical maximum length of a generated shm name is 32 (but can never be reached), with null terminator it results in `SHM_MAX_NAME_LEN=33`.

I took your note into consideration and read only SHM_MAX_NAME_LEN - 1 characters in read so that the null character is always preserved at the end.

iceoryx2/tests/service_publish_subscribe_tests.rs
elBoberido: Indeed. These limits might be a criteria for tier 2 and for tier 1 we could have tests for additional limits.

iceoryx2-pal/posix/src/macos/mman.rs
elBoberido: Oh, then in theory 3 `u32` + 2 `_` can result in 32 characters. Practically this will never happen since you use only 9 characters for the nanoseconds ... and the PID and seconds part should also not become that large.

iceoryx2-pal/posix/src/macos/mman.rs
elBoberido: Maybe I got this wrong. I assumed that this line deletes the file with the real shm name.

iceoryx2-pal/posix/src/macos/mman.rs
elfenpiff: Pid and seconds can be at most 2^32 which is equivalent to 10 ascii characters + 9 ascii characters for nanoseconds + 2 `_`. So 31 characters exactly.

iceoryx2-pal/posix/src/macos/mman.rs
elfenpiff: No it deletes the real shm name. `shm_unlink` removes the shared memory so the file must be deleted. If you have still opened the shm you are not affected since the OS does an ref count with the fds.

elBoberido: Maybe I still get it wrong. My concern is that this file gets deleted and then `shm_unlink` does not work and the shm is left in the filesystem without a possibility to remove it.

elBoberido (approved):

iceoryx2-pal/posix/src/macos/mman.rs
elfenpiff: You are completely right, this would be a problem. I adjusted the behavior, now the state file is only removed when either `shm_unlink` was successful or it failed with the errno `ENOENT` signaling that it does not exist.

elBoberido (approved):

Issue #96, opened 2024-01-17
Create decentralized service artifacts cleanup API


elfenpiff:
## (Code) Example Of Cumbersome API

Whenever an application crashes or is killed with `SIGKILL` all the resources are left behind. We need a decentral functionality to detect stale resources and clean them up.

So whenever an application is restarted  - typically during development - and detects old resources from a previous run, it automatically removes them.

PR #97, opened 2024-01-21
[#96] decentral service artifact cleanup

elfenpiff:
## Notes for Reviewer

* Cleaned up a technical debt, `FilePath::path()` and `FilePath::file_name()` are returning a `Path` or `FileName` and no longer a slice. Makes it easier to use since it prevents either addition checks or unsafe sections. In some files, those unsafe sections were removed.
* Introduces `process_state.rs` in posix, a basic building block to enable process monitoring. When this is merged, a concept in `iceoryx2_cal` called `Monitor` will be introduced that will use this and provides a safe user API.

1. [x] Relevant issues are linked
1. [x] Assign PR to reviewer

## References

Use either 'Closes #123' or 'Relates to #123' to reference the corresponding issue.

Relates to #96


doc/release-notes/iceoryx2-unreleased.md
```diff
@@ -8,21 +8,19 @@
...
  <!-- NOTE: Add new entries sorted by issue number to minimize the possibility of conflicts when merging. -->
 
- * Example text [#1](https://github.com/eclipse-iceoryx/iceoryx2/issues/1)
+ * Introduce `iceoryx2-bb-posix::process_state` for process monitoring [#96](https://github.com/eclipse-iceoryx/iceoryx2/issues/96)
 
 ### Bugfixes
 
  <!-- NOTE: Add new entries sorted by issue number to minimize the possibility of conflicts when merging. -->
 
- * Fix `clock_nanosleep` on macOS [#80](https://github.com/eclipse-iceoryx/iceoryx2/issues/80)
```
elfenpiff, line 17: Forgot to remove those. Already part of the previous release 0.2.2

iceoryx2-bb/system-types/src/file_path.rs
```diff
@@ -127,19 +128,29 @@ impl FilePath {
...
+    pub fn path(&self) -> Path {
+        // SAFETY
+        // * the file path ensures that is a valid path to a file, therefore the first part
+        //   must be a valid path
+        unsafe {
+            Path::new_unchecked(match self.rfind(PATH_SEPARATOR) {
+                None => core::slice::from_raw_parts(self.as_ptr(), 0),
+                Some(0) => self.as_bytes().get(0..1).unwrap(),
+                Some(pos) => self.as_bytes().get(0..pos).unwrap(),
+            })
```
elBoberido, line 153: Similarly, this could be changed to
```rs
let path = match self
    .as_bytes()
    .rsplitn(2, |c| *c == PATH_SEPARATOR)
    .skip(1)
    .next()
{
    Some(p) if p.len() > 0 => p,
    Some(_) => &[PATH_SEPARATOR],
    None => &[],
};

unsafe { Path::new_unchecked(path) }
```

iceoryx2-bb/container/src/semantic_string.rs
```diff
@@ -82,6 +82,14 @@ pub trait SemanticString<const CAPACITY: usize>:
         Ok(new_self)
     }
 
+    /// Creates a new name but does not verify that it does not contain invalid characters.
+    ///
+    /// # Safety
+    ///
+    ///   * The user must ensure that the bytes contain only valid characters.
+    ///
+    unsafe fn new_unchecked(bytes: &[u8]) -> Self;
```
elBoberido, line 91: I guess the user must also ensure that the slice is not too large?

iceoryx2-bb/container/src/semantic_string.rs
```diff
@@ -154,6 +162,22 @@ pub trait SemanticString<const CAPACITY: usize>:
...
 
+    /// Adds bytes to the string without checking if they only contain valid characters or
+    /// would result in a valid result.
+    ///
+    /// # Safety
+    ///
+    ///   * The user must ensure that the bytes contain only valid characters.
+    ///   * The user must ensure that the result, after the bytes were added, is valid.
+    ///
+    unsafe fn insert_bytes_unchecked(&mut self, idx: usize, bytes: &[u8]);
```
elBoberido, line 173: Same here, the size of the slice must not be too large, right?

elBoberido (commented):
The `process_state.rs` and its tests are something for tomorrow

iceoryx2-bb/system-types/src/path.rs
```diff
@@ -63,18 +64,25 @@ semantic_string! {
...
+          if i + 1 == this.value.len() && this.value[i] == PATH_SEPARATOR {
+              break;
+          }
+
+          if !(previous_char_is_path_separator && this.value[i] == PATH_SEPARATOR) {
+              raw_path[n] = this.value[i];
+              n += 1;
+          }
+
+          previous_char_is_path_separator = this.value[i] == PATH_SEPARATOR
```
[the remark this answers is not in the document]
elfenpiff: @elBoberido 

```rust
            // TODO: can the PATH_SEPARATOR size be larger than 1?
            raw_path[n] = PATH_SEPARATOR;
```

Yes it can, for windows we need to escape the path separator and it becomes two characters `\\`

iceoryx2-bb/system-types/tests/path_tests.rs
```diff
@@ -72,6 +72,24 @@ mod windows {
...
+
+        let sut = Path::new(b"0:\\a\\b").unwrap();
+        assert_that!(sut.is_absolute(), eq false);
+
+        let sut = Path::new(b"").unwrap();
+        assert_that!(sut.is_absolute(), eq false);
+
+        let sut = Path::new(b"what/ever/").unwrap();
+        assert_that!(sut.is_absolute(), eq false);
+    }
```
elBoberido, line 92: It gets more complicated https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats#path-normalization

There are also UNC paths but those are related to network paths. Then there is `\\.\` and `\\?\`. And since that is not fun enough, there is also `\foo\bar` which is relative to the root of the current drive

iceoryx2-bb/system-types/src/file_path.rs
```diff
@@ -114,33 +115,30 @@ impl FilePath {
...
     /// Returns the [`Path`] part of the [`FilePath`].
-    pub fn path(&self) -> &[u8] {
-        match self.rfind(PATH_SEPARATOR) {
-            None => unsafe { std::slice::from_raw_parts(self.as_ptr(), 0) },
-            Some(0) => self.as_bytes().get(0..1).unwrap(),
-            Some(pos) => self.as_bytes().get(0..pos).unwrap(),
-        }
+    pub fn path(&self) -> Path {
+        let path = match self.as_bytes().rsplitn(2, |c| *c == PATH_SEPARATOR).nth(1) {
+            Some(p) if !p.is_empty() => p,
```
elBoberido, line 134: And now, just read this line out loud :grin:

iceoryx2-bb/system-types/tests/path_tests.rs
elfenpiff: There is also a big list of more or less obscure windows reserved file and directory names: https://help.interfaceware.com/v6/windows-reserved-file-names

This is not the full list, and the list varies of course from windows to windows version.

When I read this, I decided that making `path` on windows safe in the sense that it always contains a legal non-reserved file/path name is nearly impossible. So it is best effort to cover the 99%. For the remaining stuff we have the low level errors, panics and a useful detailed log output.

elBoberido: How about just using `std::path::Path` on Windows? We probably don't need Windows to be no-std

elfenpiff: Yes this makes sense, and we maybe shouldn't stop at `std::path::Path` and also reuse other std constructs. But I would keep this for another PR and create a good-first-issue for that.

elBoberido (commented):
I need to create a sequence diagram with all the things that could happen during a race between the application and the monitoring process. I somehow have the feeling that it is quite racy and especially when you start multiple applications at once, including the monitoring application, this could lead to unexpected results.

iceoryx2-bb/posix/src/process_state.rs
```diff
@@ -0,0 +1,595 @@
...
+//!     // fraction of time
+//!     ProcessState::InInitialization => (),
+//!
+//!     // Process died, we have to inform other interested parties and maybe cleanup some
+//!     // resources
+//!     ProcessState::Dead => {
+//!         // monitored process crashed, perform cleanup
+//!
+//!         // after cleanup is complete, we remove the process state file to signal that
+//!         // everything is clean again
```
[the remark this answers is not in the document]
elfenpiff: Good point, in my head this was ready but not in code ;) ...

So, the only one who is allowed to remove resources that are owned by a process is the one who holds the lock.

When the process is still running, the lock is owned by the process and remove will fail. If two concurrent remove calls are made, both try to lock the file first, the one who wins can remove it.

iceoryx2-bb/posix/src/process_state.rs
```diff
@@ -0,0 +1,595 @@
...
+    ///
+    /// // start monitoring from this point on
+    /// let guard = ProcessGuard::new(&process_state_path).expect("");
+    /// ```
+    pub fn new(path: &FilePath) -> Result<Self, ProcessGuardCreateError> {
+        let origin = "ProcessGuard::new()";
+        let msg = format!(
+            "Unable to create new ProcessGuard with the file \"{}\"",
+            path
+        );
```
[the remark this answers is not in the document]
elfenpiff: I am also a bit cautious to use this but I am pragmatic when I know that those constructs are not used in hot loops.

But it would be nice to have a performance degradation test somewhere in the pipeline.

iceoryx2-bb/posix/src/process_state.rs
```diff
@@ -0,0 +1,808 @@
...
+        let lock_state = fail!(from self, when Self::get_lock_state(file),
+                            "{} since the lock state of the state file could not be acquired.", msg);
+
+        match lock_state as _ {
+            posix::F_WRLCK => Ok(ProcessState::Alive),
+            _ => match File::does_exist(&self.path) {
+                Ok(true) => match file.permission() {
+                    Ok(INIT_PERMISSION) => Ok(ProcessState::Starting),
+                    Err(_) | Ok(_) => {
+                        self.file.set(None);
```
elBoberido, line 653: Is this correct? Setting it to `None` although no cleanup has taken place? So the one who calls `state` has to do the cleanup immediately and cannot rely on calling `state` later on to get the same result?

elBoberido (commented):
An interim review. Need to finish `process_state_tests.rs`

iceoryx2-bb/posix/tests/process_state_tests.rs
```diff
@@ -0,0 +1,302 @@
...
+        ProcessCleanerCreateError::ProcessIsStillAlive
+    );
+}
+
+#[test]
+#[cfg(not(any(target_os = "linux", target_os = "freebsd", target_os = "macos")))]
+pub fn process_state_cleaner_cannot_be_acquired_twice() {
+    let path = generate_file_path();
+    let mut cleaner_path = path.clone();
+    cleaner_path.push_bytes(b"_cleanup").unwrap();
```
elBoberido, line 279: We need to implement the error trait for all errors, then we can use `?` in the tests.

elBoberido (commented):
Tests look good.

I'm just not sure if the terminology for the file names should be changed to something more generic than `.cleaner`, e.g. `.owner_lock`. It acts more or less as the lock file in iceoryx1.

Nothing to block this PR but something to think about.

iceoryx2-bb/posix/src/process_state.rs
```diff
@@ -0,0 +1,808 @@
...
+//!
+//! // remove leftovers from a previous abnormal process termination
+//! //
+//! // # Safety
+//! //  * This process owns the state file, therefore it is safe to remove a leftover from a
+//! //    previous run.
+//! unsafe { ProcessGuard::remove(&process_state_path).expect("") };
+//!
+//! // start monitoring from this point on
+//! let guard = ProcessGuard::new(&process_state_path).expect("");
```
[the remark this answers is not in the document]
elfenpiff: This I explicitly do not want. When the file still exists, it means someone must take care of the stale resources first before calling remove.

If one would have the ability to ignore that and just remove the file, they would violate the contract.

Removing this file is only allowed when the whole self-healing system crashed - with all of its processes. Then we need to clean up everything. Otherwise, one should never call remove.

I will adjust the example and make this more clear.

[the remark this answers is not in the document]
elBoberido: You are right. I totally overlooked that. This is just a building block for the actual functionality

iceoryx2-bb/posix/src/process_state.rs
```diff
@@ -0,0 +1,808 @@
...
+                }
+            },
+        };
+
+        match file.set_permission(FINAL_PERMISSION) {
+            Ok(_) => {
+                trace!(from "ProcessGuard::new()", "create process state \"{}\" for monitoring", path);
+                Ok(Self {
+                    file,
+                    _cleaner_file,
```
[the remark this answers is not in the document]
elfenpiff: Yes, but then I have to write more in this line, see `_cleaner_file: cleaner_file` since it is no where used in this object.

iceoryx2-bb/posix/src/process_state.rs
elfenpiff: It works. The `file` is required to check the lock state. If the lock state indicates that it is dead, we proactively close the file and reopen it again with every new run.

If we wouldn't do it, a `ProcessCleaner` would remove all resources but the monitor would still hold the old file descriptor pointing to a file that no longer exists in the file system.

We have no other choice than to close it and reopen it with every run to not run into this problem.

iceoryx2-bb/posix/tests/process_state_tests.rs
elfenpiff: Yes, but I would keep it for later. When the repo is too clean right from the beginning, people are maybe intimidated by our perfectness ;)

elBoberido (approved):

Issue #132, opened 2024-02-21
Starting Subscriber leads to `ExceedsMaxBorrowedSamples`


elfenpiff:
## Required information

**Operating system:**
Linux 6.7.3-arch1-2

**Rust version:**
rustc 1.76.0 (07dca489a 2024-02-04)

**Cargo version:**
rustc 1.76.0 (07dca489a 2024-02-04)

**iceoryx2 version:**
`main` branch

**Observed result or behaviour:**
When a subscriber starts, it thinks it can only borrow 0 samples and therefore emits an error when receiving new samples.

**Expected result or behaviour:**
The amount of borrowed samples is parsed correctly.

**Conditions where it occurred / Performed steps:**
```
       11 [T] SharedMemory { name: FileName { value: FixedSizeByteString<255> { len: 60, data: "iox2_18018_1427533032182014834768991936413080.
              publisher_data" } }, size: 4621245, base_address: 0x757f6076e000, has_ownership: false, file_descriptor: FileDescriptor { value:
               6, is_owned: true }, memory_lock: None }
              | open
       12 [D] Receiver { shared_memory: SharedMemory { name: FileName { value: FixedSizeByteString<255> { len: 85, data: "iox2_142753303218201
              4834768991936413080_1432365950095384959362232936474399.connection" } }, size: 190, base_address: 0x757f60e12000, has_ownership: 
              false, file_descriptor: FileDescriptor { value: 3, is_owned: true }, memory_lock: None }, borrow_counter: UnsafeCell { .. }, nam
              e: FileName { value: FixedSizeByteString<255> { len: 69, data: "1427533032182014834768991936413080_14323659500953849593622329364
              74399" } } }
              | Unable to receive another sample since already 0 samples were borrowed and this would exceed the max borrow value of 0.
       13 [D] Subscriber { dynamic_subscriber_handle: ContainerHandle { index: 0, container_id: 0 }, publisher_connections: PublisherConnectio
              ns { connections: [UnsafeCell { .. }, UnsafeCell { .. }], subscriber_id: UniqueSubscriberId(UniqueSystemId { value: 143236595009
              5384959362232936474399 }), config: Config { global: Global { root_path_unix: "/tmp/iceoryx2/", root_path_windows: "c:\\Temp\\ice
              oryx2\\", prefix: "iox2_", service: Service { directory: "services", publisher_data_segment_suffix: ".publisher_data", static_co
[...]

Issue #131, opened 2024-02-21
Publisher retrieves invalid sample when subscriber is connected


elfenpiff:
## Required information

**Operating system:**
Linux 6.7.3-arch1-2

**Rust version:**
rustc 1.76.0 (07dca489a 2024-02-04)

**Cargo version:**
rustc 1.76.0 (07dca489a 2024-02-04)

**iceoryx2 version:**
`main` branch

**Fix:**
On publisher side, for the offset calculations when using a pool allocator, we had the implicit requirement that the offsets start at zero and are a multiple of the bucket size. Introducing for the normal allocators a method that returns the smallest start point of the payload and for the shm allocators a method that returns the relative payload starting offset and use this in the SharedMemory concept as the starting point fixed the problem.

**Observed result or behaviour:**
When a new subscriber connects, the publisher suddenly retrieves an invalid sample.

**Expected result or behaviour:**
No invalid samples are retrieved

**Conditions where it occurred / Performed steps:**
```
       61 [T] SharedMemory { name: FileName { value: FixedSizeByteString<255> { len: 85, data: "iox2_1527439745112502164476023069140769_155754
              6446867922612761728747364181.connection" } }, size: 190, base_address: 0x794f63baa000, has_ownership: true, file_descriptor: Fil
              eDescriptor { value: 7, is_owned: true }, memory_lock: None }
              | create
       62 [D] PoolAllocator { buckets: UniqueIndexSet { data_ptr: RelocatablePointer { distance: 112, _phantom: PhantomData<core::cell::Unsafe
              Cell<u32>> }, capacity: 36, borrowed_indices: 21, head: 90359339105, is_memory_initialized: true }, bucket_size: 128360, bucket_
              alignment: 8, start: 133381875593496, size: 4620967, is_memory_initialized: true }
              | Tried to release memory (133381875593494) which does not belong to this allocator.
       63 [D] PoolAllocator { allocator: PoolAllocator { buckets: UniqueIndexSet { data_ptr: RelocatablePointer { distance: 112, _phantom: Pha
              ntomData<core::cell::UnsafeCell<u32>> }, capacity: 36, borrowed_indices: 21, head: 90359339105, is_memory_initialized: true }, b
              ucket_size: 128360, bucket_alignment: 8, start: 133381875593496, size: 4620967, is_memory_initialized: true }, base_address: 133
              381875593494, max_supported_alignment_by_memory: 4096 }
              | Failed to release shared memory chunk
[...]

Issue #130, opened 2024-02-21
Stop/Start subscriber leads to OutOfMemory on publisher side


elfenpiff:
## Required information

**Operating system:**
Linux 6.7.3-arch1-2

**Rust version:**
rustc 1.76.0 (07dca489a 2024-02-04)

**Cargo version:**
rustc 1.76.0 (07dca489a 2024-02-04)

**iceoryx2 version:**
`main` branch

**Fix:**
The problem had two causes:
 * The publisher did not recycle samples that were stored in the receive channel, only the retrieve channel was recycled
 * When the subscriber dies and has loaned samples, all those samples leak

The `UsedChunkList` from iceoryx1 was ported and whenever a connection is terminated all samples from the `UsedChunkList` will be recycled.

**Observed result or behaviour:**
When a subscriber is stopped and reconnected, it can lead to a OutOfMemory error on the publisher side.

**Expected result or behaviour:**
No out of memory.

**Conditions where it occurred / Performed steps:**
```
       96 [D] SharedMemoryBuilder { name: FileName { value: FixedSizeByteString<255> { len: 85, data: "iox2_1439258800234125956732939963044004
              _1512227937909763411657611488120570.connection" } }, size: 190, is_memory_locked: false, has_ownership: true, permission: OWNER_
              READ | OWNER_WRITE | OWNER_EXEC | OWNER_ALL, creation_mode: Some(OpenOrCreate), zero_memory: true, access_mode: ReadWrite, enfor
              ce_base_address: None }
              | Unable to create shared memory since it already exists.
       97 [T] SharedMemory { name: FileName { value: FixedSizeByteString<255> { len: 85, data: "iox2_1439258800234125956732939963044004_151222
              7937909763411657611488120570.connection" } }, size: 190, base_address: 0x79024f6b1000, has_ownership: true, file_descriptor: Fil
              eDescriptor { value: 6, is_owned: true }, memory_lock: None }
              | create
       98 [D] PoolAllocator { buckets: UniqueIndexSet { data_ptr: RelocatablePointer { distance: 112, _phantom: PhantomData<core::cell::Unsafe
              Cell<u32>> }, capacity: 36, borrowed_indices: 36, head: 157183357176, is_memory_initialized: true }, bucket_size: 128360, bucket
              _alignment: 8, start: 133050822349080, size: 4620967, is_memory_initialized: true }
              | No more buckets available to allocate 128360 bytes with an alignment of 8.
[...]

Issue #123, opened 2024-02-19
`iceoryx2-cal::hash` Introduce uuid type and make it convertable to `FileName`


elfenpiff:
## Brief feature description

All over the code the `uuid` is used as basis for a files and therefore for the `FileName`. It is always safe to convert the uuid to a file name, nevertheless the user has to write `unsafe {}` blocks with a safety statement that it is safe. A better approach would be to bake this all in, into its own uuid type.

PR #134, opened 2024-02-23
[#129] fix missing connections

### LINKED ISSUE #129: Publisher/Subscriber fail to connect
## Required information

**Operating system:**
Linux 6.7.3-arch1-2

**Rust version:**
rustc 1.76.0 (07dca489a 2024-02-04)

**Cargo version:**
rustc 1.76.0 (07dca489a 2024-02-04)

**iceoryx2 version:**
`main` branch

When either a publisher or subscriber is already running, and a new counterpart is started (subscriber/publisher) they sometimes fail to connect. This can be fixed when another publisher or subscriber is connecting - it seems that the ports are sometimes not informed about new participants.

Connecting ports always works

Start a single publisher and start stop subscriber multiple times with some time in between.

elfenpiff:
## Notes for Reviewer

The UsedChunkList exploits that the publisher/zero copy connection combination is using a pool allocator with equi-distant offset (buckets of same size). Therefore, we introduced number of samples and sample size.
We translate the memory addresses into indices (based on the above information) and set atomic bools to indicate if a sample is in use or not.

1. [x] Assign PR to reviewer

## References

Relates to  #116 
Closes #129 #130 #131 #132

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -274,16 +274,17 @@ windows_server_2019_x64_stable_release_task:
 ###   set_toolchain_script: rustup default 1.70.0
 ###   <<: *IOX2_COMMON_BUILD_AND_TEST_DEBUG
 
-freebsd_x64_stable_debug_task:
-  depends_on: preflight_check
-  <<: *IOX2_CONTAINER_FREEBSD_X64
-  env:
-    PATH: /root/.cargo/bin:$PATH
-    HOME: /root  # must be set manually to '/root' or 'rustup' will throw an error
-  <<: *IOX2_FREEBSD_SETUP
-  set_toolchain_script: rustup default stable
-  <<: *IOX2_CARGO_FMT_AND_CLIPPY
-  <<: *IOX2_COMMON_BUILD_AND_TEST_DEBUG
+### TODO iox2-140 uncomment when FreeBSD is fixed
+### freebsd_x64_stable_debug_task:
+###   depends_on: preflight_check
+###   <<: *IOX2_CONTAINER_FREEBSD_X64
+###   env:
+###     PATH: /root/.cargo/bin:$PATH
+###     HOME: /root  # must be set manually to '/root' or 'rustup' will throw an error
+###   <<: *IOX2_FREEBSD_SETUP
+###   set_toolchain_script: rustup default stable
+###   <<: *IOX2_CARGO_FMT_AND_CLIPPY
+###   <<: *IOX2_COMMON_BUILD_AND_TEST_DEBUG
 
 #
 # macOS
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -148,20 +148,22 @@ ubuntu_22_04_x64_stable_debug_task:
   <<: *IOX2_COMMON_BUILD_AND_TEST_DEBUG
   linux_only_doc_test_script: cargo test --doc -- --ignored
 
-### TODO commented out due to limited CI time
-### ubuntu_22_04_x64_beta_debug_task:
-###   depends_on: ubuntu_22_04_x64_stable_debug
-###   <<: *IOX2_CONTAINER_UBUNTU_22_04_X64
-###   set_toolchain_script: rustup default beta
-###   <<: *IOX2_COMMON_BUILD_AND_TEST_DEBUG
+ubuntu_22_04_x64_beta_debug_task:
+  # TODO commented out due to limited CI time
+  only_if: false
+  depends_on: ubuntu_22_04_x64_stable_debug
+  <<: *IOX2_CONTAINER_UBUNTU_22_04_X64
+  set_toolchain_script: rustup default beta
+  <<: *IOX2_COMMON_BUILD_AND_TEST_DEBUG
 
-### TODO commented out due to limited CI time
-### ubuntu_22_04_x64_nightly_debug_task:
-###   depends_on: ubuntu_22_04_x64_beta_debug
-###   allow_failures: true
-###   <<: *IOX2_CONTAINER_UBUNTU_22_04_X64
-###   set_toolchain_script: rustup default nightly
-###   <<: *IOX2_COMMON_BUILD_AND_TEST_DEBUG
+ubuntu_22_04_x64_nightly_debug_task:
+  # TODO commented out due to limited CI time
+  only_if: false
+  depends_on: ubuntu_22_04_x64_beta_debug
+  allow_failures: true
+  <<: *IOX2_CONTAINER_UBUNTU_22_04_X64
+  set_toolchain_script: rustup default nightly
+  <<: *IOX2_COMMON_BUILD_AND_TEST_DEBUG
 
 # Pipeline 3
 
@@ -177,12 +179,13 @@ ubuntu_22_04_x64_stable_release_task:
 
 # Pipeline 4
 
-### TODO commented out due to limited CI time
-### ubuntu_22_04_aarch64_min_version_debug_task:
-###   depends_on: preflight_check
-###   <<: *IOX2_CONTAINER_UBUNTU_22_04_AARCH64
-###   set_toolchain_script: rustup default 1.70.0
-###   <<: *IOX2_COMMON_BUILD_AND_TEST_DEBUG
+ubuntu_22_04_aarch64_min_version_debug_task:
+  # TODO commented out due to limited CI time
+  only_if: false
+  depends_on: preflight_check
+  <<: *IOX2_CONTAINER_UBUNTU_22_04_AARCH64
+  set_toolchain_script: rustup default 1.70.0
+  <<: *IOX2_COMMON_BUILD_AND_TEST_DEBUG
 
 ubuntu_22_04_aarch64_stable_debug_task:
   depends_on: preflight_check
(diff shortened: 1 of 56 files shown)


.cirrus.yml
```diff
@@ -19,7 +19,7 @@
 # Constants
 
 iox2_task_timeout_template: &IOX2_TASK_TIMEOUT
-  timeout_in: 30m  # This needs to be reasonable large in order not to run in a timeout in the docker image re-builds
+  timeout_in: 45m  # This needs to be reasonable large in order not to run in a timeout in the docker image re-builds
```
elfenpiff, line 23: It seems that FreeBSD 13.2 is no longer supported by CirrusCI

elBoberido: Yes, we need to switch to FreeBSD 14.0

iceoryx2-bb/memory/tests/one_chunk_allocator_tests.rs
```diff
@@ -175,22 +175,21 @@ fn one_chunk_allocator_shrink_fails_when_alignment_increases() {
...
-        },
-        is_err
-    );
+    unsafe {
+        sut.shrink(
+            NonNull::new(1234 as *mut u8).unwrap(),
+            Layout::from_size_align_unchecked(CHUNK_SIZE, CHUNK_ALIGNMENT),
+            Layout::from_size_align_unchecked(CHUNK_SIZE / 2, CHUNK_ALIGNMENT),
+        )
+        .unwrap();
```
elBoberido, line 191: It is unclear whether the panic shall come from `shrink` or `unwrap`. You could use `let _ = sut.shrink(...` to clarify the intention better.

I'm positive surprised that this also works for release builds although a `debug_assert` is used.

iceoryx2/src/port/notifier.rs
```diff
@@ -165,27 +167,37 @@ impl<Service: service::Service> Notifier<Service> {
...
             dynamic_storage,
-            dynamic_notifier_handle,
+            dynamic_notifier_handle: None,
         };
 
         if let Err(e) = new_self.populate_listener_channels() {
             warn!(from new_self, "The new Notifier port is unable to connect to every Listener port, caused by {:?}.", e);
         }
 
+        std::sync::atomic::compiler_fence(Ordering::SeqCst);
```
elBoberido, line 182: Interesting. After reading the documentation, a few question marks remain. Especially since the example should be broken on a weakly ordered CPU

elBoberido (commented):
9 files to go. Looks good so far.

iceoryx2-cal/src/zero_copy_connection/used_chunk_list.rs
```diff
@@ -0,0 +1,205 @@
...
+
+            for i in 0..self.capacity {
+                unsafe {
+                    (self.data_ptr.as_ptr() as *mut AtomicBool)
+                        .add(i)
+                        .write(AtomicBool::new(false))
+                };
+            }
+
+            self.is_memory_initialized.store(true, Ordering::Relaxed);
```
elBoberido, line 99: Can you document why relaxed is here sufficient? Is there a way the member could be read from another thread?

iceoryx2-cal/src/zero_copy_connection/used_chunk_list.rs
```diff
@@ -0,0 +1,205 @@
...
+
+    use iceoryx2_bb_elementary::owning_pointer::OwningPointer;
+
+    use super::*;
+
+    #[derive(Debug)]
+    #[repr(C)]
+    pub struct UsedChunkList<PointerType: PointerTrait<AtomicBool>> {
+        data_ptr: PointerType,
+        capacity: usize,
```
elBoberido, line 42: Food for thought. I guess `capacity` will never change after initialization. A type only with a `new` function and `Deref` would bake this invariant into the type system.

iceoryx2-cal/src/zero_copy_connection/used_chunk_list.rs
```diff
@@ -0,0 +1,205 @@
...
+            self.set(value, false)
+        }
+
+        pub fn pop(&self) -> Option<usize> {
+            self.verify_init("pop");
+
+            (0..self.capacity).find(|&i| unsafe {
+                (*self.data_ptr.as_ptr().add(i)).swap(false, Ordering::Relaxed)
+            })
+        }
```
elBoberido, line 160: Okay, this is quite expensive when the list is large and the indices at the end are in use. I guess this is only used on cleanup?

iceoryx2-cal/tests/used_chunk_list_tests.rs
```diff
@@ -0,0 +1,69 @@
...
+            assert_that!(sut.remove(i), eq true);
+            assert_that!(sut.remove(i), eq false);
+
+            assert_that!(sut.insert(i), eq true);
+        }
+
+        for i in 0..sut.capacity() {
+            assert_that!(sut.remove(sut.capacity() - i - 1), eq true);
+            assert_that!(sut.remove(sut.capacity() - i - 1), eq false);
+        }
```
elBoberido, line 53:
```suggestion
        for i in (0..sut.capacity()).rev() {
            assert_that!(sut.remove(i), eq true);
            assert_that!(sut.remove(i), eq false);
        }
```

elBoberido (commented):
Only the `mpmc::Container` left but thats something for tomorrow

iceoryx2-bb/memory/tests/one_chunk_allocator_tests.rs
elfenpiff: I added `RUSTFLAGS="-C debug-assertions"` to our release builds to be able to test those things.

iceoryx2-cal/src/zero_copy_connection/used_chunk_list.rs
elfenpiff: Nope. The RelocatableContainer is not allowed to be used, send, synced, shared with any other thread when it is not initialized - meaning `init` was not yet called.

The RelocatableContainer has a two-step initialization approach
1. Call `new_uninit`
2. Call `init`

instead of a simple `new` call. Therefore, there is no need for syncing since the container is at this point in time only accessible from one thread.

iceoryx2-cal/src/zero_copy_connection/used_chunk_list.rs
elfenpiff: Yes you are right but I will not fix this in this PR I created an issue and would tackle this in a follow up PR. (#142 )

iceoryx2-cal/src/zero_copy_connection/used_chunk_list.rs
elfenpiff: Exactly, this is only used when a subscriber goes out of scope, terminates the connection and the publisher has to recollect all samples.
I explicitly did not return a `Vec` here to avoid using the heap. The type-state pattern is also out of the question, since this must stick to the same memory position.

But I will improve it a bit by storing the last pop position and reuse it in the next call. When it reached capacity it will start with zero again.

elBoberido: No need to complicate things

iceoryx2-bb/memory/tests/one_chunk_allocator_tests.rs
elBoberido: Hmm, doesn't this counteract the optimizations you did with debug assert?

iceoryx2-cal/src/zero_copy_connection/used_chunk_list.rs
elfenpiff: Actually, it makes things easier and more efficient. I will adjust this.

iceoryx2-bb/memory/tests/one_chunk_allocator_tests.rs
elBoberido: Okay, it is only in the CI but with that one has to keep in mind to also use it for local builds. What do you think about using `#[cfg(debug_assertions)]` on the related tests?

elfenpiff: Yes, but it is only for the CI, see `.cirrus.yml`.

elBoberido: That's what I mean. When I run these tests locally in release mode they will fail and I'm pretty sure I will have forgotten about the flag in a few weeks.

iceoryx2-bb/lock-free/src/mpmc/container.rs
```diff
@@ -217,20 +227,20 @@ impl<T: Copy + Debug> RelocatableContainer for Container<T> {
...
-            + std::mem::align_of::<u32>()
-            - 1
-            + std::mem::align_of::<T>()
-            - 1
+        // UniqueIndexSet
+        (std::mem::size_of::<u32>() * (capacity + 1) + std::mem::align_of::<u32>() - 1)
+        //  ActiveIndexPtr
+        + (std::mem::size_of::<AtomicU64>() * capacity + std::mem::align_of::<u64>() - 1)
+        // data ptr
+        + (std::mem::size_of::<T>() * capacity + std::mem::align_of::<T>() - 1)
```
elBoberido, line 243: Can this be move to a common helper function?

elBoberido (approved):

Issue #153, opened 2024-03-11
Use `DynamicStorage` concept as basis for every shared memory based concept


elfenpiff:
## Brief feature description

The `DynamicStorage` is the most basic building block for shared memory based concepts. The `ZeroCopyConnection`, `CommunicationChannel`, `SharedMemory` all the same approach in creation/destruction which is already abstracted in the `DynamicStorage`.

Therefore, the code should be adjusted so that those concepts no longer use the underlying posix shared memory directly but use the dynamic storage.

PR #155, opened 2024-03-12
[#153] use dynamic storage in shared memory

### LINKED ISSUE #14: Versioning of all IPC capable `iceoryx2_cal` constructs
## Problem

When two `iceoryx2_cal` constructs interact with each other, we need to add some initial versioning information to ensure that not to incompatible versions of the same constructs are interacting. This could lead to undefined behavior and hard to unravel bugs.

They could be, for instance incompatible since the member order of shared management constructs has changed or members have been added/removed. Or the internal communication protocol was updated.

## Solution

Introduce a `const SHM_API_VERSION: u64` constant inside `iceoryx2_cal`, that must be checked before any of these constructs start to interact on an ipc level.
Constructs that use shared memory could add this constant to the beginning, as the first value, to their shared memory. If the construct does not use shared memory, one could create a file and store the number value is binary, or human-readable value, in there.

elfenpiff:
## Notes for Reviewer

* `common.rs` was the previously `posix.rs` and moved but also refactored to use a dynamic storage instead of a posix shared memory directly
* `posix.rs` was recreated from scratch and only contains a type alias, like `process_local.rs`

1. [x] Assign PR to reviewer

## References

Relates to #153 
Closes #14


iceoryx2-pal/posix/src/macos/fcntl.rs
```diff
@@ -44,7 +46,19 @@ pub unsafe fn fcntl2(fd: int, cmd: int) -> int {
 }
 
 pub unsafe fn fchmod(fd: int, mode: mode_t) -> int {
-    crate::internal::fchmod(fd, mode)
+    // TODO iox2-156, shared memory permission cannot be adjusted with fchmod, therefore setting
```
elfenpiff, line 49: This is safe since the dynamic storage uses always two mechanisms to ensure that the underlying construct is initialized.
1. permissions
2. the first 8 bytes where the version number is stored

iceoryx2-cal/src/dynamic_storage/posix_shared_memory.rs
```diff
@@ -136,45 +200,108 @@ impl<T: Send + Sync + Debug> NamedConceptBuilder<Storage<T>> for Builder<T> {
...
+        // The mem-sync is actually not required since an uninitialized dynamic storage has
+        // only write permissions and can be therefore not consumed.
+        // This is only for the case that this strategy fails on an obscure POSIX platform.
+        //
+        //////////////////////////////////////////
+        // SYNC POINT: read Data<T>::data
+        //////////////////////////////////////////
+        let package_version = unsafe { &(*init_state) }
+            .version
+            .load(std::sync::atomic::Ordering::Acquire);
```
elBoberido, line 254: Since this is not in the hot path, what do you think about using `SeqCst`?

iceoryx2-cal/src/dynamic_storage/posix_shared_memory.rs
```diff
@@ -186,75 +313,85 @@ impl<T: Send + Sync + Debug> DynamicStorageBuilder<T, Storage<T>> for Builder<T>
...
-                .write(AtomicU64::new(IS_INITIALIZED_STATE_VALUE))
-        };
+        // The mem-sync is actually not required since an uninitialized dynamic storage has
+        // only write permissions and can be therefore not consumed.
+        // This is only for the case that this strategy fails on an obscure POSIX platform.
+        //
+        //////////////////////////////////////////
+        // SYNC POINT: write Data<T>::data
+        //////////////////////////////////////////
+        unsafe { (*version_ptr).store(get_package_version().0, Ordering::Release) };
```
elBoberido, line 332: Same here

iceoryx2-cal/src/dynamic_storage/posix_shared_memory.rs
```diff
@@ -186,75 +313,85 @@ impl<T: Send + Sync + Debug> DynamicStorageBuilder<T, Storage<T>> for Builder<T>
...
+    ) -> Result<Storage<T>, DynamicStorageOpenOrCreateError> {
+        match self.open_impl() {
+            Ok(storage) => Ok(storage),
+            Err(DynamicStorageOpenError::DoesNotExist) => match self.create_impl(initial_value) {
+                Ok(storage) => Ok(storage),
+                Err(DynamicStorageCreateError::AlreadyExists) => Ok(self.open_impl()?),
+                Err(e) => Err(e.into()),
+            },
+            Err(e) => Err(e.into()),
+        }
```
elBoberido, line 394: If you really want to ensure to mitigate a race, this needs to run in a loop since it can always happen that there is a competing creation/deletion of the shared memory.

A comment of todo would also be fine in this case since it will be quite unlikely that something like that will actually happen in reality.

iceoryx2-pal/posix/src/macos/fcntl.rs
elfenpiff: I would love to, but the logger is two layers above this. To do this, we need the logger below the posix platform abstraction, like in iceoryx1.

The only thing I can do is to print it directly to the console, but then the user is spammed with a lot of useless warning messages.

iceoryx2-cal/tests/dynamic_storage_posix_shared_memory_tests.rs
```diff
@@ -0,0 +1,84 @@
...
+            .unwrap();
+
+        unsafe {
+            *(raw_shm.base_address().as_ptr() as *mut u64) = u64::MAX;
+        }
+
+        let sut = <Sut as DynamicStorage<TestData>>::Builder::new(&storage_name).open();
+
+        assert_that!(sut, is_err);
+        assert_that!(sut.err().unwrap(), eq DynamicStorageOpenError::VersionMismatch);
```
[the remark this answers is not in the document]
elfenpiff: Currently, the package version is used that is stored in `CARGO_PKG_VERSION_{MAJOR|MINOR|PATCH}`.

iceoryx2-cal/src/dynamic_storage/posix_shared_memory.rs
elfenpiff: Could you explain why? Just be more on the safer side?

iceoryx2-pal/posix/src/macos/fcntl.rs
elBoberido: Are you able to redirect the output of the log crate to the console logger? If yes, then this could become our interface. Not in this PR of course.

iceoryx2-cal/tests/dynamic_storage_posix_shared_memory_tests.rs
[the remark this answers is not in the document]
elBoberido: Yeah, I noticed that further down. This means that only the applications which are build with the same iceoryx2 version can communicate with each other. At least for the foreseeable future this makes sense. Maybe (a big maybe) at some point we might soften the restriction.

iceoryx2-cal/src/dynamic_storage/posix_shared_memory.rs
elfenpiff: This - you guy - seems like a smarty one ... I will listen to him.

iceoryx2-cal/src/dynamic_storage/posix_shared_memory.rs
```diff
@@ -383,14 +335,19 @@ impl<'builder, T: Send + Sync + Debug> DynamicStorageBuilder<'builder, T, Storag
...
-                Err(DynamicStorageCreateError::AlreadyExists) => Ok(self.open_impl()?),
-                Err(e) => Err(e.into()),
-            },
-            Err(e) => Err(e.into()),
+        loop {
+            match self.open_impl() {
+                Ok(storage) => return Ok(storage),
+                Err(DynamicStorageOpenError::DoesNotExist) => match self.create_impl() {
+                    Ok(shm) => {
+                        let shm = shm;
```
elBoberido, line 343: This can be removed

elBoberido (dismissed):

iceoryx2-pal/configuration/src/lib.rs
```diff
@@ -39,3 +39,17 @@ pub mod settings {
...
+    pub const TEST_DIRECTORY: &[u8] = b"/tmp/iceoryx2/tests/";
+    pub const SHARED_MEMORY_DIRECTORY: &[u8] = b"/dev/shm/";
+    pub const PATH_SEPARATOR: u8 = b'/';
+    pub const ROOT: &[u8] = b"/";
+    pub const FILENAME_LENGTH: usize = 255;
+    pub const PATH_LENGTH: usize = 4096;
+    pub const AT_LEAST_TIMING_VARIANCE: f32 = 0.75;
+}
+#[cfg(target_os = "macos")]
+pub use settings::*;
```
elBoberido, line 55: This does not need to be replicated multiple times with a `cfg` attribute but set only once without the attribute.

elBoberido (approved):
Just a minor thing which can also be changed in another PR if desired.

PR #174, opened 2024-04-09
iox2-#165 stresstests for service creation

### LINKED ISSUE #165: Stresstests for 'Service' creation
## Brief feature description

Create and close `Services` concurrently from multiple threads to catch robustness issues in the CI.

elBoberido:
## Notes for Reviewer

This PR adds stress tests for service creation, opening and closing.

The `assert_that` macro is also extended for an `any_of` check.

1. [x] Assign PR to reviewer

## References

Closes #165

### DIFF
--- a/iceoryx2-cal/tests/static_storage_trait_tests.rs
+++ b/iceoryx2-cal/tests/static_storage_trait_tests.rs
@@ -16,9 +16,14 @@ mod static_storage {
     use iceoryx2_bb_posix::unique_system_id::UniqueSystemId;
     use iceoryx2_bb_system_types::file_name::FileName;
     use iceoryx2_bb_testing::assert_that;
+    use iceoryx2_bb_testing::watchdog::Watchdog;
     use iceoryx2_cal::named_concept::*;
+    use iceoryx2_cal::static_storage::StaticStorageCreateError;
     use iceoryx2_cal::static_storage::*;
+    use std::sync::atomic::{AtomicU64, Ordering};
+    use std::sync::Barrier;
     use std::sync::Mutex;
+    use std::time::Duration;
 
     /// The list all storage tests requires that all other tests are not interfering and therefore
     /// we cannot let them run concurrently.
@@ -258,6 +263,48 @@ mod static_storage {
         }
 
         assert_that!(<Sut as NamedConceptMgmt>::list().unwrap(), len 0);
+    }
+
+    #[test]
+    fn concurrent_create_same_locked_storage_multiple_times_fails_for_all_but_one<
+        Sut: StaticStorage,
+    >() {
+        let _watch_dog = Watchdog::new_with_timeout(Duration::from_secs(5));
+        const NUMBER_OF_THREADS: usize = 4;
+        const NUMBER_OF_ITERATIONS: usize = 1000;
+
+        let success_counter = AtomicU64::new(0);
+        let barrier_enter = Barrier::new(NUMBER_OF_THREADS);
+        let barrier_exit = Barrier::new(NUMBER_OF_THREADS);
+        let storage_name = generate_name();
+
+        std::thread::scope(|s| {
+            let mut threads = vec![];
+            for _ in 0..NUMBER_OF_THREADS {
+                threads.push(s.spawn(|| {
+                    for _ in 0..NUMBER_OF_ITERATIONS {
+                        barrier_enter.wait();
+
+                        let sut = Sut::Builder::new(&storage_name).create_locked();
+                        match sut {
+                            Ok(_) => {
+                                success_counter.fetch_add(1, Ordering::Relaxed);
+                            }
+                            Err(e) => {
+                                assert_that!(e, eq StaticStorageCreateError::AlreadyExists);
+                            }
+                        }
+
+                        barrier_exit.wait();
+                    }
+                }));
+            }
+
+            for thread in threads {
+                thread.join().unwrap();
+            }
+            assert_that!(success_counter.load(Ordering::Relaxed), eq NUMBER_OF_ITERATIONS as u64)
+        });
     }
 
     #[test]
--- a/iceoryx2/tests/service_tests.rs
+++ b/iceoryx2/tests/service_tests.rs
@@ -34,8 +34,72 @@ mod service {
         .unwrap()
     }
 
+    trait SutFactory {
+        type Factory;
+        type CreateError: std::fmt::Debug;
+        type OpenError: std::fmt::Debug;
+
+        fn create(service_name: &ServiceName) -> Result<Self::Factory, Self::CreateError>;
+        fn open(service_name: &ServiceName) -> Result<Self::Factory, Self::OpenError>;
+
+        fn assert_create_error(error: Self::CreateError);
+        fn assert_open_error(error: Self::OpenError);
+    }
+
+    impl<Sut: Service> SutFactory for publish_subscribe::PortFactory<Sut, u64> {
+        type Factory = publish_subscribe::PortFactory<Sut, u64>;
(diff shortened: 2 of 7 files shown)


iceoryx2/tests/service_tests.rs
```diff
@@ -53,6 +63,230 @@ mod service {
...
+            let mut threads = vec![];
+            threads.push(s.spawn(|| {
+                for _ in 0..NUMBER_OF_ITERATIONS {
+                    let sut = Factory::create(&service_name).unwrap();
+                    barrier_enter.wait();
+
+                    drop(sut);
+
+                    barrier_exit.wait();
+                }
```
elBoberido, line 238: @elfenpiff it seems the creation of the service fails with sporadically with `AlreadyExists` when there are more threads than cores. Is there already a way to clean everything up?

iceoryx2-cal/tests/static_storage_trait_tests.rs
```diff
@@ -260,6 +265,48 @@ mod static_storage {
         assert_that!(<Sut as NamedConceptMgmt>::list().unwrap(), len 0);
     }
 
+    #[test]
+    fn concurrent_create_same_locked_storage_multiple_times_fails_for_all_but_one<
+        Sut: StaticStorage,
+    >() {
+        let _watch_dog = Watchdog::new_with_timeout(Duration::from_secs(5));
```
elfenpiff, line 272: Can you use just `Watchdog::new()`? It comes with a default timeout, and if it is too generous (60s), we can correct it globally.
And I think it is not a problem to be a bit more generous since the watchdog only activates when the test is deadlock that is hopefully a rare cornercase

iceoryx2/tests/service_tests.rs
```diff
@@ -53,6 +63,230 @@ mod service {
...
+        let barrier_exit = Barrier::new(NUMBER_OF_THREADS);
+
+        std::thread::scope(|s| {
+            let mut threads = vec![];
+            for _ in 0..NUMBER_OF_THREADS {
+                threads.push(s.spawn(|| {
+                    for _ in 0..NUMBER_OF_ITERATIONS {
+                        barrier_enter.wait();
+
+                        let service_name = generate_name();
```
elfenpiff, line 118: What do you think of precreating the service names before the test run so that in the section, where we want the algorithms to interfere with each other, only the exciting stuff happens?

So you could create a service name in the beginning of the thread - `Vec` with `NUMBER_OF_ITERATIONS` generated name and then access it from here. I think you would need a third init barrier for that.

iceoryx2/tests/service_tests.rs
```diff
@@ -53,6 +63,230 @@ mod service {
...
+    fn concurrent_creating_services_with_same_name_fails_for_all_but_one_impl<
+        Factory: SutFactory,
+        F: Sync,
+    >(
+        error_check: F,
+    ) where
+        F: Fn(Factory::CreateError),
+    {
+        let _watch_dog = Watchdog::new_with_timeout(Duration::from_secs(30));
+        const NUMBER_OF_THREADS: usize = 4;
```
elfenpiff, line 151: What do you think about maximizing the used threads to the number of cores available, which should be at least 2? This is what I do for other concurrent test cases. Yes, I know it then depends on the machine it is running, but my thinking was that the more meaningfully stressed a test is, the more likely a race condition reveals itself —and this is what we ultimately want.

So you would rewrite this to:
```rust
let number_of_threads = (SystemInfo::NumberOfCpuCores.value()).clamp(2, usize::MAX);
```

iceoryx2/tests/service_tests.rs
elfenpiff: @elBoberido No, sadly, not yet. It would belong into `iceoryx2/src/service/mod.rs` but it comes with some challenges. So first, the static storage must become inaccessible, then the dynamic storage of the service has to be removed, and finally, the static storage.

And this has to be tested concurrently so that when one tries to open a service that is currently removed, either the remove fails with still in-use failure or the open fails with no longer exists. Maybe it makes then also sense to provide a `force_remove` method that does not care for anything and just removes the artifacts in the order above.

But when you think further, the pair messaging-pattern/service-name is unique in the system, so with remove you have to provide the messaging pattern as well.
Maybe it makes more sense to add some kind of builder and remove a service like:
```rust
Service::event().remove("some_name");
Service::publish_subscribe().force_remove("some_name");
```

iceoryx2/tests/service_tests.rs
```diff
@@ -53,6 +63,230 @@ mod service {
...
+
+            for thread in threads {
+                thread.join().unwrap();
+            }
+        });
+    }
+
+    #[test]
+    fn concurrent_creating_services_with_unique_names_is_successful<Sut: Service>() {
+        concurent_creating_services_with_unique_names_is_successful_impl::<
```
elfenpiff, line 134: Could you please split those up in two tests. One for event and one for publish_subscribe?

iceoryx2/tests/service_tests.rs
elBoberido: I don't think another barrier is required but time will tell :)

iceoryx2/tests/service_tests.rs
elBoberido: I just noticed locally that the really intersting stuff happens when the number of threads is around two times the number of logical cores

iceoryx2/tests/service_tests.rs
elfenpiff: I realized that sometimes one variation fails and the other doesn't so when the CI fails you can - when its need to be - at least try to fix it without running a native target.

iceoryx2/tests/service_tests.rs
elBoberido: Okay, then I'll skip the test for now

iceoryx2/tests/service_tests.rs
elBoberido: It was simpler to just move the service name creation above the barrier

iceoryx2/tests/service_tests.rs
```diff
@@ -24,8 +34,90 @@ mod service {
...
+        }
+
+        fn assert_create_error(error: Self::CreateError) {
+            assert_that!(
+                error,
+                any_of([
+                    PublishSubscribeCreateError::AlreadyExists,
+                    PublishSubscribeCreateError::IsBeingCreatedByAnotherInstance,
+                ])
+            );
```
elBoberido, line 69: Fun fact. Using braces for the content after `any_of` or `eq` made rustfmt work on the macro

iceoryx2/tests/service_tests.rs
```diff
@@ -24,8 +34,90 @@ mod service {
...
+        }
+        fn assert_open_error(error: Self::OpenError) {
+            assert_that!(
+                error,
+                any_of([
+                    PublishSubscribeOpenError::DoesNotExist,
+                    PublishSubscribeOpenError::PermissionDenied,
+                    PublishSubscribeOpenError::ServiceInCorruptedState,
+                    PublishSubscribeOpenError::UnableToOpenDynamicServiceInformation,
```
elBoberido, line 78: Not sure if all of these are expected to happen when a service is opened while being closed from another thread.

elfenpiff: What do you think about merging this PR and addressing the error message cleanup in a follow-up issue? Meaning, in those scenarios the user should only receive the error message `DoesNotExist`

elfenpiff (approved):

iceoryx2/tests/service_tests.rs
elBoberido: Will be done with #183

Issue #200, opened 2024-04-19
Introduce `atomic` type for iceoryx


elfenpiff:
## Brief feature description

Our `Atomic*` strategy has currently the problems that:
* on 32-bit `AtomicU64` is not available, not even with a mutex like in the C++ STL
* cannot be tested with [loom](https://crates.io/crates/loom)
* that all atomic do not support other types with the same alignment/size that are not provided by the default types
* they cannot be overloaded for advanced error handling strategies like fail in a compare-exchange loop

The new atomic shall be exchangeable via feature flags.

## ToDo

* [x] Implement atomic
* [ ] add feature flag
* [x] Integrate atomic in iceoryx so that everything uses the atomic

PR #202, opened 2024-04-22
[#200] 32 bit atomic64

elfenpiff:

1. [x] Assign PR to reviewer

## References

Relates to #200


doc/release-notes/iceoryx2-unreleased.md
```diff
@@ -9,6 +9,7 @@
  <!-- NOTE: Add new entries sorted by issue number to minimize the possibility of conflicts when merging. -->
 
  * Subscriber buffer size can be reduced [#19](https://github.com/eclipse-iceoryx/iceoryx2/issues/19)
+ * Introduce `IceAtomic` that supports up to 128bit atomics on 32-bit architecture with a ReadWriteLock [#200](https://github.com/eclipse-iceoryx/iceoryx2/issues/200)
```
elBoberido, line 12: Philosophical question, `Ice` or `Iox`? I think you already used `Iox` in some other place

iceoryx2-pal/concurrency-sync/src/ice_atomic.rs
```diff
@@ -0,0 +1,356 @@
...
+        &self,
+        current: T,
+        new: T,
+        _success: Ordering,
+        _failure: Ordering,
+    ) -> Result<T, T> {
+        self.write_lock();
+        let data = unsafe { *self.data.get() };
+        if data != current {
+            core::sync::atomic::fence(Ordering::SeqCst);
```
elBoberido, line 174: Shouldn't the lock take care of the synchronization?

elBoberido (commented):
Did not yet look at the tests

doc/release-notes/iceoryx2-unreleased.md
elfenpiff: Mmmhhh... `iox` is more consistent with the current philosophy but `ice` sounds literally cooler.

I would prefer `ice` in the future. Maybe also for the c binding, that every function starts with `ice`.

What do you think?!

iceoryx2-pal/concurrency-sync/src/ice_atomic.rs
elfenpiff: The lock is based on an `AtomicU32` with acquire release semantics. When the user wants to have sequential consistency it is not enough. And for 32-bit I would make everything sequentially consistent in the "compatibility" ice atomic

elBoberido: I'm not sure that throwing fences everywhere is a good idea. If the lock does not properly synchronize then it is broken and needs to be fixed, else one would need to add those fences to all the places it is used. It might even be a good chance to test the look with these atomics. Unfortunately it won't be easy to get a native 32 bit target to the CI

doc/release-notes/iceoryx2-unreleased.md
elBoberido: I'm leaning more towards `iox` but maybe just because I'm more used to it

elfenpiff: I renamed it into `IoxAtomic`

iceoryx2-pal/concurrency-sync/src/ice_atomic.rs
elfenpiff: The reason for the `seq_cst` fence is that a user's algorithm may depend on being sequentially consistent, but the RwLock has only acquired release memory order.

And instead of thinking everything through to the last detail how what where can how interact, I thought: be safe use a sequentially consistent memory barrier.

Do you think this can cause bugs or problems when we also assume that 32-bit support is only tier 2?

elBoberido: That depends how we define tier 2 but at some point it needs to become lock-free

elBoberido (commented):
The idea behind the compatibility tests is a bit unclear and I did not yet fully review them

iceoryx2-pal/concurrency-sync/tests/iox_atomic_tests.rs
```diff
@@ -0,0 +1,619 @@
...
+        let result = sut.fetch_sub(n2, Ordering::Relaxed);
+
+        assert_that!(result, eq n1);
+        assert_that!(sut.load(Ordering::Relaxed), eq n1.overflowing_sub(n2).0);
+    }
+
+    fn ok_fetch_update<T: AddAssign + Copy>(value: T) -> Option<T> {
+        let mut temp = value;
+        temp += value;
+        Some(temp)
```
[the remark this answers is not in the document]
elfenpiff: This does not work since `T` only has `AddAssign` and does not support multiplication with integers.

iceoryx2-pal/concurrency-sync/tests/iox_atomic_tests.rs
```diff
@@ -0,0 +1,619 @@
...
+    }
+
+    #[test]
+    fn compatibility_new_works<T: Req>() {
+        let n = T::generate_compatibility_value();
+        let sut = IoxAtomic::<T>::new(n);
+        let compat = AtomicU32::new(n.to_u32());
+
+        assert_that!(compat.load(Ordering::Relaxed), eq sut.load(Ordering::Relaxed).to_u32());
+    }
```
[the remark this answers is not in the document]
elfenpiff: The idea is to check if the `AtomicU32` behaves exactly like the `IoxAtomic`. Since the `IoxAtomic` covers everything from `u8` to `u128` I took `AtomicU32` as a representative - it is the largest atomic supported on 32-bit platforms. Furthermore, every atomic from `u8` to `u128` behaves identically when we ignore the integer size.

Issue #103, opened 2024-01-28
Enable decentral monitoring in runtime (or node)


elfenpiff:
## Brief feature description

When an application crashes, the resources are not yet cleaned up. With #96 we introduced to a mechanism to check if a process is available or has crashed. The Node in #102 shall instantiate a monitor token so that it can be monitored from within other processes.
If the runtime/node is signaled as dead, the decentral cleanup API shall remove all resources of the dead process.

PR #260, opened 2024-07-02
[#103] thread safe node registration

elfenpiff:
## Notes for Reviewer

 * The key points that should be reviewed most carefully are the changes in: `unique_index_set.rs`
   * The struct `HeadDetails` was introduced to make the bit shifting easier
   * The head now contains a third parameter, the number of borrowed indices
   * it enables us to add the option to lock the `unique_index_set` when the last index is released
   * required for service cleanup, when the last node releases itself, the service is locked, marked for destruction and cleaned up

1. [x] Assign PR to reviewer

## References

Closes #


iceoryx2/src/service/static_config/mod.rs
```diff
@@ -24,13 +24,16 @@ pub mod publish_subscribe;
...
-use crate::service::messaging_pattern::MessagingPattern;
+pub mod messaging_pattern;
+
 use iceoryx2_bb_log::fatal_panic;
 use iceoryx2_cal::hash::Hash;
 use serde::{Deserialize, Serialize};
 
 use crate::config;
 
+use self::messaging_pattern::MessagingPattern;
```
elBoberido, line 35: Is this different from `crate::service::messaging_pattern::MessagingPattern`?

iceoryx2-cal/src/static_storage/file.rs
```diff
@@ -253,24 +253,36 @@ impl crate::named_concept::NamedConceptMgmt for Storage {
...
-            Ok(true) => (),
-            Ok(false) => return Ok(false),
+        let does_exist = || match File::does_exist(&adjusted_path) {
+            Ok(true) => Ok(true),
+            Ok(false) => Ok(false),
             Err(v) => {
                 fail!(from origin, with NamedConceptDoesExistError::UnderlyingResourcesCorrupted,
                     "{} due to an internal failure ({:?}), is the static storage in a corrupted state?", msg, v);
             }
         };
```
elBoberido, line 263: This could be done with a `map_err` instead of a full match.

elBoberido (commented):
Done with the easy part. 17 files left

iceoryx2/src/service/static_config/mod.rs
elfenpiff: Yes, this is the generic plain MessagingPattern enum. The type `MessagingPattern` is again available for `static_config` and `dynamic_config` which then contains the messaging pattern specific dynamic or static configuration.

iceoryx2-cal/src/static_storage/file.rs
elfenpiff: `or_else` also works ... I will use this instead.

elBoberido: Slightly less idiomatic but works with the macro :)

iceoryx2-bb/lock-free/src/mpmc/unique_index_set.rs
```diff
@@ -201,20 +231,47 @@ impl Drop for UniqueIndex<'_> {
...
+    aba: u16,
+    borrowed_indices: u32,
+}
+
+impl HeadDetails {
+    fn from(value: u64) -> Self {
+        Self {
+            head: ((value & 0xffffff0000000000) >> 40) as u32,
+            aba: ((value & 0x000000ffff000000) >> 24) as u16,
+            borrowed_indices: (value & 0x0000000000ffffff) as u32,
```
elBoberido, line 254: This is more common and less error prone
```suggestion
            head: ((value >> 40) as u32) & 0xFFFFFF,
            aba: ((value >> 24) as u16,
            borrowed_indices: (value as u32) & 0xFFFFFF,
```
For head it could be even just `head: (value >> 40) as u32` but the compiler will optimize that away

iceoryx2-bb/lock-free/src/mpmc/unique_index_set.rs
```diff
@@ -554,12 +599,35 @@ impl<const CAPACITY: usize> FixedSizeUniqueIndexSet<CAPACITY> {
...
+            borrowed_indices: 54321,
+        }
+        .value();
+
+        let sut = HeadDetails::from(sut_value);
+
+        assert_that!(sut.head, eq 12345);
+        assert_that!(sut.aba, eq 6789);
+        assert_that!(sut.borrowed_indices, eq 54321);
+    }
```
elBoberido, line 632: Ideally, there would be a check that the fields are not overlapping, e.g. subsequently setting each field to it's max value and checking if the others remain 0.

iceoryx2/src/node/mod.rs
```diff
@@ -408,30 +492,55 @@ impl<Service: service::Service> Node<Service> {
...
+    /// });
+    /// ```
+    pub fn list<
+        F: FnMut(
+            Result<NodeState<Service>, NodeListFailure>,
+        ) -> Result<CallbackProgression, NodeListFailure>,
+    >(
+        config: &Config,
+        mut callback: F,
+    ) -> Result<(), NodeListFailure> {
```
elBoberido, line 510: I feared that the callback with the result input parameter would be a public API. I think for the C and C++ bindings it would be less challenging to have a different API with two callbacks, one for the happy path and one for the error path. The `Result` as return parameter still makes it challenging but still easier to implement than with the `Result` as input parameter.

iceoryx2/src/node/mod.rs
```diff
@@ -408,30 +492,55 @@ impl<Service: service::Service> Node<Service> {
...
-            };
+        match Self::list_all_nodes(&monitoring_config) {
+            Ok(node_list) => {
+                for node_name in node_list {
+                    let node_id = core::str::from_utf8(node_name.as_bytes()).unwrap();
+                    let node_id = NodeId(node_id.parse::<u128>().unwrap().into());
+
+                    match NodeState::new(&node_id, config) {
+                        Ok(Some(node_state)) => {
+                            if callback(Ok(node_state))? == CallbackProgression::Stop {
```
elBoberido, line 523: Another reason to split it. It's always used like `callback(Ok(...))` or `callback(Err(...))` and the first thing in the callback is a match on the `Result` input value.

elBoberido (commented):
Except the callback signature, this looks good. Regarding the callback, I would strongly recommend not to use a `Result` as input parameter but use two callbacks, one for the happy path and one for the sad path

iceoryx2/src/service/static_config/messaging_pattern.rs
```diff
@@ -0,0 +1,71 @@
...
+        &self,
+        publisher_max_loaned_samples: usize,
+    ) -> usize {
+        match self {
+            MessagingPattern::PublishSubscribe(v) => {
+                v.max_subscribers
+                    * (v.subscriber_max_buffer_size + v.subscriber_max_borrowed_samples)
+                    + v.history_size
+                    + publisher_max_loaned_samples
+                    + 1
```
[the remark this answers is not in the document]
elfenpiff: When I remove it, all tests that verify that the publisher never runs out of memory pass. And I no longer remember why I added +1 in the first place?!

So I removed it ... let's see what happens...

elBoberido (dismissed):

iceoryx2/src/node/mod.rs
elBoberido: Since this is an API discussion and not a technical issue, it can be resolved in a follow up PR

iceoryx2/tests/node_tests.rs
```diff
@@ -62,6 +62,10 @@ mod node {
             let view = match node {
                 NodeState::<S>::Alive(ref view) => view as &dyn NodeView,
                 NodeState::<S>::Dead(ref view) => view as &dyn NodeView,
+                NodeState::<S>::Inaccessible(_) | NodeState::<S>::Undefined(_) => {
+                    assert_that!(true, eq false);
+                    panic!();
```
elBoberido, line 67: Why not just
```suggestion
                    panic!("NodeState should not be Inaccessible or Undefined");
```
Alternatively, instead of `panic`, an `unreachable` would semantically also make sense

iceoryx2/tests/node_tests.rs
```diff
@@ -263,23 +267,28 @@ mod node {
...
                                 }
-                            }
-                            Ok(CallbackProgression::Continue)
+                                NodeState::Inaccessible(node_id) => {
+                                    if node_id == *node.id() {
+                                        found_self = true;
+                                    }
+                                }
+                                NodeState::Undefined(_) => {
+                                    assert_that!(true, eq false);
```
elBoberido, line 287: Same here and everywhere else a `assert_that!(true, eq false)` is used

elBoberido (approved):
Only some nitpicks

Issue #264, opened 2024-07-05
Create C++ language binding


elBoberido:
## Brief feature description

Create a `C++` language binding for the iceoryx2 public API.

## Detailed information

Based on the `C` language binding from #210, a `C++` binding should be created in order to make iceoryx2 available to the `C++` ecosystem.

## ToDo List

* [ ] Remove all `IOX_TODO()` calls.

### Examples

* [x] complex_data_types
* [x] discovery
* [x] event
* [x] publish_subscribe
* [x] publish_subscribe_dynamic_data
* [x] publish_subscribe_with_user_header
* [x] service_attributes
* [ ] main readme updated

### Classes

* [x] attribute.hpp
* [x] attribute_set.hpp
* [x] attribute_specifier.hpp
* [x] attribute_verifier.hpp
* [x] callback_progression.hpp
* [x] config.hpp
* [x] connection_failure.hpp
* [ ] dynamic_config_event.hpp
* [ ] dynamic_config_publish_subscribe.hpp
* [x] event_id.hpp
* [x] header_publish_subscribe.hpp
* [x] listener.hpp
* [x] message_type_details.hpp
* [x] messaging_pattern.hpp
* [x] node_details.hpp
* [x] node_failure_enums.hpp
* [ ] node.hpp
* [x] node_id.hpp
* [x] node_name.hpp
* [x] node_state.hpp
* [x] notifier.hpp
* [ ] port_factory_event.hpp
* [x] port_factory_listener.hpp
* [x] port_factory_notifier.hpp
* [ ] port_factory_publisher.hpp
* [ ] port_factory_publish_subscribe.hpp
* [x] port_factory_subscriber.hpp
* [ ] publisher.hpp
* [x] sample.hpp
* [ ] sample_mut.hpp
* [x] semantic_string.hpp
* [ ] service_builder_event.hpp
* [x] service_builder.hpp
* [ ] service_builder_publish_subscribe.hpp
* [ ] service_details.hpp
* [ ] service.hpp
* [x] service_name.hpp
* [x] service_type.hpp
* [x] static_config_event.hpp
* [ ] static_config.hpp
* [x] static_config_publish_subscribe.hpp
* [ ] subscriber.hpp
* [x] unique_port_id.hpp

mossmaurice:
@elfenpiff Is this issue already completed? According to the issue description a few API calls are still missing.

elfenpiff:
[...]

PR #321, opened 2024-07-26
[#264] pubsub

elfenpiff:

1. [x] Assign PR to reviewer

## References

Relates to #264


examples/c/event/src/listener.c
```diff
@@ -0,0 +1,81 @@
...
+    // create service
+    iox2_service_name_ptr service_name_ptr = iox2_cast_service_name_ptr(service_name);
+    iox2_node_ref_h node_ref_handle = iox2_cast_node_ref_h(node_handle);
+    iox2_service_builder_h service_builder = iox2_node_service_builder(node_ref_handle, NULL, service_name_ptr);
+    iox2_service_builder_event_h service_builder_event = iox2_service_builder_event(service_builder);
+    iox2_port_factory_event_h service = NULL;
+    if (iox2_service_builder_event_open_or_create(service_builder_event, NULL, &service) != IOX2_OK) {
+        printf("Unable to create service!\n");
+        goto drop_node;
+    }
```
elBoberido, line 45: I think for this we could create a convenience API like `iox2_event_service_create_with_default_settings(char*, iox2_port_factory_event_h*)`. Not for this PR but in the short term to make the examples less intimidating.
Alternatively, instead of following the Rust builder too closely, we could also create a struct with all the settings and pass that to a function.

What do you think?

examples/c/event/src/listener.c
```diff
@@ -0,0 +1,81 @@
...
+
+    // create listener
+    iox2_port_factory_event_ref_h ref_service = iox2_cast_port_factory_event_ref_h(service);
+    iox2_port_factory_listener_builder_h listener_builder = iox2_port_factory_event_listener_builder(ref_service, NULL);
+    iox2_listener_h listener = NULL;
+    if (iox2_port_factory_listener_builder_create(listener_builder, NULL, &listener) != IOX2_OK) {
+        printf("Unable to create listener!\n");
+        goto drop_service;
+    }
```
elBoberido, line 54: Similarly, this could also be condensed to a single function call for default settings.

This might be a case where an API to write language bindings is less ideal for an end user API.

iceoryx2-ffi/cxx/include/iox2/port_factory_publisher.hpp
```diff
@@ -16,28 +16,85 @@
...
+    : m_handle { handle } {
+}
+
+template <ServiceType S, typename Payload, typename UserHeader>
+inline auto
+PortFactoryPublisher<S, Payload, UserHeader>::create() && -> iox::expected<Publisher<S, Payload, UserHeader>,
+                                                                           PublisherCreateError> {
+    auto* ref_handle = iox2_cast_port_factory_publisher_builder_ref_h(m_handle);
+
+    m_unable_to_deliver_strategy.and_then([](auto) { IOX_TODO(); });
```
elBoberido, line 83: Of all the available options, I think this is one which we need to support with the initial bindings.

iceoryx2/src/port/publisher.rs
```diff
@@ -982,11 +983,25 @@ impl<Service: service::Service, Payload: Debug, UserHeader: Debug>
             ))
         };
 
+        let payload_size = self
+            .data_segment
+            .subscriber_connections
+            .static_config
+            .message_type_details
+            .payload
+            .size;
```
elBoberido, line 992: At some point we might need to cache this value within the publisher. I would guess this will lead to quit a few cache misses.

elBoberido (commented):
Reviewed the examples and a few of the files with only small changes.

examples/c/event/src/listener.c
elfenpiff: I like it! This API requires the user really to write a lot.

iceoryx2-ffi/cxx/include/iox2/sample_mut.hpp
```diff
@@ -13,120 +13,220 @@
...
-        IOX_TODO();
+    /// Returns a mutable reference to the user_header of the [`Sample`].
+    template <typename T = UserHeader, typename = std::enable_if_t<!std::is_same_v<void, UserHeader>, T>>
+    auto user_header_mut() -> T&;
+
+    /// Returns a reference to the const payload of the sample.
+    auto payload() const -> const Payload&;
+
+    /// Returns a reference to the payload of the sample.
+    auto payload_mut() -> Payload&;
```
elBoberido, line 86: Do we want to introduce this terminology in C++.

I'm not suggestion to replace it, I'm just curious if it would make sense to stick to C++ idioms/terminology?

But I must confess that I'm lightly leaning towards the terminology in this PR.

iceoryx2-ffi/cxx/include/iox2/sample.hpp
```diff
@@ -13,59 +13,137 @@
...
+    auto* ref_handle = iox2_cast_sample_ref_h(m_handle);
+    const Payload* payload_ptr = nullptr;
+    size_t payload_len = 0;
+
+    // NOLINTBEGIN(cppcoreguidelines-pro-type-reinterpret-cast), no other way for type erasure
+    iox2_sample_payload(ref_handle, reinterpret_cast<const void**>(&payload_ptr), &payload_len);
+    // NOLINTEND(cppcoreguidelines-pro-type-reinterpret-cast)
+    IOX_ASSERT(sizeof(Payload) <= payload_len, "");
+
+    return *payload_ptr;
```
elBoberido, line 128: If you want to get rid of the `reinterpret_cast`, this is an option
```suggestion
    auto* ref_handle = iox2_cast_sample_ref_h(m_handle);
    const void* payload_ptr = nullptr;
    size_t payload_len = 0;

    iox2_sample_payload(ref_handle, &payload_ptr, &payload_len);
    IOX_ASSERT(sizeof(Payload) <= payload_len, "");

    return *static_cast<const Payload*>(payload_ptr);
```

elBoberido (commented):
Everything from cxx reviewed. Now only 7 files from ffi are left.

iceoryx2-ffi/cxx/include/iox2/port_factory_publisher.hpp
elfenpiff: Done, see #325

iceoryx2-ffi/cxx/include/iox2/publisher.hpp
```diff
@@ -13,92 +13,179 @@
...
+    auto loan() -> iox::expected<SampleMut<S, Payload, UserHeader>, PublisherLoanError>;
+
+    /// Loans/allocates a [`SampleMut`] from the underlying data segment of the [`Publisher`]
+    /// and initializes all slice elements with the default value. This can be a performance hit
+    /// and [`Publisher::loan_slice_uninit()`] can be used to loan a slice of uninitialized
+    /// [`Payload`].
+    ///
+    /// On failure it returns [`PublisherLoanError`] describing the failure.
+    auto
+    loan_slice(uint64_t number_of_elements) -> iox::expected<SampleMut<S, Payload, UserHeader>, PublisherLoanError>;
```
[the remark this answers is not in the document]
elfenpiff: No, it should only work when `iox::IsSlice<Payload> == true`. Fixed it and enable this method only for slices with an  `enable_if`.

iceoryx2-ffi/ffi/src/api/message_type_details.rs
```diff
@@ -0,0 +1,61 @@
...
+use crate::iox2_type_variant_e;
+
+#[derive(Clone, Copy)]
+#[repr(C)]
+pub struct iox2_type_detail_t {
+    pub variant: iox2_type_variant_e,
+    pub type_name: [c_char; 256],
+    pub size: usize,
+    pub alignment: usize,
+}
```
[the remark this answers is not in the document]
elfenpiff: I agree that the version is important, that's why the `DynamicStorage` concept has an internal version that corresponds to the iceoryx2 version. It is untouchable from the outside.

The `StaticStorage` still requires such a version and I wanted to encode this into the underlying name so that you have something like `myprefix_static_storage_name_v0.1.2_mysuffix`. I find a version number at this level to error prone, the question is, what structs require a version number and which doesn't. Also it can be forgotten.

From the iceoryx2 point of view it is perfect when the underlying concepts takes care of the versioning and then you do not have to worry about this on this level.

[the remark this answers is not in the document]
elfenpiff: I created an issue #327

iceoryx2-ffi/cxx/include/iox2/sample_mut.hpp
elfenpiff: I am introducing this terminology to be as consistent as possible to the Rust API. In this sense it is much easier to switch between C++ and Rust (for instance when code must be ported).
Also the Rust documentation complements the C++ docu nicely when the APIs are closely together.

iceoryx2-ffi/ffi/src/api/message_type_details.rs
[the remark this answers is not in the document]
elBoberido: I'm more concerned about recod&replay. This information then also needs to be stored somewhere when record&replay is implemented.

iceoryx2-ffi/ffi/src/api/publisher.rs
```diff
@@ -10,74 +10,369 @@
...
+/// Return [`IOX2_OK`] on success, otherwise [`iox2_publisher_loan_error_e`].
+///
+/// # Safety
+///
+/// * `publisher_handle` is valid, non-null and was obtained via [`iox2_cast_publisher_ref_h`]
+/// * The `sample_handle_ptr` is pointing to a valid [`iox2_sample_mut_h`].
+#[no_mangle]
+pub unsafe extern "C" fn iox2_publisher_loan(
+    publisher_handle: iox2_publisher_ref_h,
+    number_of_elements: c_size_t,
```
elBoberido, line 292: I'm thinking whether it would make sense to have a `iox2_publisher_loan` without `number_of_elements` and a `iox2_publisher_loan_slice` to match the Rust API more closely.

iceoryx2-ffi/ffi/src/api/publisher.rs
```diff
@@ -10,74 +10,369 @@
...
+    *sample_handle_ptr = std::ptr::null_mut();
+
+    let mut sample_struct_ptr = sample_struct_ptr;
+    fn no_op(_: *mut iox2_sample_mut_t) {}
+    let mut deleter: fn(*mut iox2_sample_mut_t) = no_op;
+    if sample_struct_ptr.is_null() {
+        sample_struct_ptr = iox2_sample_mut_t::alloc();
+        deleter = iox2_sample_mut_t::dealloc;
+    }
+    debug_assert!(!sample_struct_ptr.is_null());
```
elBoberido, line 308: Can you please move this into a lambda and call it only when the loan was successful. Else there will be a memory leak.

You can have a look at the subscriber for inspiration.

iceoryx2-ffi/ffi/src/api/sample_mut.rs
```diff
@@ -0,0 +1,271 @@
...
+        iox2_service_type_e::IPC => {
+            let sample = ManuallyDrop::into_inner(sample.ipc);
+            match sample.assume_init().send() {
+                Ok(v) => {
+                    if !number_of_recipients.is_null() {
+                        *number_of_recipients = v;
+                    }
+                }
+                Err(e) => {
+                    (sample_struct.deleter)(sample_struct);
```
elBoberido, line 219: This needs to be removed. It leads to double deletion. I think I overlooked it when I fixed the mem leak.

elBoberido (commented):
Looks good. Only a few small things

iceoryx2-ffi/ffi/src/api/message_type_details.rs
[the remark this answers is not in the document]
elfenpiff: You are right! But I would keep those two related problems separated.
 1. Detecting version mismatches of multiple iceoryx2 applications.
   * In my opinion perfectly solved by the current approach.
 2. Versioning recorded data.
   * Here we could again some kind of record & replay concept that takes care of exactly this automatically. Here we may require a trait for all recorded data that implements some versioning information.

elBoberido (approved):

PR #379, opened 2024-09-12
[#98] add CLI tools to crates.io publish script

### LINKED ISSUE #98: ROS2-like CLI for iceoryx2
## Brief feature description

A ROS2-like CLI enabling users to interact with `iceoryx2` from the terminal would be a useful addition for users to aid development of applications. 

## Detailed information

An summary of the capabilities of the ROS 2 CLI can be found [here](https://github.com/ubuntu-robotics/ros2_cheats_sheet/blob/master/cli/cli_cheats_sheet.pdf). NOTE: The `iceoryx2` does not need to necessarily provide all of the same functionality, this should just be taken as inspiration for what could be useful.

Proposal for initial for CLI functionality:

1. `introspect` - introspect the internals of the `iceoryx2` system (e.g. things stored in shared memory)
1. `processes` - manage processes running `iceoryx2` 
1. `pub` - publish to the `iceoryx2` system from the terminal
1. `rpc` - execute `iceoryx2` rpc operations from the terminal
1. `services` - manage the services currently available in `iceoryx2`
1. `sub` - subscribe to the `iceoryx2` system from the terminal

## Tasks

Follow up from #106:
1. [x] Address TODOs in commands.rs
1. [x] Ensure implementation of commands.rs follows Rust conventions
1. [x] Display installed commands before development commands
1. [x] Work out how to test command lookup in file system
1. [x] Add `iceoryx2-cli` to `internal/scripts/crates_io_publish_script.sh` to release it.

Follow up from #371:
1. [ ] Factor out CLI setup into common lib: https://github.com/eclipse-iceoryx/iceoryx2/pull/371#discussion_r1766353301

orecham:
## Notes for Reviewer

1. put all CLI tools in a single easy-to-install crate `iceoryx2-cli`
2. add a `README.md` for the CLI crate
3. add a `markdownlint` configuration to the project
    1. note that not all markdown files have been formatted, left for a separate PR
    2. at the very least it will ensure future markdown files are consistent
    3. rules can be adapted as we go
5. rename CLI tools
    1. `iox2-services` -> `iox2-service`
    1. `iox2-nodes` -> `iox2-node`
6. add the crate with CLI tools to script for publishing to crates.io
7. remove cargo crates of unimplemented CLI tools
    1. `iox2-introspect`
    1. `iox2-processes`
    1. `iox2-pub`
    1. `iox2-sub`
    1. `iox2-rpc`
8. update release notes with the tools that are implemented

1. [ ] ~~Every source code file has a copyright header with `SPDX-License-Identifier: Apache-2.0 OR MIT`~~
1. [ ] ~~Tests follow the [best practice for testing][testing]~~
1. [ ] ~~Changelog updated [in the unreleased section][changelog] including API breaking changes~~
1. [x] Assign PR to reviewer
1. [ ] All checks have passed (except `task-list-completed`)

## References

Relates #98


internal/scripts/crates_io_publish_script.sh
```diff
@@ -32,3 +32,5 @@ cargo publish -p iceoryx2-bb-memory
 cargo publish -p iceoryx2-cal
 cargo publish -p iceoryx2-bb-trait-tests
 cargo publish -p iceoryx2
+cargo publish -p iox2
```
elfenpiff, line 35: If we publish this and the services, we also need a little readme describing how to install them. I think `cargo install iox2` and `cargo install iox2-services` should work. This could be placed into `iceoryx2-cli/README.md`

orecham: @elfenpiff It's not clear exactly what you expect. Do you have an example?

elfenpiff: I try to phrase it as a use case so that it is better understandable.

* **As a** iceoryx2 user, 
* **that** uses iceoryx2 as a dependency in my projects workspace `Cargo.toml`
* **I want** install the `iox2` cli
* **So that** I can list all services and nodes.

So the question is, how does the iceoryx2 user install `iox2` cli with all of its subcommand.
1. `cargo run --bin iox2 -- services list` does not work -> `Failed to execute command: Command not found: services`
2. `cargo run --bin iox2-services -- list` -> works but is not the intended use case.

So I would like to have a `README.md` document inside `iceoryx2-cli/iox2`, since this is the document that is shown in crates.io when the crate is published, see: https://crates.io/crates/iceoryx2

In this document there is a installation guide with terminal commands on how to install the `iox2` cli with all of its subcommands so that the subcommands are installed in the right path and the user can just use it as intended.

I assume that the intended use case is, that `iox2` is in the `PATH` of the shell and the user can just call
```sh
iox2 services list
```
for instance.

elfenpiff: @elBoberido whats your take on this?

elBoberido: I guess you mean something like https://crates.io/crates/cargo-outdated or https://crates.io/crates/cargo-list. Without a README.md in the folder, the page on crates.io will look like this https://crates.io/crates/iceoryx2-bb-container

I think havin a small README.md for the cli tools would be helpful for people looking for them on crates.io. It does not need to be fancy in the beginning, just a hint on how to install and use them, maybe with an output of a cli tool as example.

orecham: @elfenpiff @elBoberido I think I understand my confusion now, there are actually two things being asked:

1. How does a user install the `iox2` CLI tools
2. How does the user use the CLI tools

For (1) I can add a short readme to the crates, but it will essentially just boil down to `cargo install iox2`. I did some research and might need to tweak the `Cargo.tml` file slightly to make this work.

For (2), the user should use `iox2 --help`. Writing documentation (like in https://crates.io/crates/cargo-list) will be tedious to maintain IMO. I can add the instruction for user to use `iox2 --help` (or equivalent) to the readmes as a compromise.

orecham: > So the question is, how does the iceoryx2 user install iox2 cli with all of its subcommand.
> 
> cargo run --bin iox2 -- services list does not work -> Failed to execute command: Command not found: services
> cargo run --bin iox2-services -- list -> works but is not the intended use case.

I replied with the reasoning for this in a comment in another PR, but if you do `cargo run --bin iox2`, the reason it did not work will be clear.

elfenpiff: @orecham I just want to have a readme for `1. How does the user install the iox2 CLI tools`. For `2. How does the user use the CLI tools` the user can call `iox2 --help`.

elBoberido: I think a short description at the top what the tool does would be helpful, together with a link to the iceoryx2 project. Followed by how to install and how to use. For the how to use section, the most common usage should be shown, e.g. `iox2 services` and how to discover more options, e.g. `iox servives --help`. The output of the most common command would also be nice but not a must have.

orecham: @elfenpiff @elBoberido I moved all tools into a single crate for easy installation and added a short readme for that crate

you can test installation locally:

```
$ cargo install --path ./iceoryx2-cli
```

users should simply be able to do:

```
$ cargo install iceoryx2_cli
```

.markdownlint.yaml
```diff
@@ -0,0 +1,260 @@
+# Default state for all rules
```
elfenpiff, line 1: Nice, but I have some questions:
1. What do you use as linter in vim, this here: https://github.com/fannheyward/coc-markdownlint
2. Can we create a CI target that checks this. At the moment I would make it purely optional, but in my opinion, every ruleset in iceoryx2 must be checked by CI otherwise it is not a valid rule set. Maybe we can use this here: https://github.com/igorshubovych/markdownlint-cli
3. Is there a auto-formatter?

elfenpiff (approved):

.markdownlint.yaml
orecham: @elfenpiff 

1. I use: https://github.com/dense-analysis/ale
2. Yes it should be easy with https://github.com/igorshubovych/markdownlint-cli or https://github.com/DavidAnson/markdownlint-cli2 (which is supposedly faster)
3. I use: https://github.com/stevearc/conform.nvim which uses one of the CLIs on-save (2)

Issue #390, opened 2024-09-22
iceoryx WaitSet integration


elfenpiff:
## Brief feature description

In iceoryx2 we require an event multiplexing mechanism like the iceoryx waitset. It is connected to `Node::wait()`  and maybe integrated into the `Node` or a separate mechanism.

## ToDo List

 * [x] Implement WaitSet
 * [x] C and C++ bindings
 * [ ] Add example how event based communication can look like

## Terms

 * **tick** - periodic event to produce something
 * **timeout** - periodic deadline event to consume something

## API

```rust
impl WaitSetBuilder {
  fn new() -> Self;
  fn allocator(); // sets custom allocator to now allocate management stuff on the heap
  fn tick(CyclicTimer); // sets tick timeout, whenever wait/run shall emit WaitEvent::Tick
  fn create();
}

pub enum Attachment<'a, T: SynchronousMultiplexing> {
  Notification(&'a T),
  Deadline((&'a T, Duration)),
  Tick(Duration),
};

impl WaitSet {
  fn attach<T>(attachment: Attachment);
  fn run(callback); // runs in a timed loop defined by tick
  fn try_run(callback); // no ticks? tries to gather all events.
}
```

### Details

 * `SynchronousMultiplexing` is a trait implemented by every file descriptor based construct where the file descriptor emits events and can be attached to `select`, `poll` or `epoll`
 * Every struct that implements `SynchronousMultiplexing` shall be attachable
   * Enables the user to implement the trait for sockets and attach them
 * A `Listener` port shall be attachable.
 * `Listener` ports from different `Node`s shall be attachable.
   * When the app acts as a gateway between 2 isolated iceoryx2 domains the gateway shall wake up whenever it receives a message from one of the domains without requiring multiple threads
 * `Node` knows all ports, the first idea was to handle background tasks like `Publisher::update_connections()` in `Node::wait()` to deliver the history to newly connected `Subscriber`s
 * `Publisher::update_connections()` problem, how is the `Subscriber`s history delivered?
   * can be handled with cyclic `Node::wait(CYCLE_TIME)` when the node calls it explicitly for every port - can cause a latency up to `CYCLE_TIME`
   * can be handled explicitly by the user - can be error prone when the user has to call this for every publisher
   * Combine pubsub service with event service with the same name.
     * specify and reserve an `IOX2_SERVICE_UPDATE` id and emit this whenever a `Subscriber` did connect to a service
[...]

PR #424, opened 2024-10-01
[#390] waitset

elfenpiff:
## Notes for Reviewer

 * Implements a high-level WaitSet by utilizing the `iceoryx2_cal::reactor`
   * WaitSet is end-user API
   * reactor is a iceoryx2 concept that can be implemented by `select`, `epoll` or `io_uring` multiplexing
 * Provides a first example demonstrating the waitset in Rust - C & C++ binding will follow in separate PR
 * The `NodeEvent` was renamed into `WaitEvent` to signal to the user that `Node::wait()` will utilize an internal `WaitSet` (in the future)
 

1. [x] Assign PR to reviewer

## References

Relates to #390


examples/rust/event_multiplexing/wait.rs
```diff
@@ -0,0 +1,81 @@
...
+    // attach all listeners to the waitset and store the guard
+    for listener in listeners.values() {
+        guards.push(waitset.attach(&listener.1)?);
+    }
+
+    println!("Waiting on the following services: {:?}", args.services);
+
+    // the callback that is called when a listener has received an event
+    let trigger_call = |attachment| {
+        if let Some((service_name, listener)) = listeners.get(&attachment) {
```
elBoberido, line 55:
```suggestion
    let trigger_call = |attachment_id| {
        if let Some((service_name, listener)) = listeners.get(&attachment_id) {
```
Or maybe just `id` to keep it short?

iceoryx2/src/port/waitset.rs
```diff
@@ -0,0 +1,349 @@
...
+//!     let listener = if attachment_id.originates_from(&listener_1) {
+//!         &listener_1
+//!     } else {
+//!         &listener_2
+//!     };
+//!
+//!     while let Ok(Some(event_id)) = listener.try_wait_one() {
+//!         println!("received notification {:?}", event_id);
+//!     }
+//! }, Duration::from_secs(1)) != Ok(WaitEvent::TerminationRequest) {}
```
elBoberido, line 46: What do you think about this
```rs
let event_handler = |attachment_id| {
    // ...
}

while waitset.timed_wait(event_handler, Duration::from_secs(1)) != Ok(...) {}
```
Else one has already forgotten whats going on when reaching the line with `}, Duration::from...`

I would also propose to switch the parameter and have the timeout as first parameter. It reads more fluid
> timed wait for 'duration' and do event handling if not a termination request

What do you think about a
```rs
waitset.run(event_handler);
```
If someone wants a timeout, it would look like
```rs
let deadline = Deadline(100ms);
waitset.attach(deadline);
```

iceoryx2/src/port/waitset.rs
```diff
@@ -0,0 +1,349 @@
...
+//! for listener in listeners.values() {
+//!     guards.push(waitset.attach(listener)?);
+//! }
+//!
+//! while waitset.timed_wait(|attachment_id| {
+//!     if let Some(listener) = listeners.get(&attachment_id) {
+//!         while let Ok(Some(event_id)) = listener.try_wait_one() {
+//!             println!("received notification {:?}", event_id);
+//!         }
+//!     }
```
elBoberido, line 86: What do you think of this?
```suggestion
//!     listeners.get(&attachment_id).map(|listener| {
//!         while let Ok(Some(event_id)) = listener.try_wait_one() {
//!             println!("received notification {:?}", event_id);
//!         }
//!     })
```

elBoberido (commented):
I need a break before reviewing the waitset test. Everything else is reviewed. Looks good.

iceoryx2/tests/waitset_tests.rs
```diff
@@ -0,0 +1,388 @@
...
+        let wait_event = sut
+            .timed_wait(
+                |_| {
+                    test_fail!("only attachments shall trigger");
+                },
+                TIMEOUT,
+            )
+            .unwrap();
+
+        assert_that!(wait_event, eq WaitEvent::Tick);
```
elBoberido, line 340: I'm wondering whether it would make sense to differentiate between a tick and a timeout. Semantically they are different. The former is expected to happen at some defined interval while the later is a notification that something did not happen.

iceoryx2/tests/waitset_tests.rs
```diff
@@ -0,0 +1,388 @@
...
+            notifier.notify().unwrap();
+        });
+
+        barrier.wait();
+        let wait_event = sut.blocking_wait(|_| {}).unwrap();
+
+        t1.join().unwrap();
+
+        assert_that!(wait_event, eq WaitEvent::Notification);
+        assert_that!(start.elapsed(), time_at_least TIMEOUT);
```
elBoberido, line 380:
```suggestion
        barrier.wait();
        let wait_event = sut.blocking_wait(|_| {}).unwrap();
        
        assert_that!(wait_event, eq WaitEvent::Notification);
        assert_that!(start.elapsed(), time_at_least TIMEOUT);
        
        t1.join().unwrap();
```

The current test checks the sleep in the thread :)

examples/rust/event_multiplexing/wait.rs
```diff
@@ -0,0 +1,81 @@
...
+                print!(" {:?}", event_id);
+            }
+
+            println!("");
+        }
+    };
+
+    // wait until at least one listener has received an event or the user has pressed CTRL+c
+    // or send SIGTERM/SIGINT
+    while waitset.timed_wait(trigger_call, CYCLE_TIME) != Ok(WaitEvent::TerminationRequest) {}
```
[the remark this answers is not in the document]
elfenpiff: What would you suggest?

iceoryx2-cal/tests/reactor_tests.rs
```diff
@@ -91,6 +91,21 @@ mod reactor {
...
+    #[test]
+    fn attach_the_same_attachment_twice_fails<Sut: Reactor>() {
+        let sut = <<Sut as Reactor>::Builder>::new().create().unwrap();
+
+        let name = generate_name();
+        let listener = unix_datagram_socket::ListenerBuilder::new(&name)
+            .create()
+            .unwrap();
+
+        let _guard = sut.attach(&listener).unwrap();
```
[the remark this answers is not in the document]
elfenpiff: In my opinion, I prefer reading this than the `Box<dyn ...` in the test name and the `Ok(())` at the end ... It is just a personal preference.

iceoryx2/src/port/waitset.rs
```diff
@@ -0,0 +1,349 @@
...
+// This program and the accompanying materials are made available under the
+// terms of the Apache Software License 2.0 which is available at
+// https://www.apache.org/licenses/LICENSE-2.0, or the MIT license
+// which is available at https://opensource.org/licenses/MIT.
+//
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! # Example
+//!
+//! ## Use [`AttachmentId::originates_from()`](crate::port::waitset::AttachmentId)
```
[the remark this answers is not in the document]
elfenpiff: Yes, but before we do this I would like to know what else would make sense there and also what kind of rule we follow. Just reexporting `AttachmentId` seems wrong.

But if we have a clear strategy then I would do it. Shall I create an issue?

iceoryx2/src/port/waitset.rs
elfenpiff: I thought about this, and this gets even more complex. Let's assume you have a `WaitSet` with multiple attachments. These attachments have a liveliness, the first every 23ms, the second every 50ms and the third every 78ms. You want to wake up when no message was received and the timeout was hit. So you require a deadline timer with a priority queue.

But this would not be attachable with the attach API since it would require firing an fd event. This could be handled completely via the internal wait time - you just need to know how long you have to wait and what timeout you hit.

So my idea was to add this later to the builder parameters as a parameter.

I like `WaitSet::run()` - but there are already very simplistic examples where you need a variable deadline. Take a look at this loop: https://github.com/elfenpiff/iceoryx2/blob/iox2-390-event-based-comm-example/examples/rust/event_based_communication/publisher.rs#L38

The goal here is to send out a sample every second. But in between new events may be triggered so you have to rewait on the waitset but now have to subtract the elapsed time.

iceoryx2/src/port/waitset.rs
elfenpiff: Why is this better?

iceoryx2/tests/waitset_tests.rs
elfenpiff: I agree with your point, but I am unsure how to realize this. How could I distinguish between tick and timeout. A tick is emitted every X seconds whilst a timeout is reached after X seconds. So if the tick time is lower than the timeout time, timeout will never be emitted, if it is greater, then timeout and tick will alternate. Mmmmhhh ... I am not sure what to do here...

iceoryx2/src/port/waitset.rs
elfenpiff: Also, what if you do not want to block ... would it be then `WaitSet::run` and `WaitSet::try_run` ?!

iceoryx2-cal/tests/reactor_tests.rs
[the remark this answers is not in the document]
elBoberido: I have to check why I formed a habit to returning an error instead of unwrapping. It might be that the error message was not fully clear when it failed due to an unwrap but maybe that changed meanwhile

iceoryx2/src/port/waitset.rs
elBoberido: A little bit less to type but it doesn't really matter. This is just personal style and one is not better than the other :)

iceoryx2/src/port/waitset.rs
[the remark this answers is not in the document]
elBoberido: The question is more like, do we remember that there is an issue :sweat_smile: 

It's not a breaking change anyway, so we can just rely on someone coming up with it again at some point int the future.

examples/rust/event_multiplexing/wait.rs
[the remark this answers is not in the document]
elBoberido: Good question. I need to take counsel with my pillow.

But I noticed that the code has a logic error. It should be
```rs
while let Ok(event) = waitset.timed_wait(trigger_call, CYCLE_TIME) && event != WaitEvent::TerminationRequest {}
```

iceoryx2/src/port/waitset.rs
elBoberido: I was thinking of something where you setup your application the then just run the event loop. There would be nor `try_run` or `timed_run`. Maybe that's something for a higher-level API and the `WaitSet` needs more fine grained control, just like it is now.

examples/rust/event_multiplexing/wait.rs
```diff
@@ -0,0 +1,83 @@
...
+                print!(" {:?}", event_id);
+            }
+
+            println!("");
+        }
+    };
+
+    // wait until at least one listener has received an event or the user has pressed CTRL+c
+    // or send SIGTERM/SIGINT
+    while waitset.run(trigger_call).is_ok() {}
```
orecham, line 70: Just wondering...

1. Why name `run` instead of `wait`?
2. What is supposed to go in the `{}` ? Does there need to be a loop here or can it just be a function which blocks until either a timeout or `SIGINT` ?
3. Would it be cleaner to define the callback at this call?

```rust
waitset.wait(|attachment_id: AttachmentId<ipc::Service>| {
        if let Some((service_name, listener)) = listener_attachments.get(&attachment_id) {
            print!("Received trigger from \"{}\" ::", service_name);

            while let Ok(Some(event_id)) = listener.try_wait_one() {
                print!(" {:?}", event_id);
            }

            println!("");
        });
        
        // If we get here, either it timed-out (if timed wait) or there was a `SIGINT`.
```

iceoryx2/src/port/waitset.rs
```diff
@@ -0,0 +1,718 @@
...
+//!
+//! let publisher_1 = pubsub.publisher_builder().create()?;
+//! let publisher_2 = pubsub.publisher_builder().create()?;
+//!
+//! let pub_1_period = Duration::from_millis(250);
+//! let pub_2_period = Duration::from_millis(718);
+//!
+//! let waitset = WaitSetBuilder::new().create::<ipc::Service>()?;
+//! let guard_1 = waitset.attach_tick(pub_1_period)?;
+//! let guard_2 = waitset.attach_tick(pub_2_period)?;
```
elBoberido, line 125: Food for thought. Would one attach an interval and get a tick? Similar to attaching a deadline and getting a timeout.

elBoberido (commented):
Intermediate result

iceoryx2/src/port/waitset.rs
orecham: Yes this is not a "tick"...

examples/rust/event_multiplexing/wait.rs
elBoberido: Regarding 1)
Further down there is an example with
```rs
while waitset.run(event_handler).is_ok() {}
```
That was a proposal which @elfenpiff liked so it was changed. The intention was to indicate that it's not just a blocking wait call but that actually something is executed.

orecham: To me, having something like:

```rust
loop {
    waitset.wait(...); // or timed_wait
}
```

is more intuitive.

elfenpiff: 1. @elBoberido wanted to have this name and I had no objections, I also have no objections to `wait`. Sounds like a duel between the two of you ... I will sit here and just watch ;)
2. Actually nothing. It is in the end the main loop of the program and the actual code goes into the callback. It looks a bit strange and the iceoryx1 waitset returned here a vector. This has the problem that we need to overallocate since we do not know beforehand how many events we received but with the callback approach we delegate this to the user.
3. Nope, this would produce a clippy warning. It is always in favor of declaring the lambda first and then provide it as an argument.

iceoryx2/src/port/waitset.rs
elfenpiff: So would you all be happy with `attach_interval` and whenever it comes it is a tick? or a wakeup?

orecham: Sounds more natural to me. The confusion is that when I see tick, I think of the smallest individual increment of a clock.

examples/rust/event_multiplexing/wait.rs
elBoberido: No strong feelings about the name.
How would you call the thingie in the wait call?

orecham: @elBoberido https://github.com/eclipse-iceoryx/iceoryx2/pull/428#discussion_r1784523618

elBoberido: @orecham sorry, I don't get it. Keep in mind, it's already Friday evening. Don't expect too much cognitive capacity left 😅

orecham: @elBoberido I mean, the argument, passed into the callback, is fine to be some ID, like the current `AttachmentId`.

The caller should track which IDs are which component they have attached, then depending on the id provided to the callback, react on the wake-up.

elBoberido: Sorry, I was notclear enough. I didn't meant the parameter of the callback but the callback itself
```rs
fn foo(attachement_id: ...) {
    // ...
}
// ...
waitset.wait(foo);
```
How would you call `foo`?
Calling it just `callback` is similar to calling an `i64` just `integer`. Which is not wrong but also does not give any semantic information.
```rs
fn add(&mut self, integer: i64) {...}
// vs
fn add(&mut self, length: i64) {...}
```

orecham: @elBoberido I didn't really want to name it anything, hence my suggestion (3) above, but this seems not feasible.

It's just a chunk of logic to execute when woken up by an event, so...

`on_wakeup` ? `on_event` ? `on_activation` ? `handler` ?

iceoryx2/tests/waitset_tests.rs
```diff
@@ -0,0 +1,466 @@
...
+            }
+        })
+        .unwrap();
+
+        assert_that!(listener_1_triggered, eq true);
+        assert_that!(receiver_1_triggered, eq true);
+    }
+
+    #[test]
+    fn run_with_tick_blocks_for_at_least_timeout<S: Service>()
```
elBoberido, line 219:
```suggestion
    fn run_with_tick_interval_blocks_for_at_least_timeout<S: Service>()
```

iceoryx2/tests/waitset_tests.rs
```diff
@@ -0,0 +1,466 @@
...
+    where
+        <S::Event as Event>::Listener: SynchronousMultiplexing,
+    {
+        let _watchdog = Watchdog::new();
+        let node = NodeBuilder::new().create::<S>().unwrap();
+        let sut = WaitSetBuilder::new().create::<S>().unwrap();
+
+        let (listener, _) = create_event::<S>(&node);
+        let _guard = sut.attach_notification(&listener);
+        let tick_guard = sut.attach_tick(TIMEOUT).unwrap();
```
elBoberido, line 229: While seeing it in action, some food for thought
```suggestion
        let tick_guard = sut.attach_tick_interval(TIMEOUT).unwrap();
```

iceoryx2/tests/waitset_tests.rs
```diff
@@ -0,0 +1,466 @@
...
+    fn run_with_deadline_blocks_for_at_least_timeout<S: Service>()
+    where
+        <S::Event as Event>::Listener: SynchronousMultiplexing,
+    {
+        let _watchdog = Watchdog::new();
+        let node = NodeBuilder::new().create::<S>().unwrap();
+        let sut = WaitSetBuilder::new().create::<S>().unwrap();
+
+        let (listener, _) = create_event::<S>(&node);
+        let guard = sut.attach_deadline(&listener, TIMEOUT).unwrap();
```
elBoberido, line 254: Same idea here, although it becomes a little bit long and I'm not even sure I like it. 
```suggestion
        let guard = sut.attach_deadline_interval(&listener, TIMEOUT).unwrap();
```

elBoberido (commented):
It seems the only thing left complaining about is naming :)

examples/rust/event_multiplexing/wait.rs
elfenpiff: I went with `on_event`.

iceoryx2-pal/posix/src/macos/select.rs
```diff
@@ -23,7 +23,16 @@ pub unsafe fn select(
     errorfds: *mut fd_set,
     timeout: *mut timeval,
 ) -> int {
-    crate::internal::select(nfds, readfds, writefds, errorfds, timeout)
+    // mac os has some kind of undocument maximum for the timeout, so if the timeout exceeds
+    // 100 years we replace it with a nullptr indicating to wait forever.
+    // If this is wrong - I don't care but I allow you to come to my grave and curse my corps.
+    // On Thursday, the 5th of October 2124 I have time, bring cookies and Earl Grey hot but
+    // with milk.
```
elBoberido: :rofl: :rofl: :rofl:

iceoryx2-bb/posix/tests/file_descriptor_set_tests.rs
```diff
@@ -155,6 +155,36 @@ fn file_descriptor_set_timed_wait_works() {
     assert_that!(result[0], eq unsafe{sut_receiver.file_descriptor().native_handle()});
 }
 
+#[test]
+fn file_descriptor_set_blocking_wait_works() {
```
elBoberido, line 159: The name is a bit ambiguous. I guess you mean that a call to `blocking_wait` immediately returns with data?

iceoryx2-pal/posix/src/windows/select.rs
```diff
@@ -25,8 +27,39 @@ pub unsafe fn select(
...
+    } else {
+        *timeout
+    };
+    let mut timeout_as_duration =
+        Duration::from_secs(timeout.tv_sec as _) + Duration::from_micros(timeout.tv_usec as _);
+
+    loop {
+        let (num_handles, _) = win32call! { winsock windows_sys::Win32::Networking::WinSock::select(nfds, readfds, writefds, errorfds, &mut timeout) };
+        if num_handles > 0 {
+            *timeval = timeout;
```
elBoberido, line 47: Where is the `timeval` parameter declared?

iceoryx2-pal/posix/src/windows/select.rs
```diff
@@ -25,8 +27,39 @@ pub unsafe fn select(
...
+        let elapsed = now.elapsed();
+        if elapsed > timeout_as_duration {
+            *timeval = timeout;
+            return num_handles;
+        } else {
+            timeout_as_duration -= elapsed;
+            timeout.tv_sec = timeout_as_duration.as_secs() as _;
+            timeout.tv_usec = (timeout_as_duration.as_micros()
+                - timeout_as_duration.as_secs() as u128 * 1_000_000)
+                as _;
```
elBoberido, line 60:
```suggestion
            timeout.tv_usec = timeout_as_duration.subsec_micros() as _;
```

iceoryx2/src/port/waitset.rs
```diff
@@ -23,11 +23,12 @@
...
 //!     the user that th *Deadline* has missed its timeout.
 //!     Whenever a *Deadline* receives an event, the timeout is reset.
-//!     One example is a sensor that shall send an update every 100ms. If after 100ms an update
+//!     One example is a sensor that shall send an update every 100ms and the applications requires
+//!     the sensor data latest after 120ms. If after 120ms an update
 //!     is not available the application must wake up and take counter measures. If the update
-//!     arrives already after 78ms, the timeout is reset back to 100ms.
-//! * **Tick** - A cyclic timeout after which the [`WaitSet`](crate::port::waitset::WaitSet)
-//!     wakes up and informs the user that the timeout has passed by providing a tick.
+//!     arrives already after 78ms, the timeout is reset back to 120ms.
```
elBoberido, line 29:
```suggestion
//!     within the timeout, the timeout is reset back to 120ms.
```

elBoberido (approved):

Issue #213, opened 2024-05-16
Increase test coverage


elfenpiff:
## Brief feature description

See title.

Issue #436, opened 2024-10-06
Pub/sub example creates unkillable process with 100% core usage and no output when switched to use string


Cypher1:
## r.e. FAQ
I've seen the 'dynamically sized' portion of the FAQ. This is not about the ability to transmit variably sized data types. This issue is about how badly iceoryx handles dynamically sized data.
Ideally this would be prevented at compile time, but at minimum the program should respond to SIGKILL/SIGTERM and should probably crash when failing to handle dynamically sized data, rather than spinning at 100% core usage.

## Required information

**Operating system:**

* OS name, version: Ubuntu 24.04.1 LTS
* Additionally, on Linux, Mac Os, Unix, output of: `uname -a`: Linux HOSTNAME 6.8.0-45-generic # 45-Ubuntu SMP PREEMPT_DYNAMIC Fri Aug 30 12:02:04 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux

**Rust version:**
rustc 1.80.1 (3f5fd8dd4 2024-08-06)

**Cargo version:**
cargo 1.80.1 (376290515 2024-07-16)

**iceoryx2 version:**
v0.4.1

**Detailed log output:**
I think this is
a) easy to reproduce and
b) is worth reproducing, so I have not collected log data myself.

**Observed result or behaviour:**
./sub runs at 100% core usage and does not respond to attempts to kill it (e.g. SIGKILL, SIGTERM). It has to be SIGSTOP'd and abandoned.

pub.rs
```
use std::process::ExitCode;
use core::time::Duration;
use iceoryx2::prelude::*;

const CYCLE_TIME: Duration = Duration::from_secs(1);

fn main() -> ExitCode {
    let mut count = 0;
    let node = NodeBuilder::new().create::<ipc::Service>().expect("1");

    // create our port factory by creating or opening the service
    let service = node.service_builder(&"My/Funk/ServiceName4".try_into().expect("2"))
        .publish_subscribe::<String>()
        .open_or_create().expect("3");

    let publisher = service.publisher_builder().create().expect("4");

    while let NodeEvent::Tick = node.wait(CYCLE_TIME) {
        let sample = publisher.loan_uninit().expect("5");
        let sample = sample.write_payload(format!("{:?}", count));
        count += 1;
        sample.send().expect("6");
    }
    ExitCode::SUCCESS
}
```

sub.rs
```
use std::process::ExitCode;
use core::time::Duration;
use iceoryx2::prelude::*;

[...]

PR #439, opened 2024-10-06
[#213] test: add test cases

xieyuschen:
## Notes for Reviewer

I have added some test cases during I'm learning the codes. It improves out test.

1. [X] Add sensible notes for the reviewer
1. [X] PR title is short, expressive and meaningful
1. [X] Relevant issues are linked in the [References](#references) section
1. [X] Every source code file has a copyright header with `SPDX-License-Identifier: Apache-2.0 OR MIT`
1. [X] Branch follows the naming format (`iox2-123-introduce-posix-ipc-example`)
1. [X] Commits messages are according to this [guideline][commit-guidelines]
    - [X] Commit messages have the issue ID (`[#123] Add posix ipc example`)
    - [X] Commit author matches [Eclipse Contributor Agreement][eca] (and ECA is signed)
1. [X] Tests follow the [best practice for testing][testing]
1. [X] Changelog updated [in the unreleased section][changelog] including API breaking changes
1. [x] Assign PR to reviewer

## References

Updates #213

### DIFF
--- a/iceoryx2/src/service/static_config/message_type_details.rs
+++ b/iceoryx2/src/service/static_config/message_type_details.rs
@@ -20,9 +20,24 @@ use serde::{Deserialize, Serialize};
 #[derive(Default, Debug, Clone, Eq, Hash, PartialEq, Serialize, Deserialize)]
 pub enum TypeVariant {
     #[default]
-    /// A fixed size type like [`u64`]
+    /// A type notated by [`#[repr(C)]`](https://doc.rust-lang.org/reference/type-layout.html#reprc).
+    /// with a constant size known at compile time is recognized as FixedSize.
+    /// The FixedSize type should satisfy the [`Sized`].
+    /// For example, all primitive types are FixedSize. The self-contained structs(without pointer members
+    /// or heap-usages) are FixedSize.
     FixedSize,
-    /// A dynamic sized type like a slice
+
+    /// A dynamic sized type strictly refers to the slice of an iceoryx2 compatible types.
+    /// The struct with pointer members or with heap usage MUSTN't be recognized as Dynamic type.
+    /// Indeed, they're the in-compatible iceoryx2 types.
+    ///
+    /// The underlying reason is the shared memory which we use to store the payload data.
+    /// If the payload type would use the heap then the type would use
+    /// process local memory that is not available to another process.
+    ///
+    /// The pointer requirement comes again from shared memory.
+    /// It has a different pointer address offset in every process rendering any absolute pointer
+    /// useless and dereferencing it would end up in a segfault.
     Dynamic,
 }
 
@@ -33,9 +48,10 @@ pub struct TypeDetail {
     pub variant: TypeVariant,
     /// Contains the output of [`core::any::type_name()`].
     pub type_name: String,
-    /// The size of the underlying type.
+    /// The size of the underlying type calculated by [`core::mem::size_of`].
     pub size: usize,
-    /// The alignment of the underlying type.
+    /// The ABI-required minimum alignment of the underlying type calculated by [`core::mem::align_of`].
+    /// It may be set by users with a larger alignment, e.g. the memory provided by allocator used by SIMD.
     pub alignment: usize,
 }
 
@@ -64,11 +80,11 @@ pub struct MessageTypeDetails {
 }
 
 impl MessageTypeDetails {
-    pub(crate) fn from<Header, UserHeader, Payload>(variant: TypeVariant) -> Self {
+    pub(crate) fn from<Header, UserHeader, Payload>(payload_variant: TypeVariant) -> Self {
         Self {
             header: TypeDetail::__internal_new::<Header>(TypeVariant::FixedSize),
             user_header: TypeDetail::__internal_new::<UserHeader>(TypeVariant::FixedSize),
-            payload: TypeDetail::__internal_new::<Payload>(variant),
+            payload: TypeDetail::__internal_new::<Payload>(payload_variant),
         }
     }
 
@@ -78,6 +94,7 @@ impl MessageTypeDetails {
         payload_start as *const u8
     }
 
+    /// returns the pointer to the user header
     pub(crate) fn user_header_ptr_from_header(&self, header: *const u8) -> *const u8 {
         let header = header as usize;
         let user_header_start = align(header + self.header.size, self.user_header.alignment);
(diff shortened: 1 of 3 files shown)


xieyuschen:
oops, I have found that alignment in 32bits is 4, not 8. let me change it:)

iceoryx2/src/service/static_config/message_type_details.rs
```diff
@@ -20,9 +20,10 @@ use serde::{Deserialize, Serialize};
 #[derive(Default, Debug, Clone, Eq, Hash, PartialEq, Serialize, Deserialize)]
 pub enum TypeVariant {
     #[default]
-    /// A fixed size type like [`u64`]
+    /// A fixed size type which is determined during compile.
+    /// A structure holds some pointer fields should be clarified as Dynamic.
```
elfenpiff, line 24: Oh this would be wrong. With dynamic we mean a slice `[MY_TYPE]` nothing else. 

Yesterday, someone already run into the issue that our payload type restrictions was not clearly documented, see #436 . 
So, any payload type that you want to transmit with iceoryx2 must be:
 * self-contained, no heap is allowed
 * cannot contain pointer members
 * must be annotated with `#[repr(C)]`

The underlying reason is the shared memory which we use to store the payload data. If the payload type would use the heap then the type would use process local memory that is not available to another process.

The pointer requirement comes again from shared memory. It has a different pointer address offset in every process rendering any absolut pointer useless and dereferencing it would end up in a segfault.

The `#[repr(C)]` requirement prevents the rust compiler of reordering the members of a struct. Just think of that two processes use the same struct but have a different member order, this would result again in undefined behavior.

And dynamic means that you can use a slice of iceoryx2 compatible types.

iceoryx2/src/service/static_config/message_type_details.rs
```diff
@@ -20,9 +20,10 @@ use serde::{Deserialize, Serialize};
...
     FixedSize,
-    /// A dynamic sized type like a slice
+    /// A dynamic sized type like a slice or it contains some pointer fields.
```
elfenpiff, line 26: This must be adjusted as well.

iceoryx2/src/service/static_config/message_type_details.rs
```diff
@@ -33,9 +34,9 @@ pub struct TypeDetail {
     pub variant: TypeVariant,
     /// Contains the output of [`core::any::type_name()`].
     pub type_name: String,
-    /// The size of the underlying type.
+    /// The size of the underlying type calculated by [`core::mem::size_of`].
     pub size: usize,
-    /// The alignment of the underlying type.
+    /// The ABI-required minimum alignment of the underlying type calculated by [`core::mem::align_of`].
```
elfenpiff, line 39: Just for completeness, the user is allowed to override this and to set a larger alignment with:

```rust
 let service = node
     .service_builder(&"My/Funk/ServiceName".try_into()?)
     .publish_subscribe::<TransmissionData>()
     .payload_alignment(256)
```

The idea is that the allocator then provides memory that can also be used for SIMD use cases where the input data often requires a larger alignment.

iceoryx2/src/service/static_config/message_type_details.rs
```diff
@@ -78,6 +79,7 @@ impl MessageTypeDetails {
         payload_start as *const u8
     }
 
+    /// returns pointer of user_header which is the field follows the header field in a structure.
```
elfenpiff, line 82: I wouldn't state this so explicitly since this is not a contract. I think `returns the pointer to the user header` would be just fine.

iceoryx2/src/service/static_config/message_type_details.rs
```diff
@@ -120,3 +122,124 @@ impl MessageTypeDetails {
...
+            payload: TypeDetail{
+                variant: TypeVariant::FixedSize,
+                type_name: "iceoryx2::service::static_config::message_type_details::tests::test_from::MyPayload".to_string(),
+                size: 16,
+                alignment: 8,
+            },
+        };
+        assert_that!(want, eq got);
+
+        let got = MessageTypeDetails::from::<i32, bool, String>(TypeVariant::Dynamic);
```
elfenpiff, line 162: Using `String` in iceoryx2 would lead to a segmentation fault. But you could use `[MyPayload]` as dynamic slice.

iceoryx2/src/service/static_config/message_type_details.rs
```diff
@@ -120,3 +122,124 @@ impl MessageTypeDetails {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use iceoryx2_bb_testing::assert_that;
+
+    #[test]
+    fn test_from() {
+        struct MyPayload {
```
elfenpiff, line 133: Must be annotated with `#[repr(C)]` - in one binary it would not make any difference but just for completeness.

iceoryx2/src/service/static_config/message_type_details.rs
```diff
@@ -120,3 +122,124 @@ impl MessageTypeDetails {
...
+                alignment: 8,
+            },
+        };
+        assert_that!(want, eq got);
+    }
+
+    #[test]
+    fn test_user_header_ptr_from_header() {
+        let details = MessageTypeDetails::from::<i32, bool, String>(TypeVariant::Dynamic);
+        struct Demo {
```
elfenpiff, line 189: Please add `#[repr(C)]`. Here it is again important, otherwise the compiler may reorder the member variables and then the pointer calculations are wrong.

iceoryx2/src/service/static_config/message_type_details.rs
```diff
@@ -120,3 +122,124 @@ impl MessageTypeDetails {
...
+        let ptr: *const u8 = &demo.header as *const _ as *const u8;
+        let user_header_ptr = details.user_header_ptr_from_header(ptr);
+        let got: *const i32 = user_header_ptr as *const i32;
+        assert_that!(unsafe { *got } , eq demo.user_header);
+    }
+
+    #[test]
+    fn test_payload_ptr_from_header() {
+        let details = MessageTypeDetails::from::<i32, i32, i32>(TypeVariant::Dynamic);
+        struct Demo {
```
elfenpiff, line 228: Please add `#[repr(C)]`

iceoryx2/src/service/static_config/message_type_details.rs
```diff
@@ -120,3 +122,124 @@ impl MessageTypeDetails {
...
+
+    #[test]
+    fn test_from() {
+        struct MyPayload {
+            _a: i32,
+            _b: bool,
+            _c: i64,
+        }
+
+        let got = MessageTypeDetails::from::<i32, i64, MyPayload>(TypeVariant::FixedSize);
```
elfenpiff, line 139: Could, you please name it `sut` instead of `got`. SUT = System Under Test. And `want` to `expected`. This is how we usually name the variables in tests - then it would be more consistent.

In my opinion it also reads better:

```rust
assert_that!(sut, eq expected);
```

iceoryx2/tests/service_static_config_tests.rs
```diff
@@ -0,0 +1,53 @@
...
+
+#[cfg(test)]
+mod service_static_config_message_type_details {
+    use iceoryx2::service::static_config::message_type_details::{TypeDetail, TypeVariant};
+    use iceoryx2_bb_testing::assert_that;
+
+    #[test]
+    fn test_internal_new() {
+        struct Tmp;
+        let got = TypeDetail::__internal_new::<Tmp>(TypeVariant::FixedSize);
```
elfenpiff, line 21: Please rename `got` and `want` here also to `sut` and `expected`

elfenpiff (commented):
@xieyuschen It is always a pleasure to start the week with an Earl Grey Hot (with milk) and a review of your PR.

Thank you very much for your continuous effort and helping us out!

xieyuschen:
> @xieyuschen It is always a pleasure to start the week with an Earl Grey Hot (with milk) and a review of your PR.
> 
> Thank you very much for your continuous effort and helping us out!

ack, will try to resolve it later today and thanks for your clarify.

iceoryx2/src/service/static_config/message_type_details.rs
xieyuschen: understood, btw not sure you are refering to the usage of `[MyPayload]` like this:
```rust
MessageTypeDetails::from::<i32, bool, &[MyPayload]>(TypeVariant::Dynamic);
```

> Using String in iceoryx2 would lead to a segmentation fault

I think we can try our best to reject the some types such as string in `MessageTypeDetails::from`, such as panic.
Not sure we can have some refinements here, but i will have a look later. How do you think about it?

iceoryx2/tests/service_static_config_tests.rs
xieyuschen: get it, but could i know what's the meaning of `sub`, i haven't thought a useful explanation.

iceoryx2/src/service/static_config/message_type_details.rs
elfenpiff: @xieyuschen The usage would be like:

```rust
MessageTypeDetails::from::<i32, bool, [MyPayload]>(TypeVariant::Dynamic);
```

> I think we can try our best to reject the some types such as string in MessageTypeDetails::from, such as panic.
Not sure we can have some refinements here, but i will have a look later. How do you think about it?

The idea was to provide a trait `ShmSend` and a corresponding proc macro. Every type that implements `ShmSend` can be used as payload type. We provide `ShmSend` for all POD types (u64, u32, bool ...) and all shared memory compatible containers from `iceoryx2_bb_containers`. The proc macro would also add `#[repr(C)]`.

xieyuschen: cool, that's great, could I know the issue to track this change?

xieyuschen: hi @elfenpiff looks your example here doesn't work. but never mind, I can skip it in this PR and test it in the following PR. 

```
        #[repr(C)]
        struct MyPayload {
            header: i32,
            _user_header: i32,
            payload: i32,
        }
        let details = MessageTypeDetails::from::<i32, i32, [MyPayload]>(TypeVariant::Dynamic);
```

```
error[E0277]: the size for values of type `[service::static_config::message_type_details::tests::test_payload_ptr_from_header::MyPayload]` cannot be known at compilation time
   --> iceoryx2/src/service/static_config/message_type_details.rs:283:60
    |
283 |         let details = MessageTypeDetails::from::<i32, i32, [MyPayload]>(TypeVariant::Dynamic);
    |                                                            ^^^^^^^^^^^ doesn't have a size known at compile-time
```

xieyuschen:
Hi @elfenpiff , I have updated the PR based on your comments. Thanks for reviewing. 

I resolved the straitfoward comments and left some comments for you to see whether you have concerns about it.

iceoryx2/tests/service_static_config_tests.rs
```diff
@@ -0,0 +1,60 @@
...
+        };
+
+        assert_that!(sut, eq expected);
+
+        let sut = TypeDetail::__internal_new::<TypeDetail>(TypeVariant::FixedSize);
+        let expected = TypeDetail {
+            variant: TypeVariant::FixedSize,
+            type_name: "iceoryx2::service::static_config::message_type_details::TypeDetail"
+                .to_string(),
+            size: 48,
```
elfenpiff, line 51: Either you use here `size_of::<TypeDetail>()` or need to distinguish between target pointer width 32 and 64.

elfenpiff (commented):
I left a comment on how to fix the ubuntu 32 bit issue. When this is fixed we can merge this PR.

iceoryx2/tests/service_static_config_tests.rs
xieyuschen: use `size_of::<TypeDetail>()` instead, i haven't checked the ci again today so i missed the failure pipeline. fixed it and I have rebased them.

xieyuschen:
> @xieyuschen fix the little CI failure with `size_of` and we can get it merged.

sure, i haven't encountered it in my local so i have missed it. but nvm, now it should be fine:)

elfenpiff (approved):

PR #448, opened 2024-10-07
[#436] Split up the sig handler into non fatal and fatal signals

elfenpiff:
## Notes for Reviewer

The `SignalHandler` registers in the beginning every signal to keep track of them so that the user can later acquire them via an internal atomic variable. The problem is, that by default it also registers `SIGSEGV` and other fatal signals. If such a signal is emitted in turns into an infinite loop where the signal callback is called.

This can be mitigated by splitting up the FetchableSignals into FatalFetchableSignals (e.g. `SIGSEGV`, ...) and NonFatalFetchableSignals (`SIGINT`, `SIGUSR1` ... ).
Certain operations are only allowed for NonFatalFetchableSignals, like acquiring the last signal that was emitted. Also the SignalHandler only registers NonFatalFetchableSignals by default so that the user can later acquire the signal that was emitted. For FatalFetchableSignals it would not make sense since then the callback of the signal is the last thing the program executes.

1. [x] Assign PR to reviewer

## References

Relates to #436

### DIFF
--- a/iceoryx2-bb/posix/src/signal.rs
+++ b/iceoryx2-bb/posix/src/signal.rs
@@ -40,7 +40,7 @@
 //!
 //! fn some_task() {}
 //!
-//! while SignalHandler::last_signal() != Some(FetchableSignal::Terminate) {
+//! while SignalHandler::last_signal() != Some(NonFatalFetchableSignal::Terminate) {
 //!     some_task();
 //! }
 //! ```
@@ -50,7 +50,7 @@
 //! ```no_run
 //! use iceoryx2_bb_posix::signal::*;
 //!
-//! SignalHandler::wait_for_signal(FetchableSignal::Terminate);
+//! SignalHandler::wait_for_signal(NonFatalFetchableSignal::Terminate);
 //! ```
 use std::{
     fmt::{Debug, Display},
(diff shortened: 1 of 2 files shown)


iceoryx2-bb/posix/src/signal.rs
```diff
@@ -75,32 +75,54 @@ use tiny_fn::tiny_fn;
...
+            $unknown_entry = $unknown_nn::$unknown_value
+          entry:
+            $($entry = $nn::$value),*
+        }
+
+        /// Represents signals that can be fetched and are fatal, meaning that they
+        /// indicate that the process is somehow corrupted and should terminate.
+        #[derive(Debug, Clone, Copy, Eq, Hash, PartialEq, Sequence)]
+        #[repr(i32)]
+        pub enum FatalFetchableSignal {
```
orecham, line 104: Nothing to change, but just out of curiosity - why the distinction of being `Fetchable` ? What is a `NonFetchableSignal` ?

orecham (approved):

iceoryx2-bb/posix/src/signal.rs
elfenpiff: For `NonFetchableSignal` you cannot register a custom callback via `sigaction`. For instance `SIGKILL` does not send anything to the process but the OS directly kills the process without informing it.

PR #451, opened 2024-10-07
[#390] waitset c binding

elfenpiff:
## Notes for Reviewer

The language bindings for C and C++. The C API is tested through the C++ API, which utilizes it fully.
The `event-multiplexing` example is not yet ported - this comes later.

1. [x] Assign PR to reviewer

## References

Relates to #390


iceoryx2-ffi/cxx/include/iox2/waitset.hpp
```diff
@@ -0,0 +1,222 @@
...
+#include "iox2/service_type.hpp"
+#include "iox2/waitset_enums.hpp"
+
+namespace iox2 {
+/// The [`Guard`] is returned by [`WaitSet::attach_deadline()`], [`WaitSet::attach_notification()`]
+/// or [`WaitSet::attach_interval()`]. As soon as it goes out-of-scope it detaches the attachment.
+/// It can also be used to determine the origin of an event in [`WaitSet::run()`] or
+/// [`WaitSet::run_once()`] via [`AttachmentId::event_from()`] or [`AttachmentId::deadline_from()`].
+template <ServiceType S>
+class Guard {
```
orecham, line 30: Is this name to generic ? Wondering if other types of guards will be needed within the `iox2` namespace. This could be changed when the need for another type of `Guard` comes though.

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
```diff
@@ -0,0 +1,222 @@
...
+    Guard(Guard&&) noexcept;
+    auto operator=(Guard&& rhs) noexcept -> Guard&;
+    ~Guard();
+
+    Guard(const Guard&) = delete;
+    auto operator=(const Guard&) = delete;
+
+  private:
+    template <ServiceType>
+    friend class WaitSet;
```
orecham, line 41: Tight coupling to `WaitSet` makes me think the name `WaitSetGuard` (or similar) is warranted.

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
orecham, line 30: Is `Guard` to generic here? Just wondering in case other types of `Guard`s will be needed in the future. Can of course be updated when the need arrises, though.

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
```diff
@@ -0,0 +1,222 @@
...
+    ~AttachmentId();
+
+    /// Creates an [`AttachmentId`] from a [`Guard`] that was returned via
+    /// [`WaitSet::attach_interval()`], [`WaitSet::attach_notification()`] or
+    /// [`WaitSet::attach_deadline()`].
+    static auto from_guard(const Guard<S>& guard) -> AttachmentId;
+
+    /// Returns true if an event was emitted from a notification or deadline attachment
+    /// corresponding to [`Guard`].
+    auto event_from(const Guard<S>& guard) const -> bool;
```
orecham, line 69: Why not `has_event` ? The name `event_from` sounds like an event will be returned.

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
```diff
@@ -0,0 +1,222 @@
...
+
+    /// Returns true if the deadline for the attachment corresponding to [`Guard`] was missed.
+    auto deadline_from(const Guard<S>& guard) const -> bool;
```
orecham, line 72: `has_missed_deadline` ?

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
```diff
@@ -0,0 +1,222 @@
...
+    /// Can be called from within a callback during [`WaitSet::run()`] to signal the [`WaitSet`]
+    /// to stop running after this iteration.
+    void stop();
+
+    /// Waits in an infinite loop on the [`WaitSet`]. The provided callback is called for every
+    /// attachment that was triggered and the [`AttachmentId`] is provided as an input argument to
+    /// acquire the source.
+    /// If an interrupt- (`SIGINT`) or a termination-signal (`SIGTERM`) was received, it will exit
+    /// the loop and inform the user via [`WaitSetRunResult`].
+    auto run(const iox::function<void(AttachmentId<S>)>& fn_call) -> iox::expected<WaitSetRunResult, WaitSetRunError>;
```
orecham, line 121: I really don't like the name `run` (as mentioned in the prior PR). What is "running"? The comment even explains that it is "waiting in an infinite loop". @elBoberido

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
```diff
@@ -0,0 +1,222 @@
...
+
+    /// Tries to wait on the [`WaitSet`]. The provided callback is called for every attachment that
+    /// was triggered and the [`AttachmentId`] is provided as an input argument to acquire the
+    /// source.
+    /// If nothing was triggered the [`WaitSet`] returns immediately.
+    auto run_once(const iox::function<void(AttachmentId<S>)>& fn_call) -> iox::expected<void, WaitSetRunError>;
```
orecham, line 127: Comment: "Tries to wait".

Name: `run_once`

??

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
```diff
@@ -0,0 +1,222 @@
...
+    /// Attaches a [`Listener`] as notification to the [`WaitSet`]. Whenever an event is received on the
+    /// object the [`WaitSet`] informs the user in [`WaitSet::run()`] to handle the event.
+    /// The object cannot be attached twice and the
+    /// [`WaitSet::capacity()`] is limited by the underlying implementation.
+    ///
+    /// # Safety
+    ///
+    /// * The [`Listener`] must life at least as long as the returned [`Guard`].
+    /// * The [`Guard`] must life at least as long as the [`WaitsSet`].
+    auto attach_notification(const Listener<S>& listener) -> iox::expected<Guard<S>, WaitSetAttachmentError>;
```
orecham, line 147: Again naming... sorry for being pedantic... But `attach_notification` "Attaches a [`Listener`]" is inconsistent.

The usage seems to be "attach listener" then "get notification".

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
```diff
@@ -0,0 +1,222 @@
...
+    /// Attaches a [`FileDescriptorView`] as notification to the [`WaitSet`]. Whenever an event is received on the
+    /// object the [`WaitSet`] informs the user in [`WaitSet::run()`] to handle the event.
+    /// The object cannot be attached twice and the
+    /// [`WaitSet::capacity()`] is limited by the underlying implementation.
+    ///
+    /// # Safety
+    ///
+    /// * The corresponding [`FileDescriptor`] must life at least as long as the returned [`Guard`].
+    /// * The [`Guard`] must life at least as long as the [`WaitsSet`].
+    auto attach_notification(FileDescriptorView file_descriptor) -> iox::expected<Guard<S>, WaitSetAttachmentError>;
```
orecham, line 158: If the goal was to just have a consistent API for attaching "things", I would just call it `attach` and let the argument inform the user of what is actually being attached.

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
```diff
@@ -0,0 +1,222 @@
...
+    /// deadline is hit, the user is informed in [`WaitSet::run()`].
+    /// The object cannot be attached twice and the
+    /// [`WaitSet::capacity()`] is limited by the underlying implementation.
+    /// Whenever the object emits an event the deadline is reset by the [`WaitSet`].
+    ///
+    /// # Safety
+    ///
+    /// * The corresponding [`Listener`] must life at least as long as the returned [`Guard`].
+    /// * The [`Guard`] must life at least as long as the [`WaitsSet`].
+    auto attach_deadline(const Listener<S>& listener,
```
orecham, line 170: I would again just call this `attach` and let the additional `deadline` argument tell the story.

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
```diff
@@ -0,0 +1,222 @@
...
+    auto attach_deadline(FileDescriptorView file_descriptor,
+                         iox::units::Duration deadline) -> iox::expected<Guard<S>, WaitSetAttachmentError>;
+
+    /// Attaches a tick event to the [`WaitSet`]. Whenever the timeout is reached the [`WaitSet`]
+    /// informs the user in [`WaitSet::run()`].
+    ///
+    /// # Safety
+    ///
+    /// * The [`Guard`] must life at least as long as the [`WaitsSet`].
+    auto attach_interval(iox::units::Duration deadline) -> iox::expected<Guard<S>, WaitSetAttachmentError>;
```
orecham, line 192:
```suggestion
    auto attach(iox::units::Duration interval) -> iox::expected<Guard<S>, WaitSetAttachmentError>;
```

iceoryx2-ffi/cxx/include/iox2/listener.hpp
```diff
@@ -83,6 +83,8 @@ class Listener {
   private:
     template <ServiceType>
     friend class PortFactoryListener;
+    template <ServiceType>
+    friend class WaitSet;
```
orecham, line 87: Just curious, what is this required for?

orecham (commented):
Looks good, comments are mainly for superficial changes. The motivation, however, is to make the API as self-explanatory to future devs as possible by reducing the number of "cognitive steps" required to understand what is being done. 

For example, with `attach_notification`, the question arises "what is a notification?", but this IMO is not information worth keeping track of when working with the WaitSet.

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
elfenpiff: I will rename `Guard` to `WaitSetGuard` and `AttachmentId` to `WaitSetAttachmentId`.

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
elfenpiff: My idea is, that it reads like this: `event_from(my_guard);` so that it is nearly like a sentence. I could rename it to `is_event_from(my_guard)`.

@orecham what do you think? The same would also go for `deadline_from`

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
elfenpiff: @elBoberido I would side here with the native speaker @orecham if you do not have any counter arguments.

So I would rename:
 * `run` -> `wait`
 * `run_once` -> `try_wait`

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
elfenpiff: The notification is defined in the rust documentation of the waitset. See:

* **Notification** - An object that emits an event. Whenever the event is detected the
    [`WaitSet`](crate::port::waitset::WaitSet) wakes up and informs the user.
    Typical use case are gateways, which receives and forwards data whenever new data
    is available.
* **Deadline** - Like a *Notification* with the exception that the *Deadline* expects an
    event after a certain predefined timeout. If the event does not arrive before the
    timeout has passed, the [`WaitSet`](crate::port::waitset::WaitSet) wakes up and informs
    the user that th *Deadline* has missed its timeout.
    Whenever a *Deadline* receives an event, the timeout is reset.
    One example is a sensor that shall send an update every 100ms and the applications requires
    the sensor data latest after 120ms. If after 120ms an update
    is not available the application must wake up and take counter measures. If the update
    arrives within the timeout, the timeout is reset back to 120ms.
* **Interval** - An time period after which the [`WaitSet`](crate::port::waitset::WaitSet)
    wakes up and informs the user that the time has passed by.
    This is useful when a [`Publisher`](crate::port::publisher::Publisher) shall send an
    heartbeat every 100ms.

So it shall read that the listener is attached to the waitset as a notification.

iceoryx2-ffi/cxx/include/iox2/listener.hpp
elfenpiff: The `WaitSet` actually attaches only file descriptors and this allows the `WaitSet` to call `iox2_listener_get_file_descriptor` with the `m_handle`.

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
orecham: For that purpose, I think `has_event_from` is more natural.

```c++
waitset.has_event_from(my_guard);
```

vs

```c++
waitset.is_event_from(my_guard);
```

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
orecham: Is knowing the these details strictly necessary for a user? It seems `Notification` is an internal abstraction for handling the events.

Documentation is good for developers that need to get into the details. My perspective is that if the developer can understand generally what is happening from the API alone, it makes the API easier to use.

My perspective is that the API to describe what the user "wants to do" rather than the how (this is for documentation). The subtleties can be expressed with a combination of the name and the arguments. The details are in the documentation.

Does the user need to know about `Notifications` abstraction here to do what they want? I think no.

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
orecham: I would have named the Rust API the same, for the same reasons. If one of the goals is user-friendliness, we should change it. It can be done separately if you want to focus on the functionality here, but if we do that I worry it will just not get done.

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
orecham: @elBoberido Doesn't the return of a `WaitSetRunResult` tell this story sufficiently?

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
elBoberido: Woul adding an `event` suffix help, e.g. `attach_notification_event` or `attach_notification_event_source`

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
elBoberido: In Rust it would not work due to not supporting function overloading.

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
elfenpiff: The user needs to understand those concepts and the API calls must be different since they are behaving differently, depending on what you attach. If you attach a notification you cannot expect a timeout and spurious wakeups, with deadlines you can.

I would be open to rename the API but they should be different and clearly state notification, deadline or interval.

iceoryx2-ffi/cxx/include/iox2/listener.hpp
elfenpiff: Yes, the user could a lot of stupid things with it and it would restrict us - then every listener would need to be file descriptor based and we have building blocks that are based on a semaphore. This would not work with this API.

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
elBoberido: I don't think so. Playing the devils advocate, it should be `WaitSetWaitResult` to be consistent with the function name :).
But even if the name would remain the same, it think it would not be sufficient on the call site, e.g.
```rs
waitset.wait(|id| { ... })?;
waitset.wait(on_event)?;
waitset.wait(event_handler)?;
```
In this case, only `on_event` would give a hint. But even then, it would be ambiguous since one could assume to wait until the condition in `on_event` becomes true and then return, like with an condition variable. For myself, I wouldn't call a free function `on_event` since that describes how it is used, not what it does. For a parameter name it would be fine, though.
There is also the familiarity with other APIs with just a `wait` fuction. These are usually just blocking and do not perform actions.
I've the feeling, there is still the iceoryx1 WaitSet in our minds, which indeed just blocks and returns a list of events/states. The current WaitSet exceeds this functionality and if the attach methods would have a callback for each event itself, it would be known as an EventDispatcher in the literature. Currently it's something in between an EventDispatcher and what we know of in iceoryx1. I guess this is the reason why it's so difficult to find a proper name.
Also, keep in mind that the cose is read much more often than written, therefore one should optimize for the reader, not the writer of the code

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
orecham: I still can't resolve that "attach notification" is actually attaching a listener. It requires a couple of cognitive jumps to make sense of. I leave it up to you though as to whether you agree or how much you want to focus on this aspect.

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
orecham: @elBoberido Ah right, my brain still lives partially in C++ world.

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
orecham: I can settle for the `wait_and_process_events` middleground.

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
elfenpiff: The original API comes from rust and there it is

```rust
fn attach_notification<T: SynchronousMultiplexing>(value: &T);
```

In the C++ API a notification can be a listener or a file descriptor - there is already the overload for that since the trait approach in C++ would be messy at best.

orecham: @elfenpiff Let's leave it for now, but I don't think it should be called `attach_notification` in the Rust API either. A`Listener` or `FileDescriptor` is attached, for which notifications are received.

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
elBoberido: Some suggestions from chatgpt, in case we want a shorter name
- wait_and_process
- process_events
- await_and_process

iceoryx2-ffi/ffi/src/api/file_descriptor.rs
```diff
@@ -0,0 +1,200 @@
...
+    is_owned: bool,
+}
+
+impl FileDescriptorBased for CFileDescriptor {
+    fn file_descriptor(&self) -> &FileDescriptor {
+        unsafe { core::mem::transmute(self) }
+    }
+}
+
+impl SynchronousMultiplexing for CFileDescriptor {}
```
elBoberido, line 39: This is superfluous and can be removed. Just change
```rs
pub type iox2_file_descriptor_ptr = *const CFileDescriptor;
pub type iox2_file_descriptor_ptr_mut = *mut CFileDescriptor;
```
to
```rs
pub type iox2_file_descriptor_ptr = *const FileDescriptor;
pub type iox2_file_descriptor_ptr_mut = *mut FileDescriptor;
```
and every other usage of `CFileDescriptor` to `FileDescriptor` and it should still work without duplicating the type.

elfenpiff: Nope, the FileDescriptor itself does not implement SynchronousMultiplexing therefore I added a CFileDescriptor that supports it - in an unsafe manner.

iceoryx2-ffi/cxx/include/iox2/waitset.hpp
elfenpiff: Argh, I read it too late.

orecham (approved):

Issue #464, opened 2024-10-14
Create File Logger


elfenpiff:
## Brief feature description

Instead of logging everything to the console, iceoryx2 shall also provide a file logger. It must be threadsafe without slowing down the hot path.

PR #465, opened 2024-10-14
[#464] file logger

elfenpiff:
## Notes for Reviewer

Implements a simple file logger that does not use any additional crates and where we can add a simple C/C++ binding for it to be able to exchange loggers from different languages.

1. [x] Assign PR to reviewer

## References

Relates to #464

### DIFF
--- a/iceoryx2-bb/log/src/logger/file.rs
+++ b/iceoryx2-bb/log/src/logger/file.rs
@@ -41,7 +41,7 @@ use std::{
     time::{Duration, Instant, SystemTime},
 };
 
-use crate::LogLevel;
+use crate::{get_log_level, LogLevel};
 
 struct Entry {
     timestamp: Duration,
@@ -154,6 +154,10 @@ impl crate::Logger for Logger {
         origin: std::fmt::Arguments,
         formatted_message: std::fmt::Arguments,
     ) {
+        if get_log_level() > log_level as u8 {
+            return;
+        }
+
         let mut state = self.state.lock().expect("Acquire internal state mutex.");
         state.buffer.push_back(Entry {
             log_level,
--- a/iceoryx2-bb/log/src/logger/file.rs
+++ b/iceoryx2-bb/log/src/logger/file.rs
@@ -1,0 +1,131 @@
+// Copyright (c) 2024 Contributors to the Eclipse Foundation
+//
+// See the NOTICE file(s) distributed with this work for additional
+// information regarding copyright ownership.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Apache Software License 2.0 which is available at
+// https://www.apache.org/licenses/LICENSE-2.0, or the MIT license
+// which is available at https://opensource.org/licenses/MIT.
+//
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use std::{
+    collections::VecDeque,
+    fmt::Debug,
+    fs::File,
+    io::Write,
+    sync::{Arc, Condvar, Mutex},
+    thread::JoinHandle,
+    time::{Duration, Instant, SystemTime},
+};
+
+use crate::LogLevel;
+
+use super::Logger;
+
+struct Entry {
+    timestamp: Duration,
+    elapsed_time: Duration,
+    log_level: LogLevel,
+    origin: String,
+    message: String,
+}
+
+impl Debug for Entry {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "timestamp: {:?}, elapsed_time: {:?}, log_level: {:?}, origin: {}, message: {}",
+            self.timestamp, self.elapsed_time, self.log_level, self.origin, self.message
+        )
+    }
+}
+
+struct State {
+    buffer: VecDeque<Entry>,
+    keep_running: bool,
+}
+
+impl Default for State {
+    fn default() -> Self {
+        Self {
+            buffer: VecDeque::new(),
+            keep_running: true,
+        }
+    }
+}
+
+pub struct FileLogger {
+    state: Arc<Mutex<State>>,
+    trigger: Arc<Condvar>,
+    start_time: Instant,
+    _background_thread: Arc<Option<JoinHandle<()>>>,
+}
+
+impl FileLogger {
+    pub fn new(file_name: &str) -> Self {
+        let state = Arc::new(Mutex::new(State::default()));
+        let trigger = Arc::new(Condvar::new());
+        let mut file = File::create(file_name).unwrap();
+
+        let self_state = state.clone();
+        let self_trigger = trigger.clone();
+
+        let write_buffer_to_file = move || loop {
+            println!("write buffer to file");
+            let mut lock = state.lock().unwrap();
+
+            lock = trigger
+                .wait_while(lock, |state: &mut State| state.buffer.is_empty())
+                .unwrap();
+
+            while let Some(entry) = lock.buffer.pop_front() {
+                println!("write_entry");
+                file.write_all(format!("{:?}\n", entry).as_bytes()).unwrap();
+            }
+            file.sync_all().unwrap();
+
+            if !lock.keep_running {
+                println!("buffer len {}", lock.buffer.len());
+                break;
+            }
+        };
+
+        Self {
+            state: self_state,
+            trigger: self_trigger,
+            _background_thread: Arc::new(Some(std::thread::spawn(write_buffer_to_file))),
+            start_time: Instant::now(),
+        }
+    }
+}
+
+impl Drop for FileLogger {
+    fn drop(&mut self) {
+        let mut state = self.state.lock().unwrap();
+        state.keep_running = false;
+        self.trigger.notify_one();
+    }
+}
+
+impl Logger for FileLogger {
+    fn log(
+        &self,
+        log_level: LogLevel,
+        origin: std::fmt::Arguments,
+        formatted_message: std::fmt::Arguments,
+    ) {
+        let mut state = self.state.lock().unwrap();
+        state.buffer.push_back(Entry {
+            log_level,
+            timestamp: SystemTime::now()
+                .duration_since(SystemTime::UNIX_EPOCH)
+                .unwrap(),
+            elapsed_time: self.start_time.elapsed(),
+            origin: origin.to_string(),
+            message: formatted_message.to_string(),
+        });
+        self.trigger.notify_one();
+    }
+}
--- a/iceoryx2-bb/log/src/logger/file.rs
+++ b/iceoryx2-bb/log/src/logger/file.rs
@@ -10,10 +10,31 @@
 //
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
+//! # Example
+//!
+//! Using the file logger.
+//!
+//! ```no_run
+//! use iceoryx2_bb_log::{info, set_logger, set_log_level, LogLevel, logger::file};
+//! use std::sync::LazyLock;
+//!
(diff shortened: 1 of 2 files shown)


iceoryx2-bb/log/src/logger/file.rs
```diff
@@ -0,0 +1,169 @@
...
+/// A logger that logs all messages into a file. It implements an active object pattern. A
+/// background thread waits on a queue of log messages and whenever a new message is added.
+pub struct Logger {
+    state: Arc<Mutex<State>>,
+    trigger: Arc<Condvar>,
+    start_time: Instant,
+    _background_thread: Arc<Option<JoinHandle<()>>>,
+}
+
+impl Logger {
```
orecham, line 87: Not really an expert on loggers, but the approach makes sense to me.

The use of a separate thread raised some question marks, though. It adds some jitter, but the tradeoff would add latency to the user thread. Fine for me. Can change if more information later reveals another approach.

orecham (commented):
Looks fine for me, will let @elBoberido comment on the Rust. Ping me if needed.

iceoryx2-bb/log/src/logger/file.rs
```diff
@@ -0,0 +1,169 @@
...
+        let mut state = self.state.lock().expect("Acquire internal state mutex.");
+        state.buffer.push_back(Entry {
+            log_level,
+            timestamp: SystemTime::now()
+                .duration_since(SystemTime::UNIX_EPOCH)
+                .expect("Acquire current system time."),
+            elapsed_time: self.start_time.elapsed(),
+            origin: origin.to_string(),
+            message: formatted_message.to_string(),
+        });
```
elBoberido, line 166: If you want to keep the approach with the lock, please create the `Entry` before acquiring the lock to prevent large offsets in the timestamps to the actual log event.

iceoryx2-bb/log/src/logger/file.rs
```diff
@@ -0,0 +1,169 @@
...
+        Self {
+            buffer: VecDeque::new(),
+            keep_running: true,
+        }
+    }
+}
+
+/// A logger that logs all messages into a file. It implements an active object pattern. A
+/// background thread waits on a queue of log messages and whenever a new message is added.
+pub struct Logger {
```
xieyuschen, line 80: Regards on the log level, currently i saw we use a static log level to do so. But if we have supported the file logger, I believe it's reasonable to spawn multiple loggers which writes to different files(such as errors.log and info.log, debug.log) in the future. 

Probably it's nice to let the logger holds its own log level, it allows us spawn many file loggers with different level.

iceoryx2-bb/log/src/logger/file.rs
```diff
@@ -0,0 +1,169 @@
...
+        origin: std::fmt::Arguments,
+        formatted_message: std::fmt::Arguments,
+    ) {
+        let mut state = self.state.lock().expect("Acquire internal state mutex.");
+        state.buffer.push_back(Entry {
+            log_level,
+            timestamp: SystemTime::now()
+                .duration_since(SystemTime::UNIX_EPOCH)
+                .expect("Acquire current system time."),
+            elapsed_time: self.start_time.elapsed(),
```
xieyuschen, line 163: Just curious, why does the creation time of a logger matters for each piece of log here. Looks like it's not a useful information for each piece of log.

iceoryx2-bb/log/src/logger/file.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+}
+
+/// A logger that logs all messages into a file. It implements an active object pattern. A
+/// background thread waits on a queue of log messages and whenever a new message is added.
+pub struct Logger {
+    state: Arc<Mutex<State>>,
+    trigger: Arc<Condvar>,
+    start_time: Instant,
+    _background_thread: Arc<Option<JoinHandle<()>>>,
```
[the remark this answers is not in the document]
xieyuschen: looks like each logger will create a new thread via `Arc::new(Some(std::thread::spawn(write_buffer_to_file)))`. I believe probably we can have an IO thread to handle all file loggers. 

for now looks fine:) we can consider to improve it in the future.

iceoryx2-bb/log/src/logger/file.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+impl crate::Logger for Logger {
+    fn log(
+        &self,
+        log_level: LogLevel,
+        origin: std::fmt::Arguments,
+        formatted_message: std::fmt::Arguments,
+    ) {
+        let mut state = self.state.lock().expect("Acquire internal state mutex.");
```
xieyuschen, line 157: seems you have missed to compare with global log level, if the input log_level is lower than the static `LOG_LEVEL`, we can do nothing and return immediately.

iceoryx2-bb/log/src/logger/file.rs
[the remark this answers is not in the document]
elfenpiff: When you take a look at the usage example, then you see that the FileLogger is somehow a singleton and every process would come with an additional logging thread when using a FileLogger.

The important thing is, this is not a mandatory thread that always has to run. You can always write your own logger or go back to the console logger that aren't using any threads.

iceoryx2-bb/log/src/logger/file.rs
elfenpiff: But this would still be possible with this approach. You could create a custom `MultiFileLogger` which uses a log file for every log level.

iceoryx2-bb/log/src/logger/file.rs
elfenpiff: * `timestamp` - useful when debugging many processes that log in many log files to have some kind of total order across these log files. But the UNIX time stamp does not really tell you if the thing occurred during the start of a program or after running for hours
* `elapsed_time` - Maybe a better name is needed. Useful when you have a bug that always occurs 10 seconds after starting the process (we already had some of them). Then you start investigating the logs that were created around 10 seconds after process startup.

So the `elapsed_time` relates somehow to the process startup time since the logger is the first thing that is created in a process.

iceoryx2-bb/log/src/logger/file.rs
elfenpiff: Good catch!

iceoryx2-bb/log/src/logger/file.rs
```diff
@@ -0,0 +1,169 @@
...
+                })
+                .expect("Wait for internal state trigger.");
+            drop(lock);
+
+            loop {
+                // acquiring the lock only to get the next buffer to minimize contention as much
+                // as possible.
+                let mut lock = state.lock().expect("Acquire internal state mutex.");
+                let buffer = lock.buffer.pop_front();
+                drop(lock);
```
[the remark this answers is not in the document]
elfenpiff: In the sense of minimizing external dependencies for the core functionality (even when a file logger is not be sooo core) I used the `std::sync::mpsc::channel`. It really improved the code in the sense of complexity - good suggestion!

iceoryx2-bb/log/src/logger/file.rs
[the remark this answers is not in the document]
xieyuschen: Looks like it's intended to use the FileLogger as a singleton, I thought we probably creates multiple loggers by `file::Logger::new(LOG_FILE)` and use them for different purpose.

> this is not a mandatory thread that always has to run.

The point here is if multiple file loggers are created, and they may all log(write files) at the same time, which leads several threads are blocked for IO.  Do all IO here in one thread looks better, but introduce new complexity.

But here we intend to use it as singleton(caller side is also maintained by us), no other concerns.

iceoryx2-bb/log/src/logger/file.rs
```diff
@@ -61,78 +67,43 @@ impl Debug for Entry {
...
+                    .expect("Writing log message into log file."),
+                Ok(Message::Stop) => break,
+                Err(e) => file
+                    .write_all(
+                        format!("[This should never happen!] File Logger got error: {:?}", e)
+                            .as_bytes(),
+                    )
+                    .expect("Write log message into log file."),
+            };
             file.sync_all().expect("Sync log file with disc.");
```
xieyuschen, line 102: will `sync_all` cost a lot for every piece of messages? I have searched the forum mentioned it [in this post](https://users.rust-lang.org/t/patterns-for-flush-sync-in-async-i-o/101563). Hdyt? 

> B) Asking the OS to dump its in-memory ... . it can be a slow and expensive operation.
> This operation is not required for data to become visible to other applications.

elBoberido: I guess the idea was to have the logs available when a crash occurs. Maybe this can be done via a crash handler. On the other side, AFAIK there are no logs in the hot path, so it should also not affect performance.

iceoryx2-bb/log/src/logger/file.rs
```diff
@@ -0,0 +1,146 @@
...
+            start_time: Instant::now(),
+        }
+    }
+}
+
+impl Drop for Logger {
+    fn drop(&mut self) {
+        self.sender
+            .send(Message::Stop)
+            .expect("Send stop notification to background thread.");
```
elBoberido, line 117: Do you also need to wait for the thread to join? I guess not?

iceoryx2-bb/log/src/logger/file.rs
elfenpiff: This is why all of this is handled in a background thread. The hot path just uses the channel to move the log message out of the hot thread into the log thread.

iceoryx2-bb/log/src/logger/file.rs
elfenpiff: No need, the thread will block in drop.

iceoryx2-bb/log/src/logger/file.rs
elfenpiff: We are talking here about single-digit micro-seconds so I think it is fine. This will never be perfect but we try to provide as much insights as possible for the next developer that has to debug something.

iceoryx2-bb/log/src/logger/file.rs
elfenpiff: I think the worst case scenario is, when a crash happens and the logger didn't sync the file and therefore the final log messages are not written.

However, I also want to highlight that this does not need to be the final implementation. If we realize this becomes a bottleneck we can put on our thinking caps and try harder.

iceoryx2-bb/log/src/logger/file.rs
```diff
@@ -0,0 +1,146 @@
...
+                Ok(Message::Entry(entry)) => file
+                    .write_all(format!("{:?}\n", entry).as_bytes())
+                    .expect("Writing log message into log file."),
+                Ok(Message::Stop) => break,
+                Err(e) => file
+                    .write_all(
+                        format!("[This should never happen!] File Logger got error: {:?}", e)
+                            .as_bytes(),
+                    )
+                    .expect("Write log message into log file."),
```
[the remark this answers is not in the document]
elfenpiff: It does exit the loop with a panic. But scenarios where this can happen are extremely rare:
1. Out of memory in storage - therefore further logging is impossible.
2. OS bug
3. Hardware failure while writing to disk

I think all of those scenarios shall be handled with a panic.

xieyuschen (approved):
LGTM, thanks for your clarify.

iceoryx2-bb/log/src/logger/file.rs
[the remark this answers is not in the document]
elBoberido: Oh, I overlooked the `expect`

elBoberido (approved):

Issue #349, opened 2024-08-06
Bazel support for the Rust crates


elBoberido:
## Brief feature description

Currently `foreign_cc` is used to build the C binding and `cc_library` to build the C++ bindings. In order to fully exhaust the capabilities of bazel, the C binding and therefore the Rust code should be build with `rules_rust`.

PR #472, opened 2024-10-16
[#349] Add service name generator to be able to run tests concurrently

### LINKED ISSUE #475: Use an alternative approach to make version available in `PackageVersion`
## Brief feature description

Currently, the `iceoryx2_bb_elementary::PackageVersion` uses the env variables `CARGO_PKG_VERSION_{MAJOR|MINOR|PATCH}` to acquire the crate version. This is essential, since this number is also stored in every shared memory construct to prevent a communication between incompatible iceoryx2 versions. Sadly, those env variables are not set by Bazel and depend on the deployed environment that does not have to be cargo. Therefore, we need a method to set the iceoryx2 version number in code and ensure that it is consistent with `CARGO_PKG_VERSION_**` stuff.

elfenpiff:
## Notes for Reviewer

Fixes some parts of the problem but not everything

1. [x] Assign PR to reviewer

## References

Relates to #349 
Closes #475

### DIFF
--- a/doc/how-to-create-an-iceoryx2-release.md
+++ b/doc/how-to-create-an-iceoryx2-release.md
@@ -79,10 +79,11 @@ number is `Xold.Yold.Zold`.
 10. Adjust the `<version>` to `X.Y.Z` in `$GIT_ROOT$/package.xml`.
 11. Call `rg "Xold\.Yold\.Zold"` and adjust all findings.
     * C and C++ examples, `BUILD.bazel` & `CMakeLists.txt`
-12. **Merge all changes to `main`.**
-13. Set tag on GitHub and add the release document as notes to the tag
+12. Adjust the major, minor and patch version number in `iceoryx2_bb_elementary::PackageVersion`
+13. **Merge all changes to `main`.**
+14. Set tag on GitHub and add the release document as notes to the tag
     description. Add also a link to the file.
-14. Check the order of all dependencies in
+15. Check the order of all dependencies in
     `$GIT_ROOT$/./internal/scripts/crates_io_publish_script.sh`.
     When calling `cargo publish -p $PACKAGE$` all dependencies, also dev-dependencies,
     must be already published to `crates.io` via `cargo publish -p`. Verify the
@@ -92,7 +93,7 @@ number is `Xold.Yold.Zold`.
     * If the publish script was started and a crate requires a dependency which
       is not available on `crates.io` the release has to be redone and the patch
       version has to increase by one for the whole workspace.
-15. Call `$GIT_ROOT$/./internal/scripts/crates_io_publish_script.sh` and publish
+16. Call `$GIT_ROOT$/./internal/scripts/crates_io_publish_script.sh` and publish
     all crates on `crates.io` and `docs.rs`.
-16. Verify that the release looks fine on `docs.rs` (click through the
+17. Verify that the release looks fine on `docs.rs` (click through the
     documentation to check if everything was generated correctly)
--- a/iceoryx2-bb/elementary/src/package_version.rs
+++ b/iceoryx2-bb/elementary/src/package_version.rs
@@ -76,10 +76,17 @@ impl PackageVersion {
                 .and_then(|s| s.parse::<u16>().ok())
                 .unwrap_or(u16::MAX);
 
-            PACKAGE_VERSION.store(
-                PackageVersion::from_version(major, minor, patch).0,
-                Ordering::Relaxed,
-            );
+            if major == 0 && minor == 0 && patch == 0 {
+                PACKAGE_VERSION.store(
+                    PackageVersion::from_version(u16::MAX, u16::MAX, u16::MAX).0,
+                    Ordering::Relaxed,
+                );
+            } else {
+                PACKAGE_VERSION.store(
+                    PackageVersion::from_version(major, minor, patch).0,
+                    Ordering::Relaxed,
+                );
+            }
         }
 
         PackageVersion::from_u64(PACKAGE_VERSION.load(Ordering::Relaxed))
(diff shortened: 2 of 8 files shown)


orecham (dismissed):
Had to do the exact same thing in the RMW 😆 .

elfenpiff:
When `set_log_level(LogLevel::TRACE)` is active one test fails with this:

```
      353 [D] Builder { storage_name: FileName { value: FixedSizeByteString<255> { len: 40, data: "26380db3f2d1d21a17def8273faec2e5097cf3e2" } }, supplementary_size: 0, ha
s_ownership: false, config: Configuration { suffix: FileName { value: FixedSizeByteString<255> { len: 8, data: ".dynamic" } }, prefix: FileName { value: FixedSizeByteStrin
g<255> { len: 5, data: "iox2_" } }, path: Path { value: FixedSizeByteString<255> { len: 14, data: "/tmp/iceoryx2/" } }, _data: PhantomData<iceoryx2::service::dynamic_confi
g::DynamicConfig> }, timeout: 500ms, initializer: , _phantom_data: PhantomData<iceoryx2::service::dynamic_config::DynamicConfig> }
| Failed to open  since the version number was not set - (it is not initialized after 500ms).
```

Which then lead to a failure when opening the dynamic service info and then the builder assumes the services is in a corrupted state.
This comes only up with bazel, never occurs with cmake or with Rust (which has the same test)

elfenpiff:
Okay, there was an actual bug in the dynamic storage posix shared memory implementation. When the creation process is still initializing the memory, the opener takes a COPY of the version number instead of a pointer to it and checks it until it becomes non-zero.

The thing is, this is a bug that should only occur on OSes that do not support adjusting the permissions of the shm during runtime like Mac OS and Windows and not linux.

And this also does not fix the failing test

iceoryx2-bb/elementary/src/package_version.rs
```diff
@@ -76,10 +76,17 @@ impl PackageVersion {
...
                 .unwrap_or(u16::MAX);
 
-            PACKAGE_VERSION.store(
-                PackageVersion::from_version(major, minor, patch).0,
-                Ordering::Relaxed,
-            );
+            if major == 0 && minor == 0 && patch == 0 {
+                PACKAGE_VERSION.store(
+                    PackageVersion::from_version(u16::MAX, u16::MAX, u16::MAX).0,
+                    Ordering::Relaxed,
```
elBoberido, line 82: Would it make sense to just terminate when these variables are not set?

Also, is it possible that there are more env variables which are expected to be set by cargo and pitentially missing from bazel?

elfenpiff: I created an issue: #475 .

We need to refactor this and ensure that we do not depend on deployment env vars since they are obviously not always set as expected.

elfenpiff: Oh Wait ... I can fix it right here.

iceoryx2-bb/elementary/src/package_version.rs
```diff
@@ -63,26 +62,11 @@ impl PackageVersion {
...
-                .unwrap_or(u16::MAX);
-
-            PACKAGE_VERSION.store(
-                PackageVersion::from_version(major, minor, patch).0,
-                Ordering::Relaxed,
-            );
-        }
-
-        PackageVersion::from_u64(PACKAGE_VERSION.load(Ordering::Relaxed))
+        PackageVersion::from_version(MAJOR, MINOR, PATCH)
```
xieyuschen, line 69: just curious, do we have a specific version convention? not sure why we didn't use the semver, which is popular in rust. so in the future we may have `beta`, `alpha` or `rc` versions.

https://crates.io/crates/semver

iceoryx2-ffi/cxx/tests/src/test.hpp
```diff
@@ -31,6 +32,15 @@ using ServiceTypeLocal = TypeServiceType<ServiceType::Local>;
 
 using ServiceTypes = ::testing::Types<ServiceTypeIpc, ServiceTypeLocal>;
 
+inline auto generate_service_name() -> ServiceName {
```
xieyuschen, line 35: looks like it deserves a unit test:)

doc/how-to-create-an-iceoryx2-release.md
```diff
@@ -79,10 +79,11 @@ number is `Xold.Yold.Zold`.
 10. Adjust the `<version>` to `X.Y.Z` in `$GIT_ROOT$/package.xml`.
 11. Call `rg "Xold\.Yold\.Zold"` and adjust all findings.
     * C and C++ examples, `BUILD.bazel` & `CMakeLists.txt`
-12. **Merge all changes to `main`.**
-13. Set tag on GitHub and add the release document as notes to the tag
+12. Adjust the major, minor and patch version number in `iceoryx2_bb_elementary::PackageVersion`
```
elBoberido, line 82: Having a script taking care of setting the version number in all known locations would be helpful. I'll add it to my todo list :)

iceoryx2-bb/elementary/src/package_version.rs
elfenpiff: In general we try to be very cautious to introduce external crates. For any other project I would agree with you but iceoryx2 shall be certified as ASIL D according to ISO26262 which is the highest safety standard in the automotive world.

This requires 100% MC/DC coverage for every line of code. Every test must be covered by a requirement, we need detailed architecture documents and documentation, a safety manual and so on.

So, every dependency we introduce must satisfy all of this! No exception. But actually, nearly none of them does it, and it is a challenge to do it. We have a strategy in place how to solve this but one of the key factors is - try to avoid external crates as much as possible.

Another answer we need to find is, how to handle or prevent supply chain attacks.

So in light of all of this, we try to be very cautious with external dependencies since the more we have the more work we have to do in the near future.

xieyuschen: >  ASIL D according to ISO26262 ... how to handle or prevent supply chain attacks.

understood, quite a bit different from the normal project. thanks for your explaination:)

PR #492, opened 2024-10-28
[#490] Enable loan_slice APIs in CXX bindings

orecham:
## Notes for Reviewer

1. Implement `iox::Slice`
2. Implement `loan_slice` APIs in C++ binding
3. Expose properties `payload_alignment` and `max_slice_len` in C++ binding

## Pre-Review Checklist for the PR Author

1. [x] Add sensible notes for the reviewer
1. [x] PR title is short, expressive and meaningful
1. [x] Relevant issues are linked in the [References](#references) section
1. [x] Every source code file has a copyright header with `SPDX-License-Identifier: Apache-2.0 OR MIT`
1. [x] Branch follows the naming format (`iox2-123-introduce-posix-ipc-example`)
1. [x] Commits messages are according to this [guideline][commit-guidelines]
    - [x] Commit messages have the issue ID (`[#123] Add posix ipc example`)
    - [x] Commit author matches [Eclipse Contributor Agreement][eca] (and ECA is signed)
1. [x] Tests follow the [best practice for testing][testing]
1. [x] Changelog updated [in the unreleased section][changelog] including API breaking changes
1. [x] Assign PR to reviewer
1. [x] All checks have passed (except `task-list-completed`)

[commit-guidelines]: https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
[eca]: http://www.eclipse.org/legal/ECA.php
[testing]: https://github.com/eclipse-iceoryx/iceoryx/blob/master/doc/website/concepts/best-practice-for-testing.md
[changelog]: https://github.com/eclipse-iceoryx/iceoryx2/blob/main/doc/release-notes/iceoryx2-unreleased.md

## Checklist for the PR Reviewer

- [x] Commits are properly organized and messages are according to the guideline
- [x] Unit tests have been written for new behavior
- [x] Public API is documented
- [x] PR title describes the changes

## Post-review Checklist for the PR Author

1. [x] All open points are addressed and tracked via issues

## References

Closes #490

PR #499, opened 2024-11-01
[#498] cleanup message type details

### LINKED ISSUE #498: Cleanup MessageTypeDetails usage

The message type details usage in the internal C API is confusing and needs to be cleaned up

elfenpiff:
## Notes for Reviewer

In this PR the API used when overriding the `MessageTypeDetails` is strictly separated from the end user API.
 * When overriding the `MessageTypeDetails` you must use `[CustomPayloadMarker]` as message type
 * You can only loan samples via `Publisher::loan_custom_payload()`
 * You can only receive samples via `Subscriber::receive_custom_payload()`.

The reason is, that in those cases the `Sample` and `SampleMut` handle the data a bit different. When calling `Publisher::loan_custom_payload(123)` and 123 corresponds to the number of elements, this number is stored in the samples header but `Sample::payload().len()` will return `123 * MessageTypeDetails::payload.size`, the bytes used by the sample. When the `MessageTypeDetails::payload.size` is not overridden, `Sample::payload().len() == Header::number_of_elements()`.

The publish subscribe header contains now the `number_of_elements` that are contained in the sample. The payload layout was removed. In the C API one can now easily obtain the underlying number of elements by just accessing this number directly from the header - no more `payload.len() / sizeof(Payload)` stuff.

1. [x] Assign PR to reviewer

## References

Closes #498

### DIFF
--- a/doc/release-notes/iceoryx2-unreleased.md
+++ b/doc/release-notes/iceoryx2-unreleased.md
@@ -74,10 +74,4 @@
    }
    ```
 
-2. Removed `payload_type_layout` from pub/sub header.
-
-    ```rust
-    // old
-    let sample = publisher.loan()?;
-    sample.header().payload_type_layout();
-    ```
+2. Removed `payload_type_layout()` from `publish_subscribe::Header`.
--- a/doc/release-notes/iceoryx2-unreleased.md
+++ b/doc/release-notes/iceoryx2-unreleased.md
@@ -38,6 +38,7 @@
 * Rename `NodeEvent` into `WaitEvent` [#390](https://github.com/eclipse-iceoryx/iceoryx2/issues/390)
 * Bazel support for the Rust crates [#349](https://github.com/eclipse-iceoryx/iceoryx2/issues/349)
 * Remove ACL dependency [#457](https://github.com/eclipse-iceoryx/iceoryx2/issues/457)
+* Publish Subscribe Header contains number of elements contained in a `Sample` [#498](https://github.com/eclipse-iceoryx/iceoryx2/issues/498)
 
 ### Workflow
 
@@ -72,3 +73,11 @@
     // ...
    }
    ```
+
+2. Removed `payload_type_layout` from pub/sub header.
+
+    ```rust
+    // old
+    let sample = publisher.loan()?;
+    sample.header().payload_type_layout();
+    ```
--- a/iceoryx2-ffi/cxx/include/iox2/header_publish_subscribe.hpp
+++ b/iceoryx2-ffi/cxx/include/iox2/header_publish_subscribe.hpp
@@ -30,8 +30,8 @@ class HeaderPublishSubscribe {
     /// Returns the [`UniquePublisherId`] of the source [`Publisher`].
     auto publisher_id() const -> UniquePublisherId;
 
-    /// Returns the [`Layout`] of the corresponding payload.
-    auto payload_type_layout() const -> iox::Layout;
+    /// Returns the number of [`Payload`] elements in the received [`Sample`].
+    auto number_of_elements() const -> uint64_t;
 
   private:
     template <ServiceType, typename, typename>
--- a/iceoryx2/src/port/publisher.rs
+++ b/iceoryx2/src/port/publisher.rs
@@ -116,6 +116,7 @@ use crate::service::naming_scheme::{
     data_segment_name, extract_publisher_id_from_connection, extract_subscriber_id_from_connection,
 };
 use crate::service::port_factory::publisher::{LocalPublisherConfig, UnableToDeliverStrategy};
+use crate::service::static_config::message_type_details::TypeVariant;
 use crate::service::static_config::publish_subscribe::{self};
 use crate::service::{self, ServiceState};
 use crate::{config, sample_mut::SampleMut};
@@ -139,6 +140,7 @@ use iceoryx2_cal::zero_copy_connection::{
     ZeroCopyConnection, ZeroCopyCreationError, ZeroCopySendError, ZeroCopySender,
 };
 use iceoryx2_pal_concurrency_sync::iox_atomic::{IoxAtomicBool, IoxAtomicU64, IoxAtomicUsize};
+use std::any::TypeId;
 use std::cell::UnsafeCell;
 use std::fmt::Debug;
 use std::sync::atomic::Ordering;
@@ -529,7 +531,11 @@ impl<Service: service::Service> DataSegment<Service> {
 
 /// Sending endpoint of a publish-subscriber based communication.
 #[derive(Debug)]
-pub struct Publisher<Service: service::Service, Payload: Debug + ?Sized, UserHeader: Debug> {
+pub struct Publisher<
+    Service: service::Service,
+    Payload: Debug + ?Sized + 'static,
+    UserHeader: Debug,
+> {
     pub(crate) data_segment: Arc<DataSegment<Service>>,
     dynamic_publisher_handle: Option<ContainerHandle>,
     payload_size: usize,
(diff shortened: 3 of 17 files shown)


doc/release-notes/iceoryx2-unreleased.md
```diff
@@ -72,3 +73,11 @@
...
    }
    ```
+
+2. Removed `payload_type_layout` from pub/sub header.
+
+    ```rust
+    // old
+    let sample = publisher.loan()?;
+    sample.header().payload_type_layout();
+    ```
```
elBoberido, line 83: I think if there is no `new`, the example can be omitted. What do you think?

iceoryx2/src/port/publisher.rs
```diff
@@ -1018,6 +1022,31 @@ impl<Service: service::Service, Payload: Debug, UserHeader: Debug>
...
+        &self,
+        slice_len: usize,
+    ) -> Result<
+        SampleMutUninit<Service, [MaybeUninit<CustomPayloadMarker>], UserHeader>,
+        PublisherLoanError,
+    > {
+        // TypeVariant::Dynamic == slice and only here it makes sense to loan more than one element
+        debug_assert!(slice_len == 1 || self.payload_type_variant() == TypeVariant::Dynamic);
+
+        self.loan_slice_uninit_impl(slice_len, self.payload_size * slice_len)
```
elBoberido, line 1047: I have the feeling there is still not a clear definition what `payload` means and it is used in different contexts with a different meaning, resulting in confusion at some point.

No need to change it now but we need to have unambiguous terminology, else this will haunt us in regular intervals.

iceoryx2/src/service/header/publish_subscribe.rs
```diff
@@ -56,8 +55,16 @@ impl Header {
...
-    pub fn payload_type_layout(&self) -> Layout {
-        self.payload_type_layout
+    /// Returns how many elements are stored inside the sample's payload.
+    ///
+    /// # Details when using
+    /// [`CustomPayloadMarker`](crate::service::builder::publish_subscribe::CustomPayloadMarker)
+    ///
+    /// In this case the number of elements relates to the element defined in the
+    /// [`MessageTypeDetails`](crate::service::static_config::message_type_details::MessageTypeDetails).
+    /// When the element has a `payload.size == 40` and the `Sample::payload().len() == 120` it
```
elBoberido, line 65: This is what I mean, the term `payload` is used in multiple contexts and a comment is necessary to point out what it means. In this case it would make more sense to talk about `element.size` or `item.size`. A payload can consist of one or multiple elements/items and all elements make up the payload size.

iceoryx2/tests/subscriber_tests.rs
```diff
@@ -69,6 +71,31 @@ mod subscriber {
         }
     }
 
+    #[test]
+    #[should_panic]
+    #[cfg(debug_assertions)]
+    fn subscriber_with_custom_payload_details_panics_when_calling_receive<Sut: Service>() {
```
elBoberido, line 77:
```suggestion
    fn subscriber_with_custom_payload_details_panics_when_calling_non_custom_receive<Sut: Service>() {
```

What do you think? I was a bit confused by the name at the beginning, since I thought a condition was missing but it's related to call `receive` instead of `receive_custom_payload`. If you agree, `publisher_with_custom_payload_details_panics_when_calling_loan_slice_uninit` could also be adjusted.

elBoberido (commented):
Looks good. Just some minor questions before merging this.

iceoryx2/src/service/header/publish_subscribe.rs
orecham: I vote for "element"

iceoryx2/src/port/publisher.rs
elfenpiff: Just out of my head:
* payload == all bytes that belong to the user produced data
* slice can be a form of payload
* a slice consists of 1..N elements, therefore we have in this context an element_size?

iceoryx2/src/service/header/publish_subscribe.rs
elfenpiff: I agree, I just speak about `payload.size` since this is the member of `MessageTypeDetails`.

I will refactor it so that it becomes more clear.

elfenpiff: I tried refactoring this, but this will be a huge task and there are a lot of files affected. I think it would be best to have a separate session for this.

>  A payload can consist of one or multiple elements/items and all elements make up the payload size.

This is only true for the sliced API. Otherwise, the element is the payload, and it will lead to confusion when only working with this API.

elfenpiff: Btw. we are discussing here a comment that becomes only relevant for an internal API that the user shall never use.
I acknowledge that this is important and we have to refine the code. But making the code reflecting the terms correctly without increasing the code complexity is a challenge here in my opinion.

orecham: The discussion is about spending some time now to reduce cognitive load for us and other maintainers later.

Although it is internal API, there is still an audience, so clarity still has its merits. In the absolute worst case, we slow down contributions or even scare away potential contributors due to the amount of inside information required.

Leave it up to you to decide if it is worth doing here.

iceoryx2-ffi/cxx/include/iox2/header_publish_subscribe.hpp
```diff
@@ -30,8 +30,8 @@ class HeaderPublishSubscribe {
     /// Returns the [`UniquePublisherId`] of the source [`Publisher`].
     auto publisher_id() const -> UniquePublisherId;
 
-    /// Returns the [`Layout`] of the corresponding payload.
-    auto payload_type_layout() const -> iox::Layout;
+    /// Returns the number of [`Payload`] elements in the received [`Sample`].
+    auto number_of_elements() const -> uint64_t;
```
orecham, line 34: I think the layout is additionally needed here.

With these changes, the code to send slice copies in `iceoryx-ffi` looks like:

```
unsafe fn send_slice_copy<S: Service>(
    publisher: &Publisher<S, PayloadFfi, UserHeaderFfi>,
    data_ptr: *const c_void,
    size_of_element: usize,
    number_of_elements: usize,
    number_of_recipients: *mut usize,
) -> c_int {
    // loan_slice_uninit(1) <= 1 is correct here since it defines the number of
    // slice elements not bytes. The element was set via TypeDetails and has a
    // defined size and alignment.
    let mut sample = match publisher.loan_custom_payload(number_of_elements) {
        Ok(sample) => sample,
        Err(e) => return e.into_c_int(),
    };

    if sample.payload().len() < size_of_element * number_of_elements {
        return iox2_publisher_send_error_e::LOAN_ERROR_EXCEEDS_MAX_LOAN_SIZE as c_int;
    }

    let sample_ptr = sample.payload_mut().as_mut_ptr();
    core::ptr::copy_nonoverlapping(data_ptr, sample_ptr.cast(), size_of_element);
    match sample.assume_init().send() {
        Ok(v) => {
            if !number_of_recipients.is_null() {
                *number_of_recipients = v;
            }
        }
        Err(e) => return e.into_c_int(),
    }

    IOX2_OK
}
```

Focusing on this part:

```
    if sample.payload().len() < size_of_element * number_of_elements {
        return iox2_publisher_send_error_e::LOAN_ERROR_EXCEEDS_MAX_LOAN_SIZE as c_int;
    }
```

This code is errorneous because it does not consider the alignment of the element. There could be an alignment set which requires more bytes than loaned by the publisher.

Alternatively, I guess the alignment could be passed in from as an argument...

orecham: I went with the alternative option, see: https://github.com/eclipse-iceoryx/iceoryx2/pull/492/commits/5411dd69d5c42a34f7585da7f0388b50347ea1d2#diff-1e40fb692713e95d0a2622c1ae500770e1954f70ddec5d2ce1e7bf0a48d40327R227

If this is OK, then nothing to change here.

iceoryx2/src/service/header/publish_subscribe.rs
orecham: @elfenpiff Let's adjust this in a follow-up if required.

orecham (approved):

PR #505, opened 2024-11-07
[#504] slotmap

### LINKED ISSUE #504: Introduce `SlotMap` to manage multiple shared memory data segments
## Brief feature description

To enable the publisher to have multiple data segments, we need to add the data segment id to the pointer. This only works when the segment id is static.
A `SlotMap` is the ideal container to store things with static id's to access/manage them.

elfenpiff:

1. [x] Assign PR to reviewer

## References

Closes #504


iceoryx2-bb/container/src/slotmap.rs
```diff
@@ -0,0 +1,639 @@
...
+            let next = self.idx_to_data_free_list[free_idx].next;
+
+            if next != INVALID {
+                self.idx_to_data_free_list[next].previous = INVALID;
+            }
+            self.idx_to_data_free_list_head = next;
+            Some(free_idx)
+        }
+
+        unsafe fn reserve_index(&mut self, idx: usize) {
```
elBoberido, line 206: This is a bit misleading, since it does not only reserve but also re-use an already used slot. How about `reserve_or_reuse_index`? `claim_index` might also work.

iceoryx2-bb/container/src/slotmap.rs
```diff
@@ -0,0 +1,639 @@
...
+        pub(crate) fn len_impl(&self) -> usize {
+            self.len
+        }
+
+        pub(crate) fn capacity_impl(&self) -> usize {
+            self.idx_to_data.capacity()
+        }
+
+        pub(crate) fn is_empty_impl(&self) -> bool {
+            self.len_impl() == 0
```
elBoberido, line 296: Out of curiosity, since you used just `self.len` further above. Is there a reason to use the method in this case?

iceoryx2-bb/container/src/slotmap.rs
```diff
@@ -0,0 +1,639 @@
...
+/// A compile-time fixed-size, shared memory compatible [`FixedSizeSlotMap`].
+#[repr(C)]
+#[derive(Debug)]
+pub struct FixedSizeSlotMap<T, const CAPACITY: usize> {
+    state: RelocatableSlotMap<T>,
+    _idx_to_data: MaybeUninit<[usize; CAPACITY]>,
+    _idx_to_data_free_list: MaybeUninit<[FreeListEntry; CAPACITY]>,
+    _data: MaybeUninit<[Option<T>; CAPACITY]>,
+    _data_next_free_index: MaybeUninit<[usize; CAPACITY]>,
+}
```
elBoberido, line 537: I think this is something that could easily be created with a proc macro. No so sure about implementing `default` via the proc macro. That might be more difficult but at least the struct members would automatically be kept in sync.

iceoryx2-bb/container/tests/slotmap_tests.rs
```diff
@@ -0,0 +1,174 @@
...
+            let key = SlotMapKey::new(2 * n + 1);
+            assert_that!(sut.insert_at(key, 0), eq true);
+        }
+
+        for _ in 0..SUT_CAPACITY / 2 {
+            let key = sut.insert(0);
+            assert_that!(key, is_some);
+            assert_that!(key.unwrap().value() % 2, eq 0);
+        }
+    }
```
elBoberido, line 152: Please check that another insert does fail instead of using the `insert_at` keys.

iceoryx2-bb/container/tests/slotmap_tests.rs
```diff
@@ -0,0 +1,174 @@
...
+
+        for i in 0..SUT_CAPACITY {
+            keys.push(sut.insert(5 * i + 3).unwrap());
+        }
+
+        for (key, value) in sut.iter() {
+            assert_that!(*value, eq 5 * key.value() + 3);
+        }
+    }
+}
```
elBoberido, line 174: How about adding a test to fill the SlotMap, remove the even keys, and the fill it again with `insert`?

iceoryx2-bb/container/src/slotmap.rs
elfenpiff: You are maybe right. Sooner or later we have to revisit the relocatable container concept and then we introduce something like this here. Currently, I would stick to the current approach since it is used also in the other containers - but I agree, I also do not like it at all!

elBoberido: I'm not sure if there is currently another way to do it, except having a proc macro that calculates the memory size and adds a `buffer: [u8; N]` member to the struct. Anyway, without a proc macro the current way is less cumbersome and with the proc macro it doesn't really matter since it's hidden from the developer.

iceoryx2-bb/container/src/slotmap.rs
```diff
@@ -233,18 +233,15 @@ pub mod details {
...
-                    let key = SlotMapKey(key);
-                    self.store_value(key, value);
-                    Some(key)
-                }
-            }
+            self.acquire_next_free_index()
+                .map(|key| SlotMapKey(key))
+                .inspect(|key| {
+                    self.store_value(*key, value);
+                })
```
elBoberido, line 240: Oh no, `inspect` was introduced with Rust 1.76 and our MSRV is 1.75.
```suggestion
            self.acquire_next_free_index().map(|key| {
                let key = SlotMapKey(key);
                self.store_value(key, value);
                key
            })
```

elBoberido (commented):
Looks good. My advice with `inpsect` tuned out to not work with Rust 1.75.

elBoberido (approved):

PR #506, opened 2024-11-08
[#491] Propagate error strings from iceoryx2 via CXX API

### LINKED ISSUE #491: Provide error description strings via C/C++ API
## Brief feature description

For convenience / logging purposes, an API to retrieve a pointer to a string describing each error variant would be useful.

## Detailed information

1. Description strings should be statically compiled into the binary to avoid dynamic allocations
2. The API should provide a pointer to the correct string given an error variant as input

orecham:
## Notes for Reviewer

Adds C/C++ API to retrieve pointer to string describing iceoryx2 errors.

1. [x] Assign PR to reviewer
1. [ ] All checks have passed (except `task-list-completed`)

## References

Closes #491


iceoryx2-ffi/ffi/src/api/service_builder_pub_sub.rs
```diff
@@ -36,32 +38,55 @@ use std::alloc::Layout;
...
     C_SUBSCRIBER_BUFFER_MUST_BE_LARGER_THAN_HISTORY_SIZE,
+    #[StaticString = "already exists"]
     C_ALREADY_EXISTS,
+    #[StaticString = "insufficient permissions"]
     C_INSUFFICIENT_PERMISSIONS,
+    #[StaticString = "internal failure"]
     C_INTERNAL_FAILURE,
+    #[StaticString = "is being created by another instance"]
     C_IS_BEING_CREATED_BY_ANOTHER_INSTANCE,
-    C_OLD_CONNECTION_STILL_ACTIVE,
```
orecham, line 64: This error doesn't exist in the iceoryx2 error enum.

iceoryx2-ffi/cxx/include/iox2/config_creation_error.hpp
```diff
@@ -25,6 +25,9 @@ enum class ConfigCreationError : uint8_t {
     /// Parts of the config file could not be deserialized. Indicates some kind of syntax error.
     UnableToDeserializeContents
 };
+
+auto error_string(const iox2::ConfigCreationError& error) -> const char*;
```
elfenpiff, line 29: We should use here the `From` approach like it is done here: https://github.com/eclipse-iceoryx/iceoryx2/blob/main/iceoryx2-ffi/cxx/include/iox2/enum_translation.hpp

Since it is a conversion from enum `iox2::ConfigCreationError` to `const char*`.

This applies to all `error_string()` free functions.

iceoryx2-ffi/cxx/src/error_string.cpp
```diff
@@ -0,0 +1,135 @@
...
+#include "iox2/semantic_string.hpp"
+#include "iox2/service_builder_event_error.hpp"
+#include "iox2/service_builder_publish_subscribe_error.hpp"
+#include "iox2/service_error_enums.hpp"
+#include "iox2/subscriber_error.hpp"
+#include "iox2/waitset_enums.hpp"
+
+namespace iox2 {
+
+auto error_string(const iox2::ConfigCreationError& error) -> const char* {
```
elfenpiff, line 31: All those implementation should be changed to the format:

```cxx
template <>
constexpr auto from<iox2::ConfigCreationError, const char *>(const iox2::ConfigCreationError value) noexcept -> const char * {
  return iox2_config_creation_error_string(iox::into<iox2_config_creation_error_e>(value));
}
```

elfenpiff: Could you also rename the file into `enum_string.cpp`. It makes sense to have this conversion also for non error enums like `ServiceType` for instance.

elfenpiff (commented):
Good work and congratulations to the proc macro!

The only thing here is, that I would follow the iceoryx type conversion approach here and use `from<EnumType, const char*>` since it is a conversion from enum to a string. In this way it is nicely consistent.

iceoryx2-ffi/cxx/src/error_string.cpp
orecham: @elfenpiff Could the generalization to all enum types be left for a separate PR? If I change this I would need to add the implementation for getting the string representation of all enums as well.

elfenpiff: @elBoberido you need to solve the tie.

I would prefer `iox::into<const char*>`. For me, it is clear that the enum should be converted into a human-readable string.

elBoberido: @orecham @elfenpiff well, I would consider both approaches bad. What I want is `IOX_LOG(DEBUG, "Error code: " << result.error())`. It's a pity we re-invent the wheel with the logger, although we already have something in hoofs, which is more ergonomic than what we have now in iceoryx2.

orecham: @elBoberido This would be nice, but users not using the iceoryx logger would run into trouble.

elBoberido: Yeah, for them there would need to be a way to get the string.

I still do not have a clear favorite. Both approaches have pros and cons but I'm leaning towards the `into` approach
- we are already using it in our code base, so the users should be familiar with the concept
- we can use the same concept for conversion into `iox::string`, `std::string` or anything else if it is not possible to have a static conversion and it's clear to the user that a conversion has some cost

elfenpiff: I agree with @orecham that this

```cxx
iox::into<const char*>(result.error()))
```

is a bit more cumbersome to use than this

```cxx
error_string(result.error())
```

But the `iox::into` is the iceoryx idiomatic way of casting things so therefore I would stick with it - this would be also the first thing I would try if I wanted to cast X to a string. In this case, I prefer consistency over ergonomic handling.

orecham: @elfenpiff @elBoberido Changed to use `iox::into`.

Cargo.Bazel.lock
```diff
@@ -1,5 +1,5 @@
 {
-  "checksum": "d7b1759c5d28cbeb3fff8b9da5ca81261aa8c593891b9eff7d3c554d07b4c8dd",
+  "checksum": "8c7bf9445e9dc6b2654902f1a47db51d487887849c7db5b8cb0e4e4f3ab30945",
```
elfenpiff, line 2: Just out of curiosity, why did this change?

elfenpiff (dismissed):

Cargo.Bazel.lock
elBoberido: I guess because `iceoryx2-bb-derive-macros` was added to `Cargo.toml` or because a different bazel version was used on macOS?

orecham: This change was developed on `ubuntu` 24.04. I didn't pay close attention to why it changed but also assumed it was because of changes in `iceoryx2-bb-elementary` and `iceoryx2-bb-derive-macros`.

elfenpiff (approved):

PR #517, opened 2024-11-19
[#497] resizable shared memory

### LINKED ISSUE #497: ResizableSharedMemory
## Brief feature description

We require a struct that allows us to resize `SharedMemory` concepts during their lifetime.

It shall be a generic struct that has the same interface as the `SharedMemory` concept but with an extended behavior. As soon as the underlying pool allocator bucket size is exceeded it will increase the pool allocator bucket size, allocate a new shared memory segment and uses this.

elfenpiff:
## Notes for Reviewer

This PR introduces the `ResizableSharedMemory` concept. It will become the basic building block for dynamic publishers and server/clients.

The basic idea is, that the `ResizableSharedMemory` will create a new segment for all chunks with a specific layout whenever during `allocate()` the current segment returns out-of-memory, alignment unsupported or size unsupported.
The segment id is stored in the `PointerOffset`. The `PointerOffset` is not allowed to be larger than 64-bit because of the underlying lock-free algorithms.
Therefore, the maximum supported shm size of iceoryx2 shrank to 65535 TB to make space for the 8 byte segment id.

Old segments are removed by the `ResizableSharedMemory` on `deallocate` when no more chunks are in use and the mapping is removed in the `ResizableSharedMemoryView` when no more chunks are registered there.

The `ResizableSharedMemory` will be used by the `Publisher` as data segment and `deallocate` is called whenever the `Subscriber` returns a sample. The `ResizableSharedMemoryView` is used by the `Subscriber` to read the incoming samples. Before reading the `Subscriber` has to register the sample there to map new segments and has to deregister it before returning it to the publisher to unmap old segments.

Please focus on 2 things in this review:

1. Is the API for the `ResizableSharedMemory` a good fit also for future use cases?
2. Are there some tests missing for the `ResizableSharedMemory`. A good way to review this, is to look at just the test names and see if there is some use case I have missed.

1. [x] Assign PR to reviewer

## References

Closes #497


iceoryx2-cal/src/resizable_shared_memory/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+//! use std::time::Duration;
+//!
+//! fn example<Alloc: ShmAllocator, Shm: SharedMemory<Alloc>, Memory: ResizableSharedMemory<Alloc, Shm>>(
+//!     name: &FileName
+//! ) {
+//!     // owner process creates a new memory
+//!     let memory = Memory::MemoryBuilder::new(name)
+//!         // hint to the underlying allocator that we need up to 128 chunks of memory
+//!         //
+//!         // note: as soon as there are more than 128 chunks requested a new resized segment is
```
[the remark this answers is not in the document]
elfenpiff: A chunk is a piece of memory. It is defined with a starting point (pointer) and a length. I contrast to a rust Slice, it owns the memory and must be explicitly released.

Chunks can be different sizes within one allocator, if the allocator supports it - like for instance a buddy allocator.

iceoryx2-cal/src/shm_allocator/mod.rs
```diff
@@ -13,44 +13,147 @@
...
+        const SEGMENT_ID: u64 = 0;
+        Self(
+            (offset as u64) << (core::mem::size_of::<SegmentId>() * NUMBER_OF_BITS_IN_BYTE)
+                | SEGMENT_ID,
+        )
+    }
+
+    /// Sets the [`SegmentId`] of the [`PointerOffset`].
+    pub fn set_segment_id(&mut self, value: SegmentId) {
+        self.0 |= value.0 as u64;
```
elBoberido, line 68:
```suggestion
        self.0 &= !((1u64 << SegmentIdUnderlyingType::BITS) - 1);
        self.0 |= value.0 as u64;
```

iceoryx2-cal/src/resizable_shared_memory/dynamic.rs
```diff
@@ -0,0 +1,698 @@
...
+
+            let mut raw_segment_id = *name.as_string();
+            raw_segment_id.remove_range(0, segment_id_start_pos);
+
+            // check nymber of digits
+            for byte in raw_segment_id.as_bytes() {
+                let is_a_digit = b'0' <= *byte && *byte <= b'9';
+                if !is_a_digit {
+                    return None;
+                }
```
elBoberido, line 508:
```suggestion
                if !byte.is_ascii_digit() {
                    return None;
                }
```

iceoryx2-cal/src/shm_allocator/mod.rs
```diff
@@ -13,44 +13,147 @@
...
+#[derive(Clone, Copy, Eq, PartialEq)]
+pub struct PointerOffset(u64);
+
+impl PointerOffset {
+    /// Creates a new [`PointerOffset`] from the given offset value with the [`SegmentId`] == 0.
+    pub fn new(offset: usize) -> PointerOffset {
+        const SEGMENT_ID: u64 = 0;
+        Self(
+            (offset as u64) << (core::mem::size_of::<SegmentId>() * NUMBER_OF_BITS_IN_BYTE)
+                | SEGMENT_ID,
```
elBoberido, line 62:
```suggestion
            (offset as u64) << SegmentIdUnderlyingType::BITS | SEGMENT_ID,
```

iceoryx2-cal/src/shm_allocator/mod.rs
```diff
@@ -13,44 +13,147 @@
...
+    }
+
+    /// Sets the [`SegmentId`] of the [`PointerOffset`].
+    pub fn set_segment_id(&mut self, value: SegmentId) {
+        self.0 |= value.0 as u64;
+    }
+
+    /// Returns the offset.
+    pub fn offset(&self) -> usize {
+        (self.0 >> (core::mem::size_of::<SegmentId>() * NUMBER_OF_BITS_IN_BYTE)) as usize
```
elBoberido, line 73:
```suggestion
        (self.0 >> SegmentIdUnderlyingType::BITS) as usize
```

iceoryx2-cal/src/shm_allocator/mod.rs
```diff
@@ -13,44 +13,147 @@
...
 
-pub trait ShmAllocatorConfig: Copy + Default {}
+/// Trait that identifies a configuration of a [`ShmAllocator`].
+pub trait ShmAllocatorConfig: Copy + Default + Debug {}
 
-#[derive(Clone, Copy, Eq, PartialEq, Debug)]
-pub struct PointerOffset(usize);
+/// Defines the [`SegmentId`] of a [`SharedMemory`](crate::shared_memory::SharedMemory)
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct SegmentId(u8);
```
elBoberido, line 26:
```suggestion
type SegmentIdUnderlyingType = u8;
/// Defines the [`SegmentId`] of a [`SharedMemory`](crate::shared_memory::SharedMemory)
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
#[repr(transparent)]
pub struct SegmentId(SegmentIdUnderlyingType);
```

iceoryx2-cal/src/shm_allocator/mod.rs
```diff
@@ -13,44 +13,147 @@
...
     }
+
+    /// Returns the maximum value the [`SegmentId`] supports.
+    pub const fn max_segment_id() -> u8 {
+        u8::MAX
+    }
 }
 
-enum_gen! { ShmAllocationError
+const NUMBER_OF_BITS_IN_BYTE: usize = 8;
```
elBoberido, line 45: This can be removed

iceoryx2-cal/src/shm_allocator/mod.rs
```diff
@@ -13,44 +13,147 @@
...
+    }
+
+    /// Returns the offset.
+    pub fn offset(&self) -> usize {
+        (self.0 >> (core::mem::size_of::<SegmentId>() * NUMBER_OF_BITS_IN_BYTE)) as usize
+    }
+
+    /// Returns the [`SegmentId`].
+    pub fn segment_id(&self) -> SegmentId {
+        SegmentId((self.0 & SegmentId::max_segment_id() as u64) as u8)
```
elBoberido, line 78:
```suggestion
        SegmentId((self.0 & ((1u64 << SegmentIdUnderlyingType::BITS) - 1) as u8)
```

Always use the bits instead of a max value. Maybe at some point you decide that max is only 254 because the last value is used for something else.
In theory `SegmentIdUnderlyingType::MAX` should also work but I'm not sure if there is not a weird target where this would not hold for some weird reason.

iceoryx2-cal/src/shm_allocator/mod.rs
```diff
@@ -13,44 +13,147 @@
...
+pub enum AllocationStrategy {
+    /// Increases the memory so that it perfectly fits the new size requirements. This may lead
+    /// to a lot of reallocations but has the benefit that no byte is wasted.
+    BestFit,
+    #[default]
+    /// Increases the memory by rounding the increased memory size up to the next power of two.
+    /// Reduces reallocations a lot at the cost of increased memory usage.
+    PowerOfTwo,
+    /// The memory is not increased. This may lead to an out-of-memory error when allocating.
+    Static,
```
elBoberido, line 114: Just wondering whether `Static` wouldn't be a better default? Not asking to change it, just to get an idea why you choose `PowerOfTwo`?

iceoryx2-cal/src/shm_allocator/pool_allocator.rs
```diff
@@ -56,6 +61,71 @@ impl PoolAllocator {
...
+                    AllocationStrategy::Static => current_layout,
+                    AllocationStrategy::BestFit => unsafe {
+                        Layout::from_size_align_unchecked(
+                            layout.size().max(current_layout.size()),
+                            layout.align().max(current_layout.align()),
+                        )
+                    },
+                    AllocationStrategy::PowerOfTwo => unsafe {
+                        Layout::from_size_align_unchecked(
+                            layout.size().max(current_layout.size()).next_power_of_two(),
```
elBoberido, line 102: I think this needs to be `next_multiple_of(new_alignment)`

iceoryx2-cal/src/resizable_shared_memory/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+//!         //       created
+//!         .max_number_of_chunks_hint(128)
+//!         // hint to the underlying allocator that the chunks are not exceeding the Layout of
+//!         // [`u16`]
+//!         //
+//!         // note: as soon as there is a chunk requested that exceeds the provided layout hint a
+//!         //       new resized segment is created
+//!         .max_chunk_layout_hint(Layout::new::<u32>())
```
elBoberido, line 55: Do you mean `[u32]`?

elBoberido (commented):
The tests are not yet reviewed but I need a break. I will follow up with the tests tomorrow.

iceoryx2-cal/src/resizable_shared_memory/dynamic.rs
```diff
@@ -0,0 +1,698 @@
...
+                        ptr.offset
+                            .set_segment_id(SegmentId::new(state.current_idx.value() as u8));
+                        return Ok(ptr);
+                    }
+                    Err(e) => self.handle_reallocation(e, state, layout, &entry.shm)?,
+                },
+                None => fatal_panic!(from self,
+                        "This should never happen! {msg} since the current shared memory segment is not available!"),
+            }
+        }
```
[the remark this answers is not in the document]
elfenpiff: No, in the `Err(e)` path `self.handle_reallocation` is called which creates and adds a new segment to the `shared_memory_map` and then we try again to call `entry.shm.allocate(layout)` and see if we now get some memory.

iceoryx2-cal/src/shm_allocator/mod.rs
```diff
@@ -13,44 +13,147 @@
...
-    pub fn value(&self) -> usize {
+    /// Returns the underlying value of the [`SegmentId`]
+    pub fn value(&self) -> u8 {
         self.0
     }
+
+    /// Returns the maximum value the [`SegmentId`] supports.
+    pub const fn max_segment_id() -> u8 {
+        u8::MAX
+    }
```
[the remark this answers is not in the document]
elfenpiff: I wouldn't do it and leave it as it is for now. In then end, we do not want to support more than 256 different data segments for a single publisher but we want to support more than 256 reallocations per publisher data segment. We can achieve this with segment id recycling.

iceoryx2-cal/src/shm_allocator/mod.rs
elfenpiff: I do not understand what kind of problem this solves.

iceoryx2-cal/src/shm_allocator/mod.rs
elfenpiff: Wouldn't be your suggestion a bit more cumbersome and error prone?!

This would require us to add this const type, write tests for it and verify that the `BITS` value fits?

iceoryx2-cal/src/shm_allocator/mod.rs
elfenpiff: You are right! This should be the default.

iceoryx2-cal/src/resizable_shared_memory/mod.rs
elfenpiff: Nope, this is one layer lower and here you can mix types as you want. First you can give a hint for a `u32` than for `[Whatever]` if you like.

iceoryx2-cal/src/resizable_shared_memory/dynamic.rs
```diff
@@ -0,0 +1,698 @@
...
+
+    fn extract_name_from_management_segment(name: &FileName) -> Option<FileName> {
+        let mut name = *name;
+        if let Ok(true) = name.strip_suffix(MANAGEMENT_SUFFIX) {
+            if let Ok(true) = name.strip_suffix(SEGMENT_ID_SEPARATOR) {
+                return Some(name);
+            }
+        }
+
+        None
```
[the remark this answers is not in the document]
elfenpiff: No it is not since the result returns a bool stating `true` or `false` and this indicates if the suffix really existed and was removed or the suffix did not exist and therefore was not removed.

iceoryx2-cal/src/resizable_shared_memory/dynamic.rs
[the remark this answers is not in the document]
elBoberido: Oh, right. I overlooked that there is a path where `?` does not return. I thought it would always be an error.

iceoryx2-cal/src/shm_allocator/mod.rs
elBoberido: Because the bits need to be cleared before they are set, else this could happen
```rs
foo.set_segment_id(7); // results in 7
foo.set_segment_id(8); // results in 7 | 8 -> 15
```

It's a public function and can therefore be called multiple times.

iceoryx2-cal/tests/resizable_shared_memory_tests.rs
```diff
@@ -0,0 +1,954 @@
...
+            .unwrap();
+
+        let sut_viewer = Sut::ViewBuilder::new(&storage_name)
+            .config(&config)
+            .open()
+            .unwrap();
+
+        let mut ptrs = vec![];
+        for i in 0..NUMBER_OF_REALLOCATIONS {
+            let size = 2 + i;
```
elBoberido, line 232: I you really want the `NUMBER_OF_REALLOCATIONS` be the real number of reallocations, even for power of two, then the size should probably be `2 * i + 1` and run from `1..=NUMBER_OF_REALLOCATIONS`

iceoryx2-cal/tests/resizable_shared_memory_tests.rs
```diff
@@ -0,0 +1,954 @@
...
```
elBoberido, line 301: I guess it would be the same here

iceoryx2-cal/tests/resizable_shared_memory_tests.rs
```diff
@@ -0,0 +1,954 @@
...
+            .max_number_of_chunks_hint(8)
+            .create()
+            .unwrap();
+
+        let result = sut.allocate(Layout::from_size_align(8, 1).unwrap());
+        assert_that!(result, is_err);
+    }
+
+    #[test]
+    fn static_allocation_strategy_increase_available_chunks<
```
elBoberido, line 460: Is the name correct? Is the static allocation strategy assumed to increase the available chunks?

iceoryx2-cal/tests/resizable_shared_memory_tests.rs
```diff
@@ -0,0 +1,954 @@
...
+        assert_that!(sut_viewer.number_of_active_segments(), eq 4);
+
+        unsafe { sut_viewer.unregister_offset(chunk_1) };
+        assert_that!(sut_viewer.number_of_active_segments(), eq 3);
+        unsafe { sut_viewer.unregister_offset(chunk_2) };
+        assert_that!(sut_viewer.number_of_active_segments(), eq 2);
+        unsafe { sut_viewer.unregister_offset(chunk_3) };
+        assert_that!(sut_viewer.number_of_active_segments(), eq 1);
+        unsafe { sut_viewer.unregister_offset(chunk_4) };
+        assert_that!(sut_viewer.number_of_active_segments(), eq 1);
```
elBoberido, line 903: So, it remains 1 even after releasing chunk_4 because it's the last added segment and kept for later, right?

iceoryx2-cal/src/shm_allocator/mod.rs
elfenpiff: Oh wow ... you are right. I moved the PointerOffset into a separate file and created also unit tests verifying that this does work now.

iceoryx2-cal/tests/resizable_shared_memory_tests.rs
elfenpiff: correct

iceoryx2-cal/tests/resizable_shared_memory_tests.rs
elfenpiff: But with a power of two strategy this will explode. And just for fun, let's assume we have a power of hundred strategy then this will be hard to test for nearly any number.

I renamed the variable into `NUMBER_OF_ITERATIONS`. The underlying strategy will define how many reallocations will happen but this is not what we want to test here. We just want to ensure that the chunk increases.

elBoberido: Right. My proposal was also flawed since it did not significantly increase the number of re-allocations.

Renaming the constant helped.

elBoberido (approved):

PR #524, opened 2024-11-24
[#390] event based comm example cpp

### LINKED ISSUE #390: iceoryx WaitSet integration

elfenpiff:
## Notes for Reviewer

Ports the rust event_based_communication example to C++. 

Additionally, introduce `FileDescriptorBased` and implement it for `FileDescriptorView` so that `WaitSet::attach_**()` can be handled easier and the user does not have to acquire low-level implementation details like the file descriptor.

1. [x] Assign PR to reviewer

## References

Closes #390


examples/rust/event_based_communication/publisher.rs
```diff
@@ -28,27 +28,40 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
             publisher.handle_event().unwrap();
         }
         CallbackProgression::Continue
     };
 
+    // Start the event loop. It will run until `CallbackProgression::Stop` is returned by the
+    // event callback or an interrupt/termination signal was received.
     waitset.wait_and_process(on_event)?;
 
+    println!("exit ...");
```
elBoberido, line 57:
```suggestion
    println!("exit");
```

no ellipses if there is no more text comming

examples/cxx/event_based_communication/src/publisher.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+  private:
+    EventBasedPublisher(Publisher<ServiceType::Ipc, TransmissionData, void>&& publisher,
+                        Listener<ServiceType::Ipc>&& listener,
+                        Notifier<ServiceType::Ipc>&& notifier);
+
+    Publisher<ServiceType::Ipc, TransmissionData, void> m_publisher;
+    Listener<ServiceType::Ipc> m_listener;
+    Notifier<ServiceType::Ipc> m_notifier;
+};
```
elBoberido, line 59: Can you move this into an `hpp` file

examples/cxx/event_based_communication/src/publisher.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+auto main() -> int {
```
elBoberido, line 61: This should probably also be moved to a separate file. It's a little bit confusion having everything in the same file.

examples/cxx/event_based_communication/src/pubsub_event.hpp
```diff
@@ -0,0 +1,63 @@
...
+    case from<PubSubEvent, size_t>(PubSubEvent::SubscriberConnected):
+        return PubSubEvent::SubscriberConnected;
+    case from<PubSubEvent, size_t>(PubSubEvent::SubscriberDisconnected):
+        return PubSubEvent::SubscriberDisconnected;
+    case from<PubSubEvent, size_t>(PubSubEvent::SentSample):
+        return PubSubEvent::SentSample;
+    case from<PubSubEvent, size_t>(PubSubEvent::ReceivedSample):
+        return PubSubEvent::ReceivedSample;
+    case from<PubSubEvent, size_t>(PubSubEvent::SentHistory):
+        return PubSubEvent::SentHistory;
```
elBoberido, line 54:
```suggestion
    case into<size_t>(PubSubEvent::PublisherConnected):
        return PubSubEvent::PublisherConnected;
    case into<size_t>(PubSubEvent::PublisherDisconnected):
        return PubSubEvent::PublisherDisconnected;
    case into<size_t>(PubSubEvent::SubscriberConnected):
        return PubSubEvent::SubscriberConnected;
    case into<size_t>(PubSubEvent::SubscriberDisconnected):
        return PubSubEvent::SubscriberDisconnected;
    case into<size_t>(PubSubEvent::SentSample):
        return PubSubEvent::SentSample;
    case into<size_t>(PubSubEvent::ReceivedSample):
        return PubSubEvent::ReceivedSample;
    case into<size_t>(PubSubEvent::SentHistory):
        return PubSubEvent::SentHistory;
```

Rule of thumb, implement `from` but use `into` :)

examples/cxx/event_based_communication/src/subscriber.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+using namespace iox2;
+
+// High-level subscriber class that contains besides a subscriber also a notifier
+// and a listener. The notifier is used to send events like
+// `PubSubEvent::ReceivedSample` or to notify the publisher that a new subscriber
+// connected.
+// The listener waits for events originating from the publisher like
+// `PubSubEvent::SentSample`.
+class EventBasedSubscriber : public FileDescriptorBased {
```
elBoberido, line 38: Please also move this into an `hpp` file.

examples/cxx/event_based_communication/src/publisher.cpp
elfenpiff: Actually, I do not want to do this since this is the publisher implementation.

If this is moved into a separate header it seems like a little example helper but this is the essence of the example and as important as `main`.

With Rust I add the advantage to move everything below main but with C++ this is not possible, therefore I decided to separate declaration from definition to make it a bit clearer.

examples/cxx/event_based_communication/src/subscriber.cpp
elfenpiff: I tried it, but I do not like it. I really prefer it when the essence is in one file in the case of an example.

The TransmissionData and the PubSubEvent are just simple helpers but for me, it is important that the user sees that they can define there own advanced publisher/subscriber, combine them with events and attach it to a waitset.

examples/cxx/event_based_communication/src/publisher.cpp
```diff
@@ -0,0 +1,179 @@
...
+constexpr uint64_t HISTORY_SIZE = 20;
+
+///////////////////////////////////////////////
+/// START: EventBasedPublisher declaration
+///////////////////////////////////////////////
+
+// High-level publisher class that contains besides a publisher also a notifier and a listener.
+// The notifier is used to send events like `PubSubEvent::SentSample` or `PubSubEvent::SentHistory`
+// and the listener to wait for new subscribers.
+class EventBasedPublisher : public FileDescriptorBased {
```
orecham, line 41: I know the Rust version is already in, but this name is really confusing to me. It sounds to me like a publisher implemented on top of some kind of event mechanism (e.g. like it is used in the terms "token-based authentication"), rather than one that augments publishing with notification.

If there is room to adapt this, I would suggest something more like `NotifyingPublisher` and `ListeningSubscriber` ?

elfenpiff: I will do rename them in the Rust and C++ example in this PR ... in 8 hours or later ;)

elBoberido (approved):

PR #537, opened 2024-12-01
[#533] add timeout to waitset

### LINKED ISSUE #533: Add timeout to `WaitSet::wait_and_process_once()`
## Brief feature description

User shall be able to provide a timeout to the waitset so that they do not wait for infinity until something arrives.

elfenpiff:
## Notes for Reviewer

From @elBoberido :

Unforunately this is already merged else I would have objected
- why is this necessary when there is already the deadline option?
- why use a surrogate for cases when no timeout is required?
  - this makes the API more cumbersome for all the users who do not need/want a timeout since it's not obvious what to use
  - either use an `Option<Duration>` or add an additional `*_with_timeout` function

Answers:

> - why is this necessary when there is already the deadline option?

Do you mean interval? Deadline is an attached listener where a notification is expected before the deadline has passed. An interval is a recurring event, and an explicit timeout option makes it possible for the user to add a simple max wait time option.

> - why use a surrogate for cases when no timeout is required?

You are right. With this PR there are now 2 calls:
 * `WaitSet::wait_and_process_once()`
 * `WaitSet::wait_and_process_once_with_timeout()`

1. [x] Assign PR to reviewer

## References

Closes #533

### DIFF
--- a/iceoryx2/src/port/waitset.rs
+++ b/iceoryx2/src/port/waitset.rs
@@ -747,7 +747,7 @@ impl<Service: crate::service::Service> WaitSet<Service> {
         mut fn_call: F,
     ) -> Result<WaitSetRunResult, WaitSetRunError> {
         loop {
-            match self.wait_and_process_once(&mut fn_call, Duration::MAX) {
+            match self.wait_and_process_once(&mut fn_call) {
                 Ok(WaitSetRunResult::AllEventsHandled) => (),
                 Ok(v) => return Ok(v),
                 Err(e) => {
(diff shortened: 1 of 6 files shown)


iceoryx2/src/port/waitset.rs
```diff
@@ -799,15 +853,17 @@ impl<Service: crate::service::Service> WaitSet<Service> {
...
     /// # Ok(())
     /// # }
     /// ```
-    pub fn wait_and_process_once<F: FnMut(WaitSetAttachmentId<Service>) -> CallbackProgression>(
+    pub fn wait_and_process_once_with_timeout<
+        F: FnMut(WaitSetAttachmentId<Service>) -> CallbackProgression,
+    >(
         &self,
         mut fn_call: F,
         timeout: Duration,
```
elBoberido, line 869: Food for thought. In other Rust APIs the `_suffix` parameter is the first, e.g. with `map_or`, where the `_or` parameter is the first and the `map` parameter the second.

elBoberido (approved):
Looks good. Just one food for thought

iceoryx2/src/port/waitset.rs
elfenpiff: I would prefer to add the parameter since it is an addition to `wait_and_process_once`. Also in my head it behaves somehow like the C++ default parameter value.

@orecham you decide the tie

elBoberido: Just a suggestion but I think the reason this is done is because the additional parameter is just a value and therefore a few characters but the closure can span a few lines and then the additional parameter is far away from the function name.

Issue #550, opened 2024-12-16
Set ids for custom events


elfenpiff:
## Brief feature description

For the event messaging pattern the user shall be able to define a custom event id whenever a notifier:
 * is created, 
 * is destroyed 
 * is identified as dead

The events shall be emitted automatically by the notifier without any additional user interaction.

PR #552, opened 2024-12-17
[#550] custom event ids

elfenpiff:

* [ ] Assign PR to reviewer

## References

Closes #


iceoryx2-ffi/cxx/include/iox2/service_builder_event.hpp
```diff
@@ -48,6 +49,30 @@ class ServiceBuilderEvent {
...
+    IOX_BUILDER_SWITCH(disable_notifier_dropped_event);
+
+    /// If the [`Service`] is created it defines the event that shall be emitted when a
+    /// [`Notifier`] is identified as dead. If [`None`] is
+    /// provided no event will be emitted.
+    IOX_BUILDER_OPTIONAL(EventId, notifier_dead_event);
+
+    /// If the [`Service`] is created it disables sending an event when a notifier was identified
+    /// as dead.
+    IOX_BUILDER_SWITCH(disable_notifier_dead_event);
```
elBoberido, line 74:
```suggestion
    /// If the [`Service`] is created it defines the event that shall be emitted by every newly
    /// created [`Notifier`].
    IOX_BUILDER_OPTIONAL(EventId, notifier_created_event);

    /// If the [`Service`] is created it defines the event that shall be emitted by every
    /// [`Notifier`] before it is dropped. If [`None`] is
    /// provided a [`Notifier`] will not emit an event.
    IOX_BUILDER_OPTIONAL(EventId, notifier_dropped_event);

    /// If the [`Service`] is created it defines the event that shall be emitted when a
    /// [`Notifier`] is identified as dead. If [`None`] is
    /// provided no event will be emitted.
    IOX_BUILDER_OPTIONAL(EventId, notifier_dead_event);
    
    /// If the [`Service`] is created it disables sending an event when a new notifier was created.
    void disable_notifier_created_event() {
        m_notifier_created_event.reset();
    }
    
    /// If the [`Service`] is created it disables sending an event when a notifier was dropped.
    void disable_notifier_dropped_event() {
        m_notifier_dropped_event.reset();
    }
    
    /// If the [`Service`] is created it disables sending an event when a notifier was identified
    /// as dead.
    void disable_notifier_dead_event() {
        m_notifier_dead_event.reset();
    }
```

No need to make it unnecessary complicated. This will also simplify the logic in `service_builder_event.cpp`

iceoryx2-ffi/cxx/include/iox2/config.hpp
```diff
@@ -212,6 +212,18 @@ class Event {
     auto event_id_max_value() && -> size_t;
     /// Set the largest event id supported by the event service
     void set_event_id_max_value(size_t value) &&;
+    /// Defines the event id value that is emitted after a new notifier was created.
+    auto notifier_created_event() && -> iox::optional<size_t>;
+    /// Sets the event id value that is emitted after a new notifier was created.
+    void set_notifier_created_event(iox::optional<size_t> value) &&;
```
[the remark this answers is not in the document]
elfenpiff: This is intentional. In the builder it is no longer an optional but for the global config and the config file it is an optional. Some = emit event, None = no event.

iceoryx2/src/service/builder/event.rs
```diff
@@ -232,6 +247,54 @@ impl<ServiceType: service::Service> Builder<ServiceType> {
...
+        self.config_details().notifier_dead_event = Some(value.as_value());
+        self.verify_notifier_dead_event = true;
+        self
+    }
+
+    /// If the [`Service`] is created it disables the event that shall be emitted when a
+    /// [`Notifier`](crate::port::notifier::Notifier) is identified as dead.
+    pub fn disable_notifier_dead_event(mut self) -> Self {
+        self.config_details().notifier_dead_event = None;
+        self.verify_notifier_dead_event = true;
```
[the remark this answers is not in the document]
elfenpiff: Under the hood it is still an optional. So by explicitly turn this on or off, the user sets it as a requirement. If the service does not follow the requirement an error is reported when the service is opened.

iceoryx2-ffi/cxx/include/iox2/service_builder_event.hpp
elfenpiff: Actually, this does not work. When this is not called, the user has not set the requirement and takes whatever QoS the service is offering. As soon as you call disable or enable with a specific event, you set this as a requirement that is ensured when opening the service.

iceoryx2-ffi/cxx/include/iox2/config.hpp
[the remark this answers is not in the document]
elBoberido: Wouldn't it make sense to have the same API with `disable`?

iceoryx2-ffi/cxx/include/iox2/service_builder_event.hpp
elBoberido: Okay, then I do not understand what you want to achieve :sweat_smile:

iceoryx2-ffi/cxx/include/iox2/config.hpp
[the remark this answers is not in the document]
elfenpiff: No. In Rust the config is a struct where all members are public and can be set by the user directly. It is also serializable and stored in the `iceoryx2.toml` file. This config entry is still an `Option<EventId>` and it does not make sense to split it up into two fields since then we confuse the user and may have entries that are in contradiction to each other.

We require here a C++ API since we cannot access the struct members directly in C++. But the API shall reflect the underlying structure of the Rust structs/members. Therefore this method has still an optional.

iceoryx2-ffi/cxx/include/iox2/service_builder_event.hpp
elBoberido: Hmm, if it's three options, a new enum might indeed make sense
```rs
enum Event {
    Default, // can be on or off; maybe Preconfigured
    Custom{id},
    Disabled,
}
```

elfenpiff: We can go that way, but then we need this for every builder option in every factory for all ports and services since it is always handled like this.

I think your suggestion is not bad at all, but would require a greater refactoring. My suggestion for now:
1. Pursue the current approach and merge the PR
2. Create an issue to track this and tackle this in v0.5 or later

elBoberido (approved):

Issue #442, opened 2024-10-06
Move default config to a standard location


elBoberido:
## Brief feature description

Currently, when an application is run, it looks for `config/iceoryx2.toml`, relative to where it was started. This is error prone and the config should be at a standard location like `/etc/iceoryx2/config.toml` or `/etc/iceoryx2/domain-id/config.toml`.

PR #616, opened 2025-02-04
[#442] add config paths to platform

elfenpiff:
## Notes for Reviewer

Adds the default config lookup paths (according to XDG - at least on UNIX platforms) to the platform abstraction.
Via the posix wrappers they become available to the user. The user object provides now a `config_dir` method to obtain the users default config dir ($HOME/.config) and the system_configuration provides the global config path `/etc` on POSIX and `C:\` on Windows.

* [x] Assign PR to reviewer

## References

Relates to #442

### DIFF
--- a/config/README.md
+++ b/config/README.md
@@ -5,6 +5,28 @@ iceoryx2 introduces global configuration settings. It enables the concurrent
 execution of multiple iceoryx2 setups on the same machine or within a single
 process by employing distinct configurations.
 
+When **iceoryx2** is started without an explicitly loaded configuration,
+such as:
+
+```rust
+use iceoryx2::config::Config;
+use iceoryx2_bb_system_types::file_path::FilePath;
+use iceoryx2_bb_container::semantic_string::SemanticString;
+
+Config::setup_global_config_from_file(
+    &FilePath::new(b"my/custom/config/file.toml")?)?;
+```
+
+it will automatically search for a configuration file in the following
+locations, in order:
+
+1. `$PWD/config/iceoryx2.toml`
+2. `$HOME/.config/iceoryx2/iceoryx2.toml`
+3. `/etc/iceoryx2/iceoryx2.toml`
+
+If no configuration file is found in these locations, **iceoryx2** will use
+its default settings.
+
 ## Sections
 
 The configuration is organized into two main sections:
--- a/iceoryx2-pal/posix/src/windows/constants.rs
+++ b/iceoryx2-pal/posix/src/windows/constants.rs
@@ -15,7 +15,7 @@
 
 use crate::posix::types::*;
 
-pub const GLOBAL_CONFIG_PATH: &str = "C:/";
+pub const GLOBAL_CONFIG_PATH: &str = "C:/ProgramData";
 pub const USER_CONFIG_PATH: &str = ".config";
 
 pub const CPU_SETSIZE: usize = core::mem::size_of::<usize>() * 8;
--- a/iceoryx2-pal/posix/src/windows/constants.rs
+++ b/iceoryx2-pal/posix/src/windows/constants.rs
@@ -15,8 +15,8 @@
 
 use crate::posix::types::*;
 
-pub const GLOBAL_PATH: &str = "C:/";
-pub const USER_PATH: &str = ".config";
+pub const GLOBAL_CONFIG_PATH: &str = "C:/";
+pub const USER_CONFIG_PATH: &str = ".config";
 
 pub const CPU_SETSIZE: usize = core::mem::size_of::<usize>() * 8;
 pub const MAX_NUMBER_OF_THREADS: usize = 1024;
--- a/iceoryx2-pal/posix/src/windows/constants.rs
+++ b/iceoryx2-pal/posix/src/windows/constants.rs
@@ -15,6 +15,9 @@
 
 use crate::posix::types::*;
 
+pub const GLOBAL_PATH: &str = "C:/";
+pub const USER_PATH: &str = ".config";
+
 pub const CPU_SETSIZE: usize = core::mem::size_of::<usize>() * 8;
 pub const MAX_NUMBER_OF_THREADS: usize = 1024;
 pub const FD_SETSIZE: usize = windows_sys::Win32::Networking::WinSock::FD_SETSIZE as _;
--- a/iceoryx2/src/config.rs
+++ b/iceoryx2/src/config.rs
@@ -406,7 +406,7 @@ impl Config {
                          when iceoryx2_bb_posix::user::User::from_self(),
                          with ConfigIterationFailure::UnableToAcquireCurrentUserDetails,
                          "{} since the current user details could not be acquired.", msg);
-        let mut user_config = user.config_dir().clone();
+        let mut user_config = *user.config_dir();
         fail!(from origin,
                 when user_config.add_path_entry(&Self::relative_config_path()),
                 with ConfigIterationFailure::TooLongUserConfigDirectory,
--- a/iceoryx2/src/config.rs
+++ b/iceoryx2/src/config.rs
@@ -71,19 +71,28 @@
 
 use core::time::Duration;
 use iceoryx2_bb_container::semantic_string::SemanticString;
-use iceoryx2_bb_elementary::lazy_singleton::*;
-use iceoryx2_bb_posix::{file::FileBuilder, shared_memory::AccessMode};
+use iceoryx2_bb_elementary::{lazy_singleton::*, CallbackProgression};
+use iceoryx2_bb_posix::{
+    file::FileBuilder, shared_memory::AccessMode, system_configuration::get_global_config_path,
+};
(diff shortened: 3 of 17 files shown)


config/README.md
```diff
@@ -5,6 +5,28 @@ iceoryx2 introduces global configuration settings. It enables the concurrent
...
+
+Config::setup_global_config_from_file(
+    &FilePath::new(b"my/custom/config/file.toml")?)?;
+```
+
+it will automatically search for a configuration file in the following
+locations, in order:
+
+1. `$PWD/config/iceoryx2.toml`
+2. `$HOME/.config/iceoryx2/iceoryx2.toml`
```
elBoberido, line 24: Btw, is the `iceoryx2` folder part of the config?

If yes, I would make that independent of the config, so the user could set the `domain` or whatever ... hmm, maybe an `uri` would make sense, i.e. `io.iceoryx2`. User could then use their own reverse domain name notation when they want to isolate themself from the general communication. `/tmp/io.iceoryx2` could then hold a list of all used `uri`, which would help the tooling to discover those islands.

iceoryx2-pal/posix/src/windows/constants.rs
```diff
@@ -15,6 +15,9 @@
 
 use crate::posix::types::*;
 
+pub const GLOBAL_CONFIG_PATH: &str = "C:/";
```
elBoberido, line 18: What do you think about using `C:/ProgramData`? Classic iceoryx is also using that folder for the default lookup since there were issues with `C:/`

iceoryx2/src/config.rs
```diff
@@ -71,19 +71,28 @@
...
 
-use iceoryx2_bb_log::{debug, fail, trace, warn};
+use iceoryx2_bb_log::{fail, fatal_panic, trace, warn};
 
 use crate::service::port_factory::publisher::UnableToDeliverStrategy;
 
-/// Path to the default config file
-pub const DEFAULT_CONFIG_FILE: &[u8] = b"config/iceoryx2.toml";
+const DEFAULT_CONFIG_FILE_NAME: &[u8] = b"iceoryx2.toml";
+const DEFAULT_CONFIG_PATH: &[u8] = b"config";
```
elBoberido, line 88: I don't understand the naming of `DEFAULT_CONFIG_PATH`, especially with the existence of `RELATIVE_CONFIG_FILE_PATH`

Edit. okay, with the context below, I understand it, but I would just remove this. Alternatively, please consider naming this `RELATIVE_LOCAL_CONFIG_PATH`

iceoryx2/src/config.rs
```diff
@@ -403,29 +485,41 @@ impl Config {
...
+                    }
+                    Err(ConfigCreationError::FailedToOpenConfigFile) => {
+                        CallbackProgression::Continue
+                    }
+                    Err(e) => {
+                        warn!(from origin,
+                                "Config file found \"{}\" but a failure occurred ({:?}) while reading the content.",
+                                config_file_path, e);
+                        CallbackProgression::Continue
+                    }
```
elBoberido, line 512: I'm not sure if this is a good idea. If there is a config file but it cannot be read, it should not fail silently and try to read the next one. This could result in application running in the wrong domain without the user noticing it.

What do you think? If there is a config file, it needs to be valid, else the application panics?

iceoryx2/src/config.rs
```diff
@@ -403,29 +485,41 @@ impl Config {
...
                 }
+            }) {
+                warn!(from origin,
+                    "A failure occurred ({:?}) while looking up all available config files.", e);
+            }
+
+            if !is_config_file_set {
+                warn!(from origin,
+                    "Since no config file could be loaded, a config with internal default values will be set.");
+                ICEORYX2_CONFIG.set_value(Config::default());
```
elBoberido, line 522: I would do this only if no config file was found, not when it's available but corrupt.

config/README.md
elfenpiff: Currently, the `iceoryx2` is a static fixed part of the config file lookup location. It cannot be part of the config since we do not know the config until we have read it and for this we need to know the full config file path.

Nevertheless, I like the underlying idea and would like to pursue it further. Maybe we could somehow define a domain that is then reflected in the file name or a subdirectory, but `/etc/iceoryx2/` or `$HOME/.config/iceoryx2` is static and unchangeable but the structure inside this folder is open for debate.

config/README.md
```diff
@@ -5,6 +5,28 @@ iceoryx2 introduces global configuration settings. It enables the concurrent
...
+use iceoryx2_bb_container::semantic_string::SemanticString;
+
+Config::setup_global_config_from_file(
+    &FilePath::new(b"my/custom/config/file.toml")?)?;
+```
+
+it will automatically search for a configuration file in the following
+locations, in order:
+
+1. `$PWD/config/iceoryx2.toml`
```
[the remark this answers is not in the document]
elfenpiff: We do not yet have this. Would it be all right that we follow up on this in another PR?

Currently, iceoryx2 already behaves like this - so we do not change anything, we just document the current state here. And for development, it is actually quite useful to have it loaded from this location.

But I agree with you, and we have to address this!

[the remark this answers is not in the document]
elfenpiff: @elBoberido here is the issue: https://github.com/eclipse-iceoryx/iceoryx2/issues/617

iceoryx2/src/config.rs
```diff
@@ -354,6 +364,82 @@ impl Default for Config {
 }
 
 impl Config {
+    fn relative_local_config_path() -> Path {
+        fatal_panic!(from "Config::default_config_path",
+            when Path::new(RELATIVE_LOCAL_CONFIG_PATH),
+            "This should never happen! The default config path contains invalid symbols.")
```
FerdinandSpitzschnueffler, line 370:
```suggestion
            "This should never happen! The relative local config path contains invalid symbols.")
```
?

iceoryx2/tests/node_tests.rs
```diff
@@ -127,6 +127,7 @@ mod node {
 
     #[test]
     fn without_custom_config_global_config_is_used<S: Service>() {
+        set_log_level(LogLevel::Trace);
```
FerdinandSpitzschnueffler, line 130: Why do we need this log level? Was it just used while developing?

iceoryx2/src/config.rs
```diff
@@ -354,6 +364,82 @@ impl Default for Config {
...
+
+    fn relative_config_path() -> Path {
+        fatal_panic!(from "Config::relative_config_path",
+            when Path::new(RELATIVE_CONFIG_FILE_PATH),
+            "This should never happen! The relative config path contains invalid symbols.")
+    }
+
+    fn iterate_over_config_files<F: FnMut(FilePath) -> CallbackProgression>(
+        mut callback: F,
+    ) -> Result<(), ConfigIterationFailure> {
```
elBoberido, line 395: What do you think about changing this to
```rs
fn get_config_file() -> Result<Option<Path>, Error> {}
```
On error, the caller can panic and if it was successful, the caller can try to open the config or use the default config if no path was found.

This would simplify the code a bit.

iceoryx2/src/config.rs
```diff
@@ -403,29 +489,40 @@ impl Config {
...
-                    ICEORYX2_CONFIG.set_value(Config::default());
+            let mut is_config_file_set = false;
+            if let Err(e) = Self::iterate_over_config_files(|config_file_path| {
+                match Config::setup_global_config_from_file(&config_file_path) {
+                    Ok(_) => {
+                        is_config_file_set = true;
+                        CallbackProgression::Stop
+                    }
+                    Err(ConfigCreationError::FailedToOpenConfigFile) => {
+                        CallbackProgression::Continue
```
elBoberido, line 509: Does this error mean that the config file is not available or that the user does not have the access rights to open the config file?

iceoryx2/src/config.rs
elfenpiff: But this function iterates over multiple config files? So one option wouldn't be sufficient.

iceoryx2/src/config.rs
elfenpiff: I introduced more fine grained error handling. If the file does not exist or the user does not have permission to access the file, the iteration will continue.

config/README.md
elfenpiff: I created issue #618 to keep track of this idea.

iceoryx2/src/config.rs
elBoberido: No, it requires one config file to be present at one of these locations. Once the config file is present it is either valid or the caller panics.

iceoryx2/src/config.rs
elBoberido: No, if the user has not permission, it should not silently use another config file. This is error prone.

iceoryx2/src/config.rs
elfenpiff: Aaah understood!

elBoberido (approved):

Issue #707, opened 2025-04-29
Implement `ZeroCopySend` for `StaticConfig`


orecham:
## Cumbersome API

The [`StaticConfig`](https://github.com/eclipse-iceoryx/iceoryx2/blob/main/iceoryx2/src/service/static_config/mod.rs#L43) for services contains fields which are not zero-copy-compatible thus it must be serialised to be shared between processes.

## Improvement Suggestion

1. Set the `ServiceName` to a `FixedSizeByteString` with length 255.
1. Set the `Vector` in `AttributeSet` to a `FixedSizeVec` of capacity 16.
1. Set the `Attribute::key` to a `FixedSizeByteString` with length 64.
1. Set the `Attribute::value` to a `FixedSizeByteString` with length 256

Put all length constants into a separate file in `iceoryx2/src/constants.rs`

PR #719, opened 2025-05-04
[#707] Implement ZeroCopySend for StaticConfig

orecham:
## Notes for Reviewer

1. Use self-contained types for fields of `StaticConfig`
2. Removed use of vectors in implementation of `attributes.rs`
3. Update `ZeroCopySend` derive macro to support enum variants
4. Implement `ZeroCopySend` for `StaticConfig` and all types it has nested

* [ ] Assign PR to reviewer

## References

Closes #

### DIFF
--- a/doc/release-notes/iceoryx2-unreleased.md
+++ b/doc/release-notes/iceoryx2-unreleased.md
@@ -286,7 +286,7 @@
     let attribute = Attribute::new("my_key", "my_value");
 
     // new - keys or value known to fit within maximum length
-    let attribute = Attribute::new("my_key".try_into().unwrap(), "my_value".try_into().unwrap());
+    let attribute = Attribute::new(&"my_key".try_into().unwrap(), &"my_value".try_into().unwrap());
 
     // new - key or value length unknown
     fn get_key() -> &str;
@@ -298,5 +298,5 @@
     let value = AttributeValue::try_from(get_value()).unwrap_or_else(|e| {
         // Handle error: e.g., log error, use default value, or propagate error
     });
-    let attribute = Attribute::new(key, value);
+    let attribute = Attribute::new(&key, &value);
     ```
--- a/doc/release-notes/iceoryx2-unreleased.md
+++ b/doc/release-notes/iceoryx2-unreleased.md
@@ -275,3 +275,8 @@
 15. Renamed `AttributeVerifier::keys()` to `AttributeVerifier::required_keys()`
 
 16. Renamed `AttributeVerifier::attributes()` to `AttributeVerifier::required_attributes()`
+
+17. Attribute keys and values are now `SemanticStrings`
+    1. All arguments of methods in `Attribute`, `AttributeSpecifier`,
+        `AttributeVerifier`, and `AttributeSet` changed from `&str` to
+        `FixedSizeByteStrings` that implement the `SemanticString` trait
--- a/doc/release-notes/iceoryx2-unreleased.md
+++ b/doc/release-notes/iceoryx2-unreleased.md
@@ -179,7 +179,7 @@
        .open_or_create()?;
    ```
 
-2. Renamed `PublisherLoanError` into `LoanError`
+2. Renamed `PublisherLoanError` to `LoanError`
 
    ```rust
    // old
@@ -197,7 +197,7 @@
    };
    ```
 
-3. Renamed `PublisherSendError` into `SendError`
+3. Renamed `PublisherSendError` to `SendError`
 
    ```rust
    // old
@@ -215,7 +215,7 @@
    };
    ```
 
-4. Renamed `SubscriberReceiveError` into `ReceiveError`
+4. Renamed `SubscriberReceiveError` to `ReceiveError`
 
    ```rust
    // old
@@ -234,43 +234,44 @@
    ```
 
 5. Renamed `PublisherSendError::ConnectionBrokenSincePublisherNoLongerExists`
-   into `SendError::ConnectionBrokenSinceSenderNoLongerExists`
+   to `SendError::ConnectionBrokenSinceSenderNoLongerExists`
 
 6. Renamed `ConnectionFailure::UnableToMapPublishersDataSegment`
-   into `ConnectionFailure::UnableToMapSendersDataSegment`
+   to `ConnectionFailure::UnableToMapSendersDataSegment`
 
 7. Renamed `AttributeSet::len()`
-   into `AttributeSet::number_of_attributes()`
+   to `AttributeSet::number_of_attributes()`
 
 8. Renamed `AttributeSet::get_key_value_len()`
-   into `AttributeSet::number_of_key_values()`
+   to `AttributeSet::number_of_key_values()`
 
 9. Renamed `AttributeSet::get_key_value_at()`
-   into `AttributeSet::key_value()`
+   to `AttributeSet::key_value()`
 
 10. Renamed `AttributeSet::get_key_values()`
-   into `AttributeSet::iter_key_values()`
+   to `AttributeSet::iter_key_values()`
 
 11. Renamed `ServiceId::max_len()`
-   into `ServiceId::max_number_of_characters()`
+   to `ServiceId::max_number_of_characters()`
 
 12. Renamed the config entry `global.service.publisher-data-segment-suffix`
    to `global.service.data-segment-suffix` and changed the default value to
    `.data`.
 
-13. The following types no longer implement `Copy`
-
-14. Methods of `ServiceName` return a `ServiceNameError` instead of
+13. Methods of `ServiceName` return a `ServiceNameError` instead of
    a `SemanticStringError`
 
-15. The following types no longer implement `Copy`
+14. The following types no longer implement `Copy`
    (they only implement `Clone`):
+    * `FixedSizeByteString`
+    * `SemanticString`
+    * `Base64URL`
+    * `FileName`
+    * `FilePath`
+    * `GroupName`
+    * `UserName`
+    * `ServiceId`
 
-* `FixedSizeByteString`
-* `SemanticString`
-* `Base64URL`
-* `FileName`
-* `FilePath`
-* `GroupName`
-* `UserName`
-* `ServiceId`
+15. Renamed `AttributeVerifier::keys()` to `AttributeVerifier::required_keys()`
+
+16. Renamed `AttributeVerifier::attributes()` to `AttributeVerifier::required_attributes()`
--- a/doc/release-notes/iceoryx2-unreleased.md
+++ b/doc/release-notes/iceoryx2-unreleased.md
@@ -105,6 +105,8 @@
     [#646](https://github.com/eclipse-iceoryx/iceoryx2/issues/646)
 * Improve naming in `AttributeSet` methods and `ServiceId`
     [#697](https://github.com/eclipse-iceoryx/iceoryx2/issues/697)
+* Implement `ZeroCopySend` for `StaticConfig` and its nested types
+    [#707](https://github.com/eclipse-iceoryx/iceoryx2/issues/707)
 * Efficient `Clone` for `FixedSizeByteString`
     [#717](https://github.com/eclipse-iceoryx/iceoryx2/issues/717)
 * Use Rust's native endian swapping
@@ -257,7 +259,12 @@
    `.data`.
 
 13. The following types no longer implement `Copy`
-   (the only implement `Clone`):
+
+14. Methods of `ServiceName` return a `ServiceNameError` instead of
+   a `SemanticStringError`
+
+15. The following types no longer implement `Copy`
+   (they only implement `Clone`):
 
 * `FixedSizeByteString`
 * `SemanticString`
--- a/doc/release-notes/iceoryx2-unreleased.md
+++ b/doc/release-notes/iceoryx2-unreleased.md
@@ -280,3 +280,23 @@
     1. All arguments of methods in `Attribute`, `AttributeSpecifier`,
         `AttributeVerifier`, and `AttributeSet` changed from `&str` to
         `FixedSizeByteStrings` that implement the `SemanticString` trait
+
+    ```rust
+    // old
+    let attribute = Attribute::new("my_key", "my_value");
+
+    // new - keys or value known to fit within maximum length
+    let attribute = Attribute::new("my_key".try_into().unwrap(), "my_value".try_into().unwrap());
+
+    // new - key or value length unknown
+    fn get_key() -> &str;
+    fn get_value() -> &str;
+
+    let key = AttributeKey::try_from(get_key()).unwrap_or_else(|e| {
+        // Handle error: e.g., log error, use default key, or propagate error
+    });
+    let value = AttributeValue::try_from(get_value()).unwrap_or_else(|e| {
+        // Handle error: e.g., log error, use default value, or propagate error
+    });
+    let attribute = Attribute::new(key, value);
+    ```
--- a/iceoryx2-bb/container/src/byte_string.rs
+++ b/iceoryx2-bb/container/src/byte_string.rs
@@ -29,7 +29,6 @@
 //! println!("removed byte {}", some_string.remove(0));
 //! ```
 
-use core::str::FromStr;
 use core::{
     cmp::Ordering,
     fmt::{Debug, Display},
@@ -263,15 +262,17 @@ impl<const CAPACITY: usize, const BYTE_CAPACITY: usize> From<&[u8; BYTE_CAPACITY
     }
 }
 
-impl<const CAPACITY: usize> From<&str> for FixedSizeByteString<CAPACITY> {
-    fn from(value: &str) -> Self {
+impl<const CAPACITY: usize> TryFrom<&str> for FixedSizeByteString<CAPACITY> {
+    type Error = FixedSizeByteStringModificationError;
+
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
         if CAPACITY < value.len() {
-            fatal_panic!(from "FixedSizeByteString::from<&str>()", "The string does not fit into the FixedSizeByteString");
+            return Err(FixedSizeByteStringModificationError::InsertWouldExceedCapacity);
         }
 
         let mut new_self = Self::new();
         new_self.push_bytes(value.as_bytes()).unwrap();
-        new_self
+        Ok(new_self)
     }
 }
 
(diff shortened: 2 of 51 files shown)


iceoryx2/src/service/attribute.rs
```diff
@@ -98,26 +98,37 @@
...
-    value: String,
+    key: AttributeKeyString,
+    value: AttributeValueString,
 }
 
 impl Attribute {
     /// Acquires the service attribute key
     pub fn key(&self) -> &str {
-        &self.key
+        self.key.as_str().unwrap()
```
orecham, line 126: All of these `unwrap()` calls will be looked at more closely before this PR is merged. The solution for each case is however not immediately obvious.

Done like this as a "quick intermediate step" to get the foundations in place.

orecham: Analysed all instances of `unwrap()` - those that are there should be fine to keep in. The failure mode is prevented by invariants in the structs (i.e. construction with `&str` which guarantees valid utf-8).

iceoryx2/src/service/attribute.rs
```diff
@@ -164,14 +175,14 @@ impl AttributeSpecifier {
 /// is opened.
 #[derive(Debug)]
 pub struct AttributeVerifier {
-    attribute_set: AttributeSet,
+    required_attributes: AttributeSet,
```
orecham, line 178: @elfenpiff I renamed some things, please check it's OK with you. I find these more explicit names easier to follow in the code ...

orecham:
@elfenpiff @FerdinandSpitzschnueffler Friendly ping for a review !

elfenpiff (commented):
Looks good so far. Just some minor idiomatic things.

iceoryx2/src/service/static_config/message_type_details.rs
```diff
@@ -156,19 +177,19 @@ mod tests {
         let expected = MessageTypeDetails{
             header:  TypeDetail{
                 variant: TypeVariant::FixedSize,
-                type_name: "i32".to_string(),
+                type_name: "i32".try_into().unwrap(),
```
orecham, line 180: @elfenpiff Thoughts on use of `unwrap()` here ? It seemed like overkill to use `fatal_panic!()` for these ...

iceoryx2-bb/container/src/byte_string.rs
```diff
@@ -250,6 +262,20 @@ impl<const CAPACITY: usize, const BYTE_CAPACITY: usize> From<&[u8; BYTE_CAPACITY
...
+impl<const CAPACITY: usize> TryFrom<&str> for FixedSizeByteString<CAPACITY> {
+    type Error = FixedSizeByteStringModificationError;
+
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
+        if CAPACITY < value.len() {
+            return Err(FixedSizeByteStringModificationError::InsertWouldExceedCapacity);
+        }
+
+        let mut new_self = Self::new();
+        new_self.push_bytes(value.as_bytes()).unwrap();
```
orecham, line 274: `unwrap()` is used here because the rest of this file was already using it

iceoryx2-bb/system-types/src/file_name.rs
```diff
@@ -74,7 +75,8 @@ semantic_string! {
   normalize: normalize
 }
 
-#[derive(Debug, Clone, Eq)]
+#[derive(Debug, Clone, Eq, ZeroCopySend)]
```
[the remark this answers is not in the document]
orecham: @elfenpiff Follow-up created: https://github.com/eclipse-iceoryx/iceoryx2/issues/732

iceoryx2-bb/derive-macros/src/lib.rs
```diff
@@ -204,8 +223,80 @@ pub fn zero_copy_send_derive(input: TokenStream) -> TokenStream {
                 #type_name_impl
             },
         },
+        Data::Enum(ref data_enum) => {
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: @orecham @elfenpiff I only looked at lib.rs and the corresponding compile tests but these look good to me.

iceoryx2-bb/container/src/byte_string.rs
```diff
@@ -263,15 +262,17 @@ impl<const CAPACITY: usize, const BYTE_CAPACITY: usize> From<&[u8; BYTE_CAPACITY
...
 
-impl<const CAPACITY: usize> From<&str> for FixedSizeByteString<CAPACITY> {
-    fn from(value: &str) -> Self {
+impl<const CAPACITY: usize> TryFrom<&str> for FixedSizeByteString<CAPACITY> {
+    type Error = FixedSizeByteStringModificationError;
+
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
         if CAPACITY < value.len() {
-            fatal_panic!(from "FixedSizeByteString::from<&str>()", "The string does not fit into the FixedSizeByteString");
+            return Err(FixedSizeByteStringModificationError::InsertWouldExceedCapacity);
```
elfenpiff, line 270: Please never use `return Err(...)` directly, always use the `fail!` macro instead. This way we get for every `Err` - even when it is handled - at least a debug output.
In this case you would write:

```rust
fail!(from "FixedSizeByteString::from<&str>()",
  with FixedSizeByteStringModificationError::InsertWouldExceedCapacity,
  "The provided string \"{}\" does not fit into the FixedSizeByteString with capacity {}",
  str, capacity);
```

iceoryx2/src/service/attribute.rs
```diff
@@ -207,7 +207,12 @@ impl AttributeVerifier {
 
     /// Requires that a specific key is defined.
     pub fn require_key(mut self, key: &str) -> Self {
-        self.required_keys.push(key.into());
+        self.required_keys.push(
+            fatal_panic!(
```
elfenpiff, line 218: This is bad. In theory, it could come from some kind of user input and would lead to a fatal crash.

The option is to not use `&str` for the key argument but the `AttributeKeyString` type. The user can then use the function like:
```rust
Attribute.require_key("my key value".try_from()?);
```

iceoryx2/src/service/attribute.rs
```diff
@@ -126,9 +125,10 @@ pub struct Attribute {
 impl Attribute {
     /// Creates an attribute instance
     pub fn new(key: &str, value: &str) -> Self {
```
elfenpiff: We need to adjust the API here and use `pub fn new(key: &AttributeKeyString, value: &AttributeValueString) -> Self {`

and the user can use it like:

```rust
Attribute::new("hello".try_into()?, "world".try_into()?)
```

Otherwise a user input of a longer string would lead to a fatal crash which is a very bad idea.

iceoryx2/src/service/static_config/message_type_details.rs
elfenpiff: This is completely alright. One sees directly in the code that this is fine!

iceoryx2/src/service/attribute.rs
```diff
@@ -98,26 +98,60 @@
...
+
+use iceoryx2_bb_container::{byte_string::FixedSizeByteString, vec::FixedSizeVec};
+use iceoryx2_bb_derive_macros::ZeroCopySend;
 use iceoryx2_bb_elementary::CallbackProgression;
+use iceoryx2_bb_log::fatal_panic;
 use serde::{Deserialize, Serialize};
 
+use crate::constants::{MAX_ATTRIBUTES, MAX_ATTRIBUTE_KEY_LENGTH, MAX_ATTRIBUTE_VALUE_LENGTH};
+
+type AttributeKeyString = FixedSizeByteString<MAX_ATTRIBUTE_KEY_LENGTH>;
```
elfenpiff, line 110: Could you please turn both into a `semantic_string`, see: `iceoryx2_bb_container/semantic_string.rs`

It should be easy with the `semantic_string!` macro:

```rust
semantic_string! {
  // Name of the type
  name: AttributeKeyString,
  // The underlying capacity of the FixedSizeByteString
  capacity: MAX_ATTRIBUTE_KEY_LENGTH
  // Callable that shall return true when the provided string contains invalid content
  invalid_content: |string: &[u8]| {
    // empty keys are not allowed to be empty
    string.is_empty()
  },
  // Callable that shall return true when the provided string contains invalid characters
  invalid_characters: |string: &[u8]| {
     false
  },
  // When a SemanticString has multiple representations of the same semantic content, this
  // callable shall convert the content to a uniform representation.
  // Example: The path to `/tmp` can be also expressed as `/tmp/` or `////tmp////`
  normalize: |this: &AttributeKeyString| {
      this.clone()
  }
}
```

iceoryx2/src/service/attribute.rs
```diff
@@ -98,26 +98,60 @@
...
+    pub fn new(key: &str, value: &str) -> Result<Self, AttributeError> {
+        Ok(Self {
+            key: AttributeKeyString::try_from(key).map_err(|_| AttributeError::KeyTooLong)?,
+            value: AttributeValueString::try_from(value)
+                .map_err(|_| AttributeError::ValueTooLong)?,
+        })
+    }
+
     /// Acquires the service attribute key
     pub fn key(&self) -> &str {
```
elfenpiff, line 144: This should return a reference to `AttributeKeyString`

iceoryx2/src/service/attribute.rs
```diff
@@ -98,26 +98,60 @@
...
-        &self.key
+        fatal_panic!(from self,
+             when self.key.as_str(),
+             "This should never happen! The underlying attribute key does not contain a valid UTF-8 string.")
     }
 
     /// Acquires the service attribute value
     pub fn value(&self) -> &str {
```
elfenpiff, line 151: This should return a reference to `AttributeValueString`

iceoryx2/src/service/attribute.rs
```diff
@@ -98,26 +98,60 @@
...
+pub enum AttributeError {
+    /// The attribute key exceeds the maximum allowed length.
+    KeyTooLong,
+    /// The attribute value exceeds the maximum allowed length.
+    ValueTooLong,
 }
 
 impl Attribute {
+    /// Creates an attribute instance
+    pub fn new(key: &str, value: &str) -> Result<Self, AttributeError> {
```
elfenpiff, line 135: This here is fine as it is since it returns an `AttributeError` when the key or value is too long.

iceoryx2/src/service/attribute.rs
```diff
@@ -174,42 +208,59 @@ impl AttributeVerifier {
 
     /// Requires a value for a specific key. A key is allowed to have multiple values.
     pub fn require(mut self, key: &str, value: &str) -> Self {
```
elfenpiff, line 210: Please adjust the API to:
```rust
pub fn require(mut self, key: &AttributeKeyString, value: &AttributeValueString) -> Self {
```

iceoryx2/src/service/attribute.rs
```diff
@@ -174,42 +208,59 @@ impl AttributeVerifier {
...
-        self.attribute_set.add(key, value);
+        self.required_attributes.add(key, value);
         self
     }
 
     /// Requires that a specific key is defined.
     pub fn require_key(mut self, key: &str) -> Self {
```
elfenpiff, line 216: Please adjust the API to:

```rust
pub fn require_key(mut self, key: &AttributeKeyString) -> Self {
```

iceoryx2/src/service/attribute.rs
```diff
@@ -258,7 +302,7 @@ impl AttributeSet {
     /// Returns the number of values stored under a specific key. If the key does not exist it
     /// returns 0.
     pub fn number_of_key_values(&self, key: &str) -> usize {
```
elfenpiff, line 304: Please adjust the API to 

```rust
pub fn number_of_key_values(&self, key: &AttributeKeyString) -> usize {
```

iceoryx2/src/service/attribute.rs
elfenpiff: With this `semantic_string` macro you generate a full string class, with all methods and functions but it ensures with every modification that all contracts are uphold.

iceoryx2/src/service/attribute.rs
orecham: To be consistent with the rest of the methods, I made the arguments the fixed strings and changed the return type to `Self`.

The capacity violations will now be conveyed by the `FixedSizeByteString` arguments.

iceoryx2/src/service/attribute.rs
```diff
@@ -99,59 +99,99 @@
...
+
+    use iceoryx2_bb_container::semantic_string;
+    use iceoryx2_bb_container::semantic_string::SemanticString;
+    use iceoryx2_bb_derive_macros::ZeroCopySend;
+
+    use crate::constants::MAX_ATTRIBUTE_KEY_LENGTH;
+
+    semantic_string! {
+      /// Fixed string for service attribute keys.
+      name: FixedString,
```
elfenpiff, line 124: Since it is a string with a semantic meaning and content it should not have a generic name. Can we remove the `key` namespace and call it `AttributeKey` and the other one `AttributeValue`. All the other semantic strings are named in a similar fashion (see. `GroupName`, `UserName` or `FilePath` for instance).

iceoryx2/src/service/attribute.rs
```diff
@@ -99,59 +99,99 @@
...
 
 impl Attribute {
     /// Creates an attribute instance
-    pub fn new(key: &str, value: &str) -> Result<Self, AttributeError> {
-        Ok(Self {
-            key: AttributeKeyString::try_from(key).map_err(|_| AttributeError::KeyTooLong)?,
-            value: AttributeValueString::try_from(value)
-                .map_err(|_| AttributeError::ValueTooLong)?,
-        })
+    pub fn new(key: key::FixedString, value: value::FixedString) -> Self {
```
elfenpiff, line 183: The parameters should be references. `pub fn new(key: &key::FixedString, value: &value::FixedString) -> Self {`

iceoryx2/src/service/attribute.rs
```diff
@@ -172,7 +212,7 @@ impl AttributeSpecifier {
     }
 
     /// Defines a value for a specific key. A key is allowed to have multiple values.
-    pub fn define(mut self, key: &str, value: &str) -> Self {
+    pub fn define(mut self, key: key::FixedString, value: value::FixedString) -> Self {
```
elfenpiff, line 215: Please use references here as well.

iceoryx2/src/service/attribute.rs
```diff
@@ -207,19 +247,14 @@ impl AttributeVerifier {
     }
 
     /// Requires a value for a specific key. A key is allowed to have multiple values.
-    pub fn require(mut self, key: &str, value: &str) -> Self {
+    pub fn require(mut self, key: key::FixedString, value: value::FixedString) -> Self {
```
elfenpiff, line 250: Please use references here as well.

iceoryx2/src/service/attribute.rs
```diff
@@ -207,19 +247,14 @@ impl AttributeVerifier {
...
 
     /// Requires that a specific key is defined.
-    pub fn require_key(mut self, key: &str) -> Self {
-        self.required_keys.push(
-            fatal_panic!(
-                from "AttributeVerifier::require_key", 
-                when key.try_into(),
-                "Attempted to require an attribute key that does not fit into the underlying FixedSizeByteString")
-        );
+    pub fn require_key(mut self, key: key::FixedString) -> Self {
```
elfenpiff, line 256: Please use references here as well.

iceoryx2/src/service/attribute.rs
```diff
@@ -286,11 +321,8 @@ impl AttributeSet {
         Self(AttributeStorage::new())
     }
 
-    pub(crate) fn add(&mut self, key: &str, value: &str) {
-        self.0.push(fatal_panic!(
-                from "AttributeSet::add(key: &str, value: &str)", 
-                when Attribute::new(key, value),
-                "Attribute key or value exceeds maximum capacity"));
+    pub(crate) fn add(&mut self, key: key::FixedString, value: value::FixedString) {
```
elfenpiff, line 324: Please use references here as well.

doc/release-notes/iceoryx2-unreleased.md
```diff
@@ -275,3 +275,8 @@
 15. Renamed `AttributeVerifier::keys()` to `AttributeVerifier::required_keys()`
 
 16. Renamed `AttributeVerifier::attributes()` to `AttributeVerifier::required_attributes()`
+
+17. Attribute keys and values are now `SemanticStrings`
```
elfenpiff, line 279: Can you add a code example please with old and new like in the other cases.

iceoryx2/src/service/attribute.rs
elfenpiff: With this it would also fit better to the C++ binding where they are called `Attribute::Key` and `Attribute::Value` (they are aliases inside a class and now thinking it would have made sense to name them `AttributeKey` and `AttributeValue`) Something I will adjust!

iceoryx2/src/service/attribute.rs
orecham: @elfenpiff 

I purposely chose not to use references in all of these instances. The reasons:
1. In cases where the key or value are no longer needed by the caller, computation time (for a copy) is saved
2. In the cases where the key or value is needed for later reuse by the caller, they must explicitly call `clone()`, which makes the computation time required for the copy visible

@elBoberido Could you help break the stale mate here ? 😆

doc/release-notes/iceoryx2-unreleased.md
orecham: @elfenpiff Added an example for one of the APIs, but the change is similar (the same?) for all other instances.

iceoryx2/src/service/attribute.rs
orecham: Made aliases:

```rust
type AttributeKey = key::FixedString;
type AttributeValue = key::FixedString;
```

The modules are required due to (perhaps?) a bug in `semantic_string!` meaning it can only be used once per module.

iceoryx2/src/service/attribute.rs
elBoberido: @orecham AFAIK, moving is done via memcpy and since those types are fixed size instead of just a pointer to some heap allocated memory, it might be more expensive to move them into the function instead of passing a reference and call clone manually. I'm not 100% sure, though.

cc @elfenpiff

orecham: OK, so then we have:

1. When `AttributeKey` or `AttributeValue` are moved, it is potentially done using a memcpy (I did not consider this) which more computationally expensive than passing the reference and ensuring `clone` is used within the implementation
2. Using `clone()` has the potential to be more performant due to things like `PlacementDefault`

which means references at the API is better. I have updated the code.

@elfenpiff @elBoberido

elfenpiff (approved):

PR #801, opened 2025-05-30
[#593] Initial Zenoh Tunnel

### LINKED ISSUE #593: tunnel
## Brief feature description

We require a tunnel to connect multiple iceoryx2 instances on different machines with each other. The difference to a gateway is, that the payload of the messages will contain the underlying iceoryx2 protocol and are not readable by native instances - it is just a binary blob for them.

### Requirements

 * The gateway shall be configurable
   * it comes with a allow/deny list - services can automatically be shared with the world or are explicitly excluded
   * the configuration is in TOML - like the rest of the iceoryx2 configuration
   * a gateway config shall also contain protocol specific configurations
   * the config shall contain options for peer2peer as well as multicast communications
     * peer2peer example: provide a list of all iceoryx2 endpoints with ip/port
* If possible, the gateway shall not come with extra background threads. The user shall be able to handle everything in the main thread.
  * Certain protocol enforce background threads, but our implementation shall not
* The gateway shall be traffic efficient
  * only publish things that are required from an endpoint in the network
* Gateway shall be usable right out-of-the-box
  * Sane default configuration
  * iceoryx2 provides a service discovery service that is provided by every gateway. this allows other gateways to discover local services
* Service Discovery is written so that it does not require a background thread
   * it shall be split into two parts: 1. the library that performs the service discovery, 2. a binary that can be started and provides the service discovery service
   * the library shall be usable from within an arbitrary application. in the main loop one can call the loop callback of the service discovery to update the internal state and send out updates
   * publish-subscribe: service diff details like: service X, new publisher
   * request-response: to provide full service state
   * event: to notify others about service diff details
   * requires a user side struct for convenience to handle the service diffs(pubsub), state(rpc), events with one convenenient API
* so that samples are not forwarded in circle we need to be able to determine the machine origin of a sample
   * (if possible) find a way to generate a machine unique id automatically and use it in the gateway, otherwise the user must provide a unique id
   * the pubsub header must be extended so that the id can be stored there
* gateways and service discovery must be able to support multiple service styles:
   * ipc::Service, local::Service and future gpu::Service or hypervisor::Service
   * gateways shall support different iceoryx2 configs, see: https://github.com/eclipse-iceoryx/iceoryx2/tree/main/config
   and https://docs.rs/iceoryx2/latest/iceoryx2/config/index.html

### Next Steps

 - [ ] implement service discovery
    * use `Service::list()` to acquire all static service infos, see: https://docs.rs/iceoryx2/latest/iceoryx2/service/trait.Service.html#method.list
    * connect to service to acquire dynamic service info, see:
```rust
// event
let service = self.node.service_builder(element.static_details.name()).event().open().unwrap();
let dynamic_config = service.dynamic_config();

// pubsub
ipc::Service::list(Config::global_config(), |element| {
match element.static_details.messaging_pattern() {
iceoryx2::service::static_config::messaging_pattern::MessagingPattern::PublishSubscribe(config) => {
if let Ok(service) = unsafe { self.node.service_builder(element.static_details.name())
                        .publish_subscribe::<[CustomPayloadMarker]>()
                        .user_header::<CustomHeaderMarker>()
                        .__internal_set_payload_type_details(&config.message_type_details().payload)
                        .__internal_set_user_header_type_details(&config.message_type_details().user_header)
                        .open() } {
                        let dynamic_config = service.dynamic_config();
```
 * event, use: https://docs.rs/iceoryx2/latest/iceoryx2/service/dynamic_config/event/struct.DynamicConfig.html
 * pubsub, use: https://docs.rs/iceoryx2/latest/iceoryx2/service/dynamic_config/publish_subscribe/index.html
 * The internal but public methods acquire more details:
     * pubsub: `dynamic_config.__internal_list_subscribers()`, `dynamic_config.__internal_list_publishers()`
     * event: `dynamic_config.__internal_list_listeners()`, `dynamic_config.__internal_list_notifiers()`

orecham:
## Notes for Reviewer

Foundation implementation of a `Tunnel` that utilizes `zenoh` to to bridge `iceoryx2` instances on different hosts. **This implementation is not yet optimized and has yet to undergo testing in realistic scenarios.**

This PR is the **first step** towards a fully functional tunnel. Includes:
1. Polling execution mode
1. Propagation of `PublishSubscribe` services
1. Propagation of `Event` services
1. Ability to embed execution in user threads
1. CLI to execute in a separate process

**Next Steps:**

1. Propagation of `RequestResponse` services
1. Whitelist peers to connect with
1. Whitelist services to tunnel
1. Reactive execution mode

**Usage:**

```
cargo build -p iceoryx2-cli --bin iox2-tunnel

cargo run --bin iox2 -- tunnel-debug zenoh --help      # see CLI options
cargo run --bin iox2 -- tunnel-debug zenoh             # tunnel over zenoh with default options
```

* [ ] Assign PR to reviewer

## References

Closes #593

### DIFF
--- a/iceoryx2-cli/Cargo.toml
+++ b/iceoryx2-cli/Cargo.toml
@@ -39,7 +39,6 @@ name = "iceoryx2_cli"
 path = "lib/src/lib.rs"
 
 [dependencies]
-iceoryx2-tunnels-zenoh = { workspace = true }
 iceoryx2-services-discovery = { workspace = true }
 iceoryx2 = { workspace = true }
 iceoryx2-bb-log = { workspace = true }
@@ -64,6 +63,10 @@ ron = { workspace = true }
 toml = { workspace = true }
 dirs = { workspace = true }
 dialoguer = { workspace = true }
+
+# Zenoh does not build on FreeBSD
+[target.'cfg(not(target_os = "freebsd"))'.dependencies]
+iceoryx2-tunnels-zenoh = { workspace = true }
 zenoh = { workspace = true }
 
 [dev-dependencies]
--- a/iceoryx2-cli/iox2-tunnel/src/main.rs
+++ b/iceoryx2-cli/iox2-tunnel/src/main.rs
@@ -10,92 +10,110 @@
 //
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-#[cfg(not(debug_assertions))]
-use human_panic::setup_panic;
-#[cfg(debug_assertions)]
-extern crate better_panic;
+#[cfg(target_os = "freebsd")]
+fn main() {
+    println!("Not supported on this platform");
+}
 
+#[cfg(not(target_os = "freebsd"))]
 mod cli;
 
-use anyhow::Result;
-use clap::Parser;
-use cli::Cli;
-use cli::Transport;
+#[cfg(not(target_os = "freebsd"))]
+mod supported_platform {
 
-use iceoryx2::prelude::*;
-use iceoryx2_bb_log::error;
-use iceoryx2_bb_log::info;
-use iceoryx2_bb_log::set_log_level_from_env_or;
-use iceoryx2_bb_log::LogLevel;
-
-use iceoryx2_tunnels_zenoh::Scope;
-use iceoryx2_tunnels_zenoh::Tunnel;
-use iceoryx2_tunnels_zenoh::TunnelConfig;
-
-fn main() -> Result<()> {
     #[cfg(not(debug_assertions))]
-    {
-        setup_panic!();
-    }
+    use human_panic::setup_panic;
     #[cfg(debug_assertions)]
-    {
-        better_panic::Settings::debug()
-            .most_recent_first(false)
-            .lineno_suffix(true)
-            .verbosity(better_panic::Verbosity::Full)
-            .install();
-    }
+    extern crate better_panic;
 
-    set_log_level_from_env_or(LogLevel::Warn);
+    use super::cli;
 
-    let cli = match Cli::try_parse() {
-        Ok(cli) => cli,
-        Err(e) => {
-            // --help and --version is treated as a parse error by clap
-            // printing the error is actually printing the result of those commands ...
-            let _ = e.print();
-            return Ok(());
+    use anyhow::Result;
+    use clap::Parser;
+    use cli::Cli;
+    use cli::Transport;
+
+    use iceoryx2::prelude::*;
+    use iceoryx2_bb_log::error;
+    use iceoryx2_bb_log::info;
+    use iceoryx2_bb_log::set_log_level_from_env_or;
+    use iceoryx2_bb_log::LogLevel;
+
+    use iceoryx2_tunnels_zenoh::Scope;
+    use iceoryx2_tunnels_zenoh::Tunnel;
+    use iceoryx2_tunnels_zenoh::TunnelConfig;
+
+    pub fn main() -> Result<()> {
+        #[cfg(not(debug_assertions))]
+        {
+            setup_panic!();
         }
-    };
+        #[cfg(debug_assertions)]
+        {
+            better_panic::Settings::debug()
+                .most_recent_first(false)
+                .lineno_suffix(true)
+                .verbosity(better_panic::Verbosity::Full)
+                .install();
+        }
 
-    if let Some(transport) = cli.transport {
-        match transport {
-            Transport::Zenoh(_zenoh_options) => {
-                let tunnel_config = TunnelConfig {
-                    discovery_service: cli.discovery_service,
-                };
-                let iox_config = iceoryx2::config::Config::default();
-                let zenoh_config = zenoh::Config::default();
+        set_log_level_from_env_or(LogLevel::Warn);
 
-                let mut tunnel =
-                    Tunnel::<ipc::Service>::create(&tunnel_config, &iox_config, &zenoh_config)?;
-                let waitset = WaitSetBuilder::new().create::<ipc::Service>()?;
+        let cli = match Cli::try_parse() {
+            Ok(cli) => cli,
+            Err(e) => {
+                // --help and --version is treated as a parse error by clap
+                // printing the error is actually printing the result of those commands ...
+                let _ = e.print();
+                return Ok(());
+            }
+        };
 
-                if cli.reactive {
-                    // TODO(functionality): Make tunnel (or its endpoints) attachable to waitset
-                    unimplemented!("Reactive mode is not yet supported.");
-                } else {
-                    let rate = cli.poll.unwrap_or(100);
-                    info!("Polling rate {}ms", rate);
-
-                    let guard = waitset.attach_interval(core::time::Duration::from_millis(rate))?;
-                    let tick = WaitSetAttachmentId::from_guard(&guard);
-
-                    let on_event = |id: WaitSetAttachmentId<ipc::Service>| {
-                        if id == tick {
-                            if let Err(e) = tunnel.discover(Scope::Both) {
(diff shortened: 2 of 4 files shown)


iceoryx2-tunnels/zenoh/tests/tunnel_tests.rs
```diff
@@ -314,6 +314,9 @@ mod zenoh_tunnel {
             Some(MAX_RETRIES),
         );
 
+        // Wait for Zenoh's backgorund thread to establish match...
+        std::thread::sleep(Duration::from_millis(100));
```
elfenpiff: I would try to avoid waiting/sleeping and introduce here a barrier if possible to blocking wait until the match is established.

To handle potential deadlocks you can use the [WatchDog](https://github.com/eclipse-iceoryx/iceoryx2/blob/main/iceoryx2-pal/testing/src/watchdog.rs) which terminates the tests when it runs more than 10 seconds (or a user specified time).

elfenpiff: Maybe a infinite loop with a sleep inside the loop and check via the zenoh API if the match is established. Otherwise we create here a flaky tests which crashes whenever the CI is under load or increases massively the CI runtime.

Again, the watchdog will take care of the blocking calls and terminates the tests when it runs into a deadlock

orecham: @elfenpiff Yeah this is definitely not nice, but it was able to prove what I thought was previously causing the test failures.

The following is happening:

1. Two gateway instances are created
2. Each instance has their own zenoh "session"
3. One gateway creates a zenoh publisher (to propagate data from `iceoryx2`)
4. The other gateway detects this zenoh publisher, creates a zenoh subscriber to match with it (to propagate data to `iceoryx2`)
5. The match does not occur immediately as it depends on the zenoh background thread waking up to establish the match
6. If a sample is sent on the zenoh publisher (from step 3) before the match is set up, the test will fail

There is one more potential option I have found, but it involves exposing the zenoh endpoints within tunnel via the API. The zenoh publisher offers an API to check the matching status of the publisher (https://docs.rs/zenoh/latest/zenoh/pubsub/struct.Publisher.html#method.matching_status).

I hesitated to expose this because it's functionality that would only really be needed for testing. Do you see any other options?

> check via the zenoh API if the match is established

Ah this sparked an idea (or maybe this is what you were advising), maybe it's possible to check independently from the endpoints. Will look into it ...

iceoryx2-cli/iox2-tunnel/src/cli.rs
```diff
@@ -0,0 +1,70 @@
...
+use iceoryx2_cli::help_template;
+use iceoryx2_cli::HelpOptions;
+
+#[derive(Parser)]
+#[command(
+    name = "iox2 tunnel",
+    bin_name = "iox2 tunnel",
+    about = "Launch a tunnel between iceoryx2 instances.",
+    long_about = None,
+    version = env!("CARGO_PKG_VERSION"),
```
elfenpiff, line 25: This sadly does not work with Bazel; therefore, we created this here: https://github.com/eclipse-iceoryx/iceoryx2/blob/main/iceoryx2-bb/elementary/src/package_version.rs

iceoryx2-tunnels/end-to-end-testing/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "iceoryx2-tunnels-end-to-end-testing"
+description = "iceoryx2: binaries for tunnel end-to-end testing"
```
elfenpiff: Should we write `iceoryx2: [internal] binaries ...` to make it clear that this is not public API?

My idea was that every crate which is not explicitly tagged with `[internal` is public API and must follow semver but if a user uses internal API we provide no guarantees whatsoever

iceoryx2-cli/iox2-tunnel/src/cli.rs
orecham: https://github.com/eclipse-iceoryx/iceoryx2/issues/816

iceoryx2-cli/Cargo.toml
```diff
@@ -64,6 +63,10 @@ ron = { workspace = true }
 toml = { workspace = true }
 dirs = { workspace = true }
 dialoguer = { workspace = true }
+
+# Zenoh does not build on FreeBSD
+[target.'cfg(not(target_os = "freebsd"))'.dependencies]
```
elfenpiff, line 68: Is it possible to add this also to the global `Cargo.toml` so that we do not have to explicitly exclude the tunnel when we build on FreeBSD?

iceoryx2-cli/iox2-tunnel/src/main.rs
```diff
@@ -10,92 +10,110 @@
...
-use iceoryx2::prelude::*;
-use iceoryx2_bb_log::error;
-use iceoryx2_bb_log::info;
-use iceoryx2_bb_log::set_log_level_from_env_or;
-use iceoryx2_bb_log::LogLevel;
-
-use iceoryx2_tunnels_zenoh::Scope;
-use iceoryx2_tunnels_zenoh::Tunnel;
-use iceoryx2_tunnels_zenoh::TunnelConfig;
+#[cfg(not(target_os = "freebsd"))]
```
elfenpiff, line 21: Why this hop with the `supported_platform`. Would it be possible to have a `main` for non-supported platforms like the one above and another main for the supported ones.

This would simplify the code a bit.

internal/scripts/ci_build_and_test_freebsd.sh
```diff
@@ -75,19 +75,19 @@ export PATH=$PATH:$HOME/.cargo/bin
 rustup default $RUST_TOOLCHAIN
 # export RUSTFLAGS="-C debug-assertions"
 cargo fmt --all -- --check
-cargo clippy -- -D warnings
+cargo clippy --workspace --exclude iceoryx2-tunnels-zenoh -- -D warnings
```
elfenpiff, line 78: If the `cfg` thing works in the root `Cargo.toml` we could remove this again ... Otherwise we have to adjust also the readme a bit and point out that on freebsd the build process is a bit different and it would be great if we could avoid this.

iceoryx2-cli/Cargo.toml
orecham: @elfenpiff Not possible to do in workspaces:

```
error: failed to parse manifest at `/Users/orecham/Code/iceoryx2/Cargo.toml`

Caused by:
  this virtual manifest specifies a `target` section, which is not allowed
```

iceoryx2-cli/iox2-tunnel/src/main.rs
orecham: @elfenpiff It's to contain the dependencies (`use` call)

internal/scripts/ci_build_and_test_freebsd.sh
orecham: @elfenpiff AFAIK it does not work.

elfenpiff (approved):

internal/scripts/ci_build_and_test_freebsd.sh
orecham: @elBoberido Thoughts on this? Should more effort be spent to find another solution (do you happen to know one?)

elBoberido: Unfortunately, I also do not have have another solution, except to not build the zenoh tunnel by default but only if a cfg option is set. Similar to what I proposed here https://github.com/eclipse-iceoryx/iceoryx2/pull/807#discussion_r2138318492

But I'm not sure if it could easily be applied to a whole crate

PR #805, opened 2025-06-06
[#777] Request current discovery state from discovery service

### LINKED ISSUE #777: Request current discovery state from discovery service
## Brief Description

1. The [discovery service](https://github.com/eclipse-iceoryx/iceoryx2/tree/main/iceoryx2-services/discovery) currently publishes discovery updates to connected subscribers
2. Late joining subscribers can only receive updates as far back as the configured `history` of the publisher

It should be made possible for participants to get in sync with the entire discovery state at any time.

## Improvement Suggestion

The discovery service should create a `RequestResponse` service for participants to request the entire discovery state.

RTK-jangid:
## Notes for Reviewer

- Added a `RequestResponse` server in `discovery_service.rs` that responds with a list of all currently tracked services.
- This allows clients to query the discovery layer for runtime visibility into what services are currently active.
- Added a new example client under `examples/discovery_service/discovery_service_client.rs` that connects to the server,sends a query, and prints the discovered services.

* [x] Assign PR to reviewer

## References

Closes #777

### DIFF
--- a/examples/rust/discovery_service/discovery_service_client.rs
+++ b/examples/rust/discovery_service/discovery_service_client.rs
@@ -28,12 +28,12 @@ fn main() -> Result<(), Box<dyn core::error::Error>> {
     let client = service.client_builder().create()?;
 
     let waitset = WaitSetBuilder::new().create::<ipc::Service>()?;
-    let guard_1 = waitset.attach_interval(CYCLE_TIME)?;
+    let guard = waitset.attach_interval(CYCLE_TIME)?;
 
     let pending_response = client.send_copy(())?;
 
     let on_event = |attachment_id: WaitSetAttachmentId<ipc::Service>| {
-        if attachment_id.has_event_from(&guard_1) {
+        if attachment_id.has_event_from(&guard) {
             while let Some(response) = pending_response.receive().unwrap() {
                 for service in response.payload().iter() {
                     println!("Service ID: {:?}", service.service_id().as_str());
--- a/examples/rust/discovery_service/discovery_service_client.rs
+++ b/examples/rust/discovery_service/discovery_service_client.rs
@@ -27,7 +27,7 @@ fn main() -> Result<(), Box<dyn core::error::Error>> {
 
     let client = service.client_builder().create()?;
 
-    let pending_response = client.loan_uninit()?.write_payload(()).send()?;
+    let pending_response = client.send_copy(())?;
 
     // IMPORTANT: We need to wait for the request to be sent before we can receive responses.
     while !node.wait(CYCLE_TIME).is_ok() {}
--- a/examples/rust/discovery_service/discovery_service_client.rs
+++ b/examples/rust/discovery_service/discovery_service_client.rs
@@ -14,7 +14,7 @@ use core::time::Duration;
 use iceoryx2::{prelude::*, service::static_config::StaticConfig};
 use iceoryx2_services_discovery::service_discovery::service_name;
 
-const CYCLE_TIME: Duration = Duration::from_secs(1);
+const CYCLE_TIME: Duration = Duration::from_millis(10);
 
 fn main() -> Result<(), Box<dyn core::error::Error>> {
     set_log_level_from_env_or(LogLevel::Info);
@@ -27,20 +27,27 @@ fn main() -> Result<(), Box<dyn core::error::Error>> {
 
     let client = service.client_builder().create()?;
 
+    let waitset = WaitSetBuilder::new().create::<ipc::Service>()?;
+    let guard_1 = waitset.attach_interval(CYCLE_TIME)?;
+
     let pending_response = client.send_copy(())?;
 
-    // IMPORTANT: We need to wait for the request to be sent before we can receive responses.
-    while !node.wait(CYCLE_TIME).is_ok() {}
-
-    while let Some(response) = pending_response.receive()? {
-        for service in response.payload().iter() {
-            println!("Service ID: {:?}", service.service_id().as_str());
-            println!("Service Name: {:?}", service.name().as_str());
-            println!();
+    let on_event = |attachment_id: WaitSetAttachmentId<ipc::Service>| {
+        if attachment_id.has_event_from(&guard_1) {
+            while let Some(response) = pending_response.receive().unwrap() {
+                for service in response.payload().iter() {
+                    println!("Service ID: {:?}", service.service_id().as_str());
+                    println!("Service Name: {:?}", service.name().as_str());
+                    println!();
+                }
+                println!("exit");
+                return CallbackProgression::Stop;
+            }
         }
-    }
+        CallbackProgression::Continue
+    };
 
-    println!("exit");
+    waitset.wait_and_process(on_event)?;
 
     Ok(())
 }
--- a/examples/rust/discovery_service/discovery_service_client.rs
+++ b/examples/rust/discovery_service/discovery_service_client.rs
@@ -1,0 +1,46 @@
+// Copyright (c) 2025 Contributors to the Eclipse Foundation
+//
+// See the NOTICE file(s) distributed with this work for additional
+// information regarding copyright ownership.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Apache Software License 2.0 which is available at
+// https://www.apache.org/licenses/LICENSE-2.0, or the MIT license
+// which is available at https://opensource.org/licenses/MIT.
+//
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use core::time::Duration;
+use iceoryx2::{prelude::*, service::static_config::StaticConfig};
+use iceoryx2_services_discovery::service_discovery::service_name;
+
+const CYCLE_TIME: Duration = Duration::from_secs(1);
+
+fn main() -> Result<(), Box<dyn core::error::Error>> {
+    set_log_level_from_env_or(LogLevel::Info);
+    let node = NodeBuilder::new().create::<ipc::Service>()?;
+
+    let service = node
+        .service_builder(service_name().try_into()?)
+        .request_response::<(), [StaticConfig]>()
+        .open_or_create()?;
+
+    let client = service.client_builder().create()?;
+
+    let pending_response = client.loan_uninit()?.write_payload(()).send()?;
+
+    // IMPORTANT: We need to wait for the request to be sent before we can receive responses.
+    while !node.wait(CYCLE_TIME).is_ok() {}
+
+    while let Some(response) = pending_response.receive()? {
+        for service in response.payload().iter() {
+            println!("Service ID: {:?}", service.service_id().as_str());
+            println!("Service Name: {:?}", service.name().as_str());
+            println!();
+        }
+    }
+
+    println!("exit");
+
+    Ok(())
+}
--- a/iceoryx2-services/discovery/src/service_discovery/service.rs
+++ b/iceoryx2-services/discovery/src/service_discovery/service.rs
@@ -27,6 +27,7 @@ use iceoryx2::{
         builder::{
             event::EventOpenOrCreateError, publish_subscribe::PublishSubscribeOpenOrCreateError,
         },
+        port_factory::request_response::PortFactory,
         static_config::StaticConfig,
         Service as ServiceType, ServiceDetails,
     },
@@ -301,11 +302,13 @@ impl Default for Config {
 /// # Type Parameters
 ///
 /// * `S` - The service type that this discovery service operates on.
+#[allow(dead_code)]
 pub struct Service<S: ServiceType> {
     discovery_config: Config,
     iceoryx_config: IceoryxConfig,
     _node: Node<S>,
     publisher: Option<Publisher<S, Payload, ()>>,
+    request_response: Option<PortFactory<S, (), (), [StaticConfig], ()>>,
     server: Option<Server<S, (), (), [StaticConfig], ()>>,
     notifier: Option<Notifier<S>>,
     tracker: Tracker<S>,
@@ -365,15 +368,14 @@ impl<S: ServiceType> Service<S> {
             tracker.sync(iceoryx_config)?;
         }
 
-        let mut server = None;
-        if discovery_config.enable_server {
+        let (request_response, server) = if discovery_config.enable_server {
             let request_response = node
                 .service_builder(service_name())
                 .request_response::<(), [StaticConfig]>()
                 .open_or_create()
                 .map_err(|_| CreationError::ServiceCreationFailure)?;
 
-            server = Some(
+            let server = Some(
                 request_response
                     .server_builder()
                     .initial_max_slice_len(discovery_config.initial_max_slice_len)
@@ -381,13 +383,17 @@ impl<S: ServiceType> Service<S> {
                     .create()
                     .map_err(|_| CreationError::ServiceCreationFailure)?,
             );
-        }
+            (Some(request_response), server)
+        } else {
+            (None, None)
+        };
 
         Ok(Service::<S> {
             discovery_config: discovery_config.clone(),
             iceoryx_config: iceoryx_config.clone(),
             _node: node,
             publisher,
+            request_response,
             server,
             notifier,
             tracker,
--- a/iceoryx2-services/discovery/src/service_discovery/service.rs
+++ b/iceoryx2-services/discovery/src/service_discovery/service.rs
@@ -265,6 +265,12 @@ pub struct Config {
 
     /// The maximum number of listeners to the service permitted.
     pub max_listeners: usize,
+
+    /// Whether to enable the server for handling requests.
+    pub enable_server: bool,
+
+    /// The initial maximum slice length for the server.
+    pub initial_max_slice_len: usize,
 }
 
 impl Default for Config {
@@ -280,6 +286,8 @@ impl Default for Config {
             max_borrrowed_samples: defaults.publish_subscribe.subscriber_max_borrowed_samples,
             send_notifications: true,
             max_listeners: defaults.event.max_listeners,
+            enable_server: true,
+            initial_max_slice_len: 10,
         }
     }
 }
@@ -356,16 +364,24 @@ impl<S: ServiceType> Service<S> {
         if discovery_config.sync_on_initialization {
             tracker.sync(iceoryx_config)?;
         }
-        let server = Some(
-            node.service_builder(service_name())
+
+        let mut server = None;
+        if discovery_config.enable_server {
+            let request_response = node
+                .service_builder(service_name())
                 .request_response::<(), [StaticConfig]>()
                 .open_or_create()
-                .map_err(|_| CreationError::ServiceCreationFailure)?
-                .server_builder()
-                .allocation_strategy(AllocationStrategy::PowerOfTwo)
-                .create()
-                .map_err(|_| CreationError::ServiceCreationFailure)?,
-        );
+                .map_err(|_| CreationError::ServiceCreationFailure)?;
+
+            server = Some(
+                request_response
+                    .server_builder()
+                    .initial_max_slice_len(discovery_config.initial_max_slice_len)
+                    .allocation_strategy(AllocationStrategy::PowerOfTwo)
+                    .create()
+                    .map_err(|_| CreationError::ServiceCreationFailure)?,
+            );
+        }
 
         Ok(Service::<S> {
             discovery_config: discovery_config.clone(),
--- a/iceoryx2-services/discovery/src/service_discovery/service.rs
+++ b/iceoryx2-services/discovery/src/service_discovery/service.rs
@@ -11,13 +11,15 @@
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
 use crate::service_discovery::{SyncError, Tracker};
-use iceoryx2::prelude::ZeroCopySend;
+use iceoryx2::port::ReceiveError;
+use iceoryx2::prelude::{AllocationStrategy, ZeroCopySend};
 use iceoryx2::{
     config::Config as IceoryxConfig,
     node::{Node, NodeBuilder, NodeCreationFailure},
     port::{
         notifier::{Notifier, NotifierCreateError, NotifierNotifyError},
         publisher::{Publisher, PublisherCreateError},
+        server::Server,
         LoanError, SendError,
     },
     prelude::ServiceName,
@@ -147,6 +149,9 @@ pub enum SpinError {
 
     /// Failed to send a notification about service changes.
     NotifyFailure,
+
+    /// Server error while requesting, receiving or loaning services.
+    ServerSpinError(ServerSpinError),
 }
 
 impl core::fmt::Display for SpinError {
@@ -181,6 +186,51 @@ impl From<SendError> for SpinError {
 impl From<NotifierNotifyError> for SpinError {
     fn from(_: NotifierNotifyError) -> Self {
         SpinError::NotifyFailure
+    }
+}
+
+impl From<ServerSpinError> for SpinError {
+    fn from(error: ServerSpinError) -> Self {
+        SpinError::ServerSpinError(error)
+    }
+}
+
+/// Errors that can occur when running the server of the service discovery_service.
+#[derive(Debug, Eq, PartialEq, Clone, Copy)]
+pub enum ServerSpinError {
+    /// The caller does not have sufficient permissions to create the service.
+    ReceptionFailure,
+
+    /// Failed to send a response to the client.
+    ResponseSendFailure,
+
+    /// Failed to loan a response sample.
+    LoanFailure,
+}
+
+impl core::fmt::Display for ServerSpinError {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        std::write!(f, "ServerSpinError::{:?}", self)
+    }
+}
+
+impl core::error::Error for ServerSpinError {}
+
+impl From<ReceiveError> for ServerSpinError {
+    fn from(_: ReceiveError) -> Self {
+        ServerSpinError::ReceptionFailure
+    }
+}
+
+impl From<SendError> for ServerSpinError {
+    fn from(_: SendError) -> Self {
+        ServerSpinError::ResponseSendFailure
+    }
+}
+
+impl From<LoanError> for ServerSpinError {
+    fn from(_: LoanError) -> Self {
+        ServerSpinError::LoanFailure
     }
 }
 
@@ -248,6 +298,7 @@ pub struct Service<S: ServiceType> {
     iceoryx_config: IceoryxConfig,
     _node: Node<S>,
     publisher: Option<Publisher<S, Payload, ()>>,
+    server: Option<Server<S, (), (), [StaticConfig], ()>>,
     notifier: Option<Notifier<S>>,
     tracker: Tracker<S>,
 }
@@ -305,12 +356,23 @@ impl<S: ServiceType> Service<S> {
         if discovery_config.sync_on_initialization {
             tracker.sync(iceoryx_config)?;
         }
+        let server = Some(
+            node.service_builder(service_name())
+                .request_response::<(), [StaticConfig]>()
+                .open_or_create()
+                .map_err(|_| CreationError::ServiceCreationFailure)?
+                .server_builder()
+                .allocation_strategy(AllocationStrategy::PowerOfTwo)
+                .create()
+                .map_err(|_| CreationError::ServiceCreationFailure)?,
+        );
 
         Ok(Service::<S> {
             discovery_config: discovery_config.clone(),
             iceoryx_config: iceoryx_config.clone(),
             _node: node,
             publisher,
+            server,
             notifier,
             tracker,
         })
@@ -346,7 +408,6 @@ impl<S: ServiceType> Service<S> {
         // Detect changes
         let (added_ids, removed_services) = self.tracker.sync(&self.iceoryx_config)?;
         let changes_detected = !added_ids.is_empty() || !removed_services.is_empty();
-
         // Publish
         for id in &added_ids {
             if let Some(service) = self.tracker.get(id) {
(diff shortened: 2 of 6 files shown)


iceoryx2-services/discovery/src/service_discovery/service.rs
```diff
@@ -304,11 +306,22 @@ impl<S: ServiceType> Service<S> {
             tracker.sync(iceoryx_config)?;
         }
 
+        let server = Some(node
+            .service_builder(service_name())
+            .request_response::<u8, [StaticConfig]>()
```
orecham: This is one possible design. This utilizes dynamic allocation, which is probably OK for this service since it is not intended to be a safety-critical component. Fine for this PR.

---

**Info Only:**
Another idea (more in-line with the safety-critical use-cases that `iceoryx2` targets) could be to use the payload type of `StaticConfig` (instead of `[StaticConfig]`), allow the [response buffer size](https://github.com/eclipse-iceoryx/iceoryx2/blob/main/iceoryx2/src/service/mod.rs#L66) to be configurable, and send multiple `StaticConfig` in response to a single request until the entire state has been shared.
If the buffer were to be filled, the request could remain active (up to a point), and more samples could be sent at a later time (e.g. in subsequent `spin()` operations).

This design achieves the functionality with a fixed-size memory buffer and thus lends itself to safety-critical use-cases.

iceoryx2-services/discovery/src/service_discovery/service.rs
```diff
@@ -387,6 +400,62 @@ impl<S: ServiceType> Service<S> {
 
         Ok(())
     }
+    /// Returns the service details of the discovery service.
+    pub fn start_server(&mut self)-> Result<(), Box<dyn core::error::Error>> {
```
orecham: The existing design expects that the `spin()` method is called periodically, in which all required operations are done. This logic should be moved into there.

I would split the different steps into methods and have the `spin()` method call them.

Thus, the behaviour would be that, for each iteration (`spin()` call):
* The tracker is updated
* Newly added/removed services are communicated via the `Publisher`
* Pending `Server` requests are responded to

iceoryx2-services/discovery/src/service_discovery/service.rs
```diff
@@ -387,6 +400,62 @@ impl<S: ServiceType> Service<S> {
...
+                    // for (idx, service) in services.iter().enumerate() {
+                    //     println!("  service {}", idx);
+                    //     let response = active_request.loan_uninit()?;
+                    //     let response = response.write_payload(service.clone());
+                    //     while(!self._node.wait(CYCLE_TIME).is_ok()){}
+                    //     response.send()?;
+                    // }
+
+                    // Allocate space for all StaticConfig entries
+                    use std::panic::{catch_unwind, AssertUnwindSafe};
```
orecham: Are these just for some debugging?

iceoryx2-services/discovery/src/service_discovery/service.rs
```diff
@@ -246,6 +247,7 @@ pub struct Service<S: ServiceType> {
     iceoryx_config: IceoryxConfig,
     _node: Node<S>,
     publisher: Option<Publisher<S, Payload, ()>>,
+    server: Option<Server<S, u8, (), [StaticConfig], ()>>,
```
orecham: What is the `u8` for ?

RTK-jangid:
@orecham I have updated the PR. Please review

test was passed with `max_buffer_response: 10` and fails with `max_buffer_response: 3` as total `num_services = 6`
Screenshot from 2025-06-07 14-27-55

iceoryx2-services/discovery/src/service_discovery/service.rs
```diff
@@ -385,6 +451,39 @@ impl<S: ServiceType> Service<S> {
...
         }
 
+        // Handle server requests
+        self.start_server()?;
+
+        Ok(())
+    }
+
+
+    /// Returns the service details of the discovery service.
```
orecham: Documentation style is not consistent with the rest of this file.

iceoryx2-services/discovery/src/service_discovery/service.rs
```diff
@@ -385,6 +451,39 @@ impl<S: ServiceType> Service<S> {
...
+    ///     /// # Returns
+    ///
+    /// A result containing `()` if successful.
+    ///     /// # Errors
+    ///
+    /// Returns a `SpinError` if there was an error publishing events or sending
+    /// notifications.
+    pub fn start_server(&mut self)-> Result<(), ServerSpinError> {
```
orecham:
```suggestion
    pub fn handle_requests(&mut self)-> Result<(), ServerSpinError> {
```
(Or choose another name yourself)
Current name is misleading, nothing is being "started".

iceoryx2-services/discovery/src/service_discovery/service.rs
```diff
@@ -246,6 +300,7 @@ pub struct Service<S: ServiceType> {
     iceoryx_config: IceoryxConfig,
     _node: Node<S>,
     publisher: Option<Publisher<S, Payload, ()>>,
+    server: Option<Server<S, (), (), StaticConfig, ()>>,
```
orecham: I see you decided to change to a non-dynamic implementation. Cool.

In this case, it is worth adding a test for the behaviour of the server where `max_response_buffer_size` is exceeded before a request could be fulfilled. The behaviour would be something like:

1. Start some services
2. Create a discovery service with `max_response_buffer_size` < number of services started
3. Make a request
4. Spin the server once to receive some responses
5. Verify that the request is not yet completed (using `PendingResponse::is_connected()`)
6. Spin the server until the point where all responses should be received
7. Verify all services are received
8. Verify that the request is completed  (using `PendingResponse::is_connected()`)

iceoryx2-services/discovery/src/service_discovery/service.rs
```diff
@@ -385,6 +451,39 @@ impl<S: ServiceType> Service<S> {
...
+    ///     /// # Returns
+    ///
+    /// A result containing `()` if successful.
+    ///     /// # Errors
+    ///
+    /// Returns a `SpinError` if there was an error publishing events or sending
+    /// notifications.
+    pub fn start_server(&mut self)-> Result<(), ServerSpinError> {
+        if let Some(server) = self.server.as_mut() {
+            while let Some(active_request) = server.receive()? {
```
orecham, line 495: Is the case where the response buffer is full being handled?

iceoryx2-cli/iox2-service/src/commands.rs
```diff
@@ -238,4 +238,4 @@ pub fn discovery(
         .map_err(|e| anyhow!("error waiting on waitset: {:?}", e))?;
 
     Ok(())
-}
+}
```
orecham: Revert this to avoid polluting file history.

orecham (commented):
Looking good. Just need to verify that the one edge case is handled, then it should be good to merge.

Edit: There are a lot of formatting issues caught by CI - are you using `rust_analyzer` to keep the formatting consistent?

iceoryx2-services/discovery/src/service_discovery/service.rs
orecham: Another edge cases:

* How should the server react if the response buffer remains full ?

If you can think of any others, tests for those behaviours should be added too. Maybe some are more obvious to you as the implementor.

RTK-jangid:
@orecham 
please review I have handled the edge cases

examples/rust/discovery_service/discovery_service_client.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    let service = node
+        .service_builder(service_name().try_into()?)
+        .request_response::<(), StaticConfig>()
+        .enable_safe_overflow_for_responses(false)
+        .open_or_create()?;
+
+    let client = service
+        .client_builder()
+        // We guess that the samples are at most 16 bytes in size.
```
elfenpiff: I think this comment belongs to some code that was removed.

examples/rust/discovery_service/discovery_service_client.rs
```diff
@@ -0,0 +1,63 @@
...
+        // This is just a hint to the underlying allocator and is purely optional
+        // The better the guess is the less reallocations will be performed
+        // The underlying sample size will be increased with a power of two strategy
+        // when [`Client::loan_slice()`] or [`Client::loan_slice_uninit()`] requires more
+        // memory than available.
+        .create()?;
+
+    let request = client.loan_uninit()?;
+    let request = request.write_payload(());
+    let pending_response = request.send()?;
```
elfenpiff: When the payload is the unit type then `let pending_response = client.send_copy(())?;` make the code more readable - with this you could remove the two lines above.

examples/rust/discovery_service/discovery_service_client.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    // IMPORTANT: We need to wait for the request to be sent before we can receive responses.
```
elfenpiff: The request is sent in line 41. I think this line waits until `node.wait()` receives a signal like `SIGINT` or `SIGTERM` - when the user presses ctrl+c for instance.

examples/rust/discovery_service/discovery_service_client.rs
```diff
@@ -0,0 +1,63 @@
...
+    while !node.wait(CYCLE_TIME).is_ok() {}
+
+    while let Some(response) = pending_response.receive()? {
```
elfenpiff: I think this can be removed.

examples/rust/discovery_service/discovery_service_client.rs
```diff
@@ -0,0 +1,63 @@
...
+        println!("Service ID: {:?}", response.service_id().as_str());
+        println!("Service Name: {:?}", response.name().as_str());
+        println!();
+    }
+    // acquire all responses to our request from our buffer that were sent by the servers
+    while pending_response.is_connected() {
+        while let Some(response) = pending_response.receive()? {
```
elfenpiff: I think this loop has to be restructured a bit, otherwise the client will dismiss all received messages just because the server finished sending them.

So it should loop in a while true loop and break only when it no longer receives a response and the `pending_response` is no longer connected.

iceoryx2-cal/src/zero_copy_connection/common.rs
```diff
@@ -728,11 +728,14 @@ pub mod details {
                 AdaptiveWaitBuilder::new()
                     .create()
                     .unwrap()
-                    .wait_while(|| {
-                        self.storage.get().channels[channel_id.value()]
-                            .submission_queue
-                            .is_full()
-                    })
+                    .timed_wait_while(
```
elfenpiff: Please revert this. This breaks the contract of this function, it is called `blocking_send` and shall block without a timeout.

Why did you adjust this?

iceoryx2-services/discovery/src/service_discovery/service.rs
```diff
@@ -203,6 +253,9 @@ pub struct Config {
     /// The maximum number of samples the subscriber retains in its buffer.
     pub max_buffer_size: usize,
 
+    /// The maximum number of samples the subscriber retains in its response buffer.
```
elfenpiff: In this context we speak of the `Client` as the receiving port/endpoint and not the subscriber. Please always use the link notation when referring to other classes, so please write ``[`Client`]``

iceoryx2-services/discovery/src/service_discovery/service.rs
```diff
@@ -304,11 +359,25 @@ impl<S: ServiceType> Service<S> {
             tracker.sync(iceoryx_config)?;
         }
 
+        let server = Some(
```
elfenpiff: You should keep the service object here, otherwise the service is created with the server and then, if the client does not connect on time the service is cleaned up in that sense that it is no longer discoverable.

iceoryx2-services/discovery/tests/service_discovery_service_tests.rs
```diff
@@ -195,4 +200,191 @@ mod service_discovery_service {
...
+            .iter()
+            .map(|name| {
+                node.service_builder(name)
+                    .publish_subscribe::<u64>()
+                    .create()
+                    .unwrap()
+            })
+            .collect();
+
+        // === Spawn discovery service in a separate thread ===
```
orecham: What's your reasoning for a separate thread ? You have more control if you manage its execution in the test thread. You can explicitly decide when to progress the execution.

iceoryx2-services/discovery/src/service_discovery/service.rs
```diff
@@ -385,6 +453,43 @@ impl<S: ServiceType> Service<S> {
...
+
+                for service in services.iter() {
+                    if !self.discovery_config.include_internal
+                        && ServiceName::has_iox2_prefix(service.name())
+                    {
+                        continue;
+                    }
+                    let response = active_request.loan_uninit()?;
+                    let response = response.write_payload(service.clone());
+                    response.send()?;
```
elfenpiff: When the response buffer is full this call would not be able to deliver it. When now two separate clients request a discovery result and on client does not receive it on purpose, it deadlocks the other client and this service.

Since iceoryx2 is intended for mission-critical system we must always ensure that a failure/misbehavior in one process does not affect any other process. In this case a rogue client could deadlock this service and prevent that any other client receives any more updates.

This must be handled differently.

iceoryx2-services/discovery/tests/service_discovery_service_tests.rs
```diff
@@ -195,4 +200,191 @@ mod service_discovery_service {
...
+
+        let client = service.client_builder().create().unwrap();
+
+        let request = client.loan_uninit()?;
+        let request = request.write_payload(());
+        let pending_response = request.send()?;
+
+        let mut counter = 0;
+
+        // IMPORTANT: We need to wait for the request to be sent before we can receive responses.
```
orecham: Instead of waiting, you can progress the execution of the service by calling `spin()` here.

iceoryx2-services/discovery/tests/service_discovery_service_tests.rs
```diff
@@ -195,4 +200,191 @@ mod service_discovery_service {
...
+        while !node.wait(CYCLE_TIME).is_ok() {}
+
+        while let Some(response) = pending_response.receive()? {
+            assert_that!(service_names.contains(response.name()), eq true);
+            counter += 1;
+        }
+        // acquire all responses to our request from our buffer that were sent by the servers
+        while pending_response.is_connected() {
```
orecham: This could cause an infinite loop and stall CI.

elfenpiff (changes requested):
@RTK-jangid thank you for your contribution. The main issue I see with your pull request is that violates the freedom of interference paradigm which we have to follow strictly since iceoryx2 runs in mission critical systems.

In your case, since you disabled the overflow and set the `UnableToDeliveryStrategy` to `Block` the buffer in a buggy client could run full, this would deadlock the server and any other client would no longer receive messages from the server.

One solution could be to go for dynamic data and send the result in one package. The client buffer is set to 1 to save some memory. I assume that such a sample could be around ~250kb (1000 services and the service name has a size of 255 bytes)

I am start to ask myself if request response might be the wrong messaging pattern here since we might end up with a setup of 1 server and 1000 clients which would be a huge waste of memory (250kb * 1000 = 255mb reserved memory for the worst case to ensure freedom of interference from a memory perspective) - especially in the embedded world.
Maybe the upcoming blackboard pattern would be more fitting.

iceoryx2-services/discovery/tests/service_discovery_service_tests.rs
```diff
@@ -195,4 +200,191 @@ mod service_discovery_service {
...
+        let request = client.loan_uninit()?;
+        let request = request.write_payload(());
+        let pending_response = request.send()?;
+
+        let mut counter = 0;
+
+        // IMPORTANT: We need to wait for the request to be sent before we can receive responses.
+        while !node.wait(CYCLE_TIME).is_ok() {}
+
+        while let Some(response) = pending_response.receive()? {
```
orecham: How about something like this:

```rust
let mut sut = Service::<ipc::Service>::create(&discovery_config, &iceoryx_config_clone).unwrap();

...

let count_responses = || {
    let mut num = 0;
    while let Some(response) = pending_response.receive().unwrap() {
        assert_that!(service_names.contains(response.name()), eq true);
        num += 1;
    }
    num
};

let pending_response = request.send()?;

let mut counter = 0;

sut.spin();
let counter += count_responses();
assert_that!(counter, eq MAX_BUFFER_SIZE);
assert_that!(pending_response.is_connected(), true);

sut.spin();
let counter += count_responses();
assert_that!(counter, eq NUM_SERVICES);
assert_that!(pending_response.is_connected(), false);
```

Here I am assuming:

1. The server will send as many responses as it can until the buffer is full
1. It takes exactly two executions of `spin()` to send all responses
1. The `pending_response` is kept connected by the server while there are still more responses to send

If you are targeting different behaviour, I would still propose a similar approach to testing the bahaviour as it is more controlled.

iceoryx2-services/discovery/tests/service_discovery_service_tests.rs
```diff
@@ -195,4 +200,191 @@ mod service_discovery_service {
...
+
+        // Optional: if `sut.spin()` exits, join the thread
+        // Or detach it if it runs indefinitely (and your test exits after verification)
+        drop(handle);
+
+        Ok(())
+    }
+
+    #[test]
+    fn check_response_buffer_if_remains_full() -> Result<(), Box<dyn std::error::Error>> {
```
orecham: This test has similar issues.

iceoryx2-services/discovery/tests/service_discovery_service_tests.rs
```diff
@@ -195,4 +200,191 @@ mod service_discovery_service {
...
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        let iceoryx_config = generate_isolated_config();
+
+        // create some services in the main thread
+        let node = NodeBuilder::new()
+            .config(&iceoryx_config)
+            .create::<ipc::Service>()
+            .unwrap();
+
+        let service_names: Vec<_> = (0..6).map(|_| generate_name()).collect();
```
orecham: The magic numbers throughout this test should be named and re-used where appropriate. e.g.

```rust
const NUM_SERVICES: usize = 6;
const MAX_BUFFER_SIZE: usize = NUM_SERVICES / 2;
// ...etc

let service_names: Vec<_> = (0..NUM_SERVICES).map(|_| generate_name()).collect();
// ..etc.
```

iceoryx2-cal/src/zero_copy_connection/common.rs
orecham: This change broke some unit tests. Please revert.

It should be possible to finish this PR without changing the core. Changes to the core need to be done with care and thorough testing so it's better to avoid this unless absolutely deemed necessary.

orecham:
> @RTK-jangid thank you for your contribution. The main issue I see with your pull request is that violates the freedom of interference paradigm which we have to follow strictly since iceoryx2 runs in mission critical systems.
> 
> In your case, since you disabled the overflow and set the `UnableToDeliveryStrategy` to `Block` the buffer in a buggy client could run full, this would deadlock the server and any other client would no longer receive messages from the server.
> 
> One solution could be to go for dynamic data and send the result in one package. The client buffer is set to 1 to save some memory. I assume that such a sample could be around ~250kb (1000 services and the service name has a size of 255 bytes)
> 
> I am start to ask myself if request response might be the wrong messaging pattern here since we might end up with a setup of 1 server and 1000 clients which would be a huge waste of memory (250kb * 1000 = 255mb reserved memory for the worst case to ensure freedom of interference from a memory perspective) - especially in the embedded world. Maybe the upcoming blackboard pattern would be more fitting.

@elfenpiff For context, using a static buffer was an idea from me. See this: https://github.com/eclipse-iceoryx/iceoryx2/pull/805#discussion_r2132442320

@RTK-jangid Let's revert to your original solution using the dynamic data. 

The comments related to managing the execution of the service in the test are still valid, though.

RTK-jangid:
@orecham @elfenpiff 
Okay I would try with dynamic data but here also this parameter `.initial_max_slice_length(10)` needs to be configurable.
For e.g
if the value of `initial_max_slice_length = 10` the response will only send 10 services in single result.

So I wil try to explore more on Slice Type Api for now!!

Thank you for your Precious time!

elfenpiff:
@RTK-jangid 
> Okay I would try with dynamic data but here also this parameter `.initial_max_slice_length(10)` needs to be configurable. For e.g if the value of `initial_max_slice_length = 10` the response will only send 10 services in single result.

See: https://github.com/eclipse-iceoryx/iceoryx2/blob/main/examples/rust/request_response_dynamic_data/server.rs#L36
You can combine it with an allocation strategy and then the server reallocates memory if you require more then 10

> Thank you for your Precious time!

You are welcome! Always happy to help.

RTK-jangid:
Hi @orecham  @elfenpiff ,

I have implemented sending data dynamically. Please review! and let me know if there are some changes.

FYI: I also have written a unit test for the server. I have checked the server with 1000 random services and getting those services from discovery to client took **3.82 seconds**

examples/rust/discovery_service/discovery_service_client.rs
orecham: @elfenpiff Can you elaborate?

iceoryx2-services/discovery/src/service_discovery/service.rs
orecham: @RTK-jangid Will you do this?

iceoryx2-services/discovery/tests/service_discovery_service_tests.rs
```diff
@@ -196,4 +197,67 @@ mod service_discovery_service {
             test_fail!("expected DiscoveryEvent::Added for the internal service")
         }
     }
+
+    #[test]
+    fn get_current_discovery_states() -> Result<(), Box<dyn core::error::Error>> {
```
orecham, line 202: Nice clean test :)

iceoryx2-services/discovery/src/service_discovery/service.rs
```diff
@@ -304,12 +355,23 @@ impl<S: ServiceType> Service<S> {
         if discovery_config.sync_on_initialization {
             tracker.sync(iceoryx_config)?;
         }
+        let server = Some(
+            node.service_builder(service_name())
+                .request_response::<(), [StaticConfig]>()
+                .open_or_create()
+                .map_err(|_| CreationError::ServiceCreationFailure)?
+                .server_builder()
```
orecham:
```suggestion
                .server_builder()
                .initial_max_slice_length(discovery_config.initial_max_response_size) // this is a guess for the maximum "number of `StaticConfig` elements to expect, you can choose another name for the config if you like
```

By exposing the `initial_max_slice_length` a user can provide an initial guess for the maximum response size. If the guess is exceeded, more memory is allocated - this takes time, but with a good initial guess should happen less often and the performance should be better.

iceoryx2-services/discovery/tests/service_discovery_service_tests.rs
```diff
@@ -196,4 +197,67 @@ mod service_discovery_service {
...
+            .iter()
+            .map(|name| {
+                node.service_builder(name)
+                    .publish_subscribe::<u64>()
+                    .create()
+                    .unwrap()
+            })
+            .collect();
+
+        let discovery_config = Config {
```
orecham, line 225: If you add the config `initial_max_response_size` you could set it to `NUM_SERVICES` so allocation only occurs once (when the server is created).

orecham:
> I have checked the server with 1000 random services and getting those services from discovery to client took **3.82 seconds**

@RTK-jangid 

I would guess a number of allocations are occurring until the memory segment is large enough for 1000 services since the service is not told up-front to expect up to 1000 services.

examples/rust/discovery_service/discovery_service_client.rs
orecham: @RTK-jangid Will you do this?

orecham (commented):
Looking good. A few findings remain to to address / respond to.

orecham:
@RTK-jangid Rebasing onto the latest main should fix the CI failure on `static-code-analysis-cpp` - it's not stemming from this PR.

RTK-jangid:
@orecham I have pushed the changes requested.

RTK-jangid:
Hey @orecham 
Is there some issue or work I can take up while this PR is getting reviewed?

iceoryx2-services/discovery/src/service_discovery/service.rs
orecham: @RTK-jangid Did you miss this one or are you choosing to not address it ? If the latter, could you explain your reasoning?

In case there is some confusion, this finding is suggesting to keep the **service** object alive in addition to the **server** object.

examples/rust/discovery_service/discovery_service_client.rs
```diff
@@ -0,0 +1,46 @@
...
+        .open_or_create()?;
+
+    let client = service.client_builder().create()?;
+
+    let pending_response = client.send_copy(())?;
+
+    // IMPORTANT: We need to wait for the request to be sent before we can receive responses.
+    while !node.wait(CYCLE_TIME).is_ok() {}
+
+    while let Some(response) = pending_response.receive()? {
```
orecham: Could you use a [`Waitset`](https://github.com/eclipse-iceoryx/iceoryx2/blob/main/iceoryx2/src/waitset.rs#L109) here instead?

Set up a periodic timer to wake up and check for a response. When the response is received, exit cleanly.

Using the waitset will have the application gracefully handle other signals too, like `SIGINT`.

iceoryx2-services/discovery/src/service_discovery/service.rs
```diff
@@ -386,6 +463,39 @@ impl<S: ServiceType> Service<S> {
...
+    ///
+    /// # Errors
+    ///
+    /// Returns a `ServerSpinError` if there was an error in Responsding,
+    /// Loaning or recieving Requests.
+    pub fn handle_discovery_requests(&mut self) -> Result<(), ServerSpinError> {
+        if let Some(server) = &mut self.server {
+            while let Some(active_request) = server.receive()? {
+                // Handle the request
+                let mut service_details = self.tracker.get_all();
```
orecham, line 497: The service details are being copied multiple times here. Once out of the tracker to this intermediate container, and then once into the shared memory.

Can you see a way to reduce this to one copy?

iceoryx2-services/discovery/src/service_discovery/service.rs
```diff
@@ -214,6 +264,12 @@ pub struct Config {
 
     /// The maximum number of listeners to the service permitted.
     pub max_listeners: usize,
+
+    /// Whether to enable the server for handling requests.
+    pub enable_server: bool,
```
orecham, line 271: Not a bad idea. What was your reasoning, though?

orecham:
> Hey @orecham Is there some issue or work I can take up while this PR is getting reviewed?

@RTK-jangid 

Would this be interesting to you? https://github.com/eclipse-iceoryx/iceoryx2/issues/832

RTK-jangid:
> > Hey @orecham Is there some issue or work I can take up while this PR is getting reviewed?
> 
> @RTK-jangid
> 
> Would this be interesting to you? #832


Thank for the suggesting I think I will be able to do it. So I will take a look into this issue :)

RTK-jangid:
@orecham 
I have pushed the changes

I have added a comment above on the doubt for the service object for `request_response`. Please clarify

orecham:
> I have added a comment above on the doubt for the service object for `request_response`. Please clarify

@RTK-jangid I can't find your comment for some reason (maybe I am blind?)

Could you add it to this thread: https://github.com/eclipse-iceoryx/iceoryx2/pull/805#discussion_r2161336194

iceoryx2-services/discovery/src/service_discovery/service.rs
orecham: @RTK-jangid Forgot to ping you here. Could you share your reasoning (just so I am on the same page).

RTK-jangid:
@orecham 
for this [#805 (comment)](https://github.com/eclipse-iceoryx/iceoryx2/pull/805#discussion_r2161350815).
I had commented this:

I thought By giving this option we can save some bytes acquired by the server which does some memory initialisation on creation.
If someone doesn't wants to use the discovery server he can save some memory

I had added the above comments to the thread.
Are my comments not visible or this reasoning was not enough?

RTK-jangid:
@orecham 
For this comment [#805 (comment)](https://github.com/eclipse-iceoryx/iceoryx2/pull/805#discussion_r2161336194) I had commented this:

@orecham I think I am getting confused here. Right now the code is this:

```
        let mut server = None;
        if discovery_config.enable_server {
            let request_response = node
                .service_builder(service_name())
                .request_response::<(), [StaticConfig]>()
                .open_or_create()
                .map_err(|_| CreationError::ServiceCreationFailure)?;

            server = Some(
                request_response
                    .server_builder()
                    .initial_max_slice_len(discovery_config.initial_max_slice_len)
                    .allocation_strategy(AllocationStrategy::PowerOfTwo)
                    .create()
                    .map_err(|_| CreationError::ServiceCreationFailure)?,
            );
        }

```

Now should I add resquest_response object in the Service like this?

```
pub struct Service<S: ServiceType> {
    discovery_config: Config,
    iceoryx_config: IceoryxConfig,
    _node: Node<S>,
    publisher: Option<Publisher<S, Payload, ()>>,
    resquest_response<PortFactory<S, (), (), [StaticConfig], ()>>
    server: Option<Server<S, (), (), [StaticConfig], ()>>,
    notifier: Option<Notifier<S>>,
    tracker: Tracker<S>,
}
```
If I haven't understood it right then please correct me on this. And also Where will we use service object again? please elaborate

This is my first ever open source PR so it will take some time to get hang of the standards and basic understandings.

orecham:
@RTK-jangid 

> This is my first ever open source PR so it will take some time to get hang of the standards and basic understandings

No problem ! 😄 

> If I haven't understood it right then please correct me on this. And also Where will we use service object again? please elaborate

This code:

```rust
let request_response = node
    .service_builder(service_name())
    .request_response::<(), [StaticConfig]>()
    .open_or_create()
    .map_err(|_| CreationError::ServiceCreationFailure)?;
```

is creating a "service" in the `iceoryx2` communication infrastructure. The value it returns is a struct that implements the [`Service`](https://github.com/eclipse-iceoryx/iceoryx2/blob/main/iceoryx2/src/service/mod.rs#L654) trait. Here are some relevant docs: https://docs.rs/iceoryx2/0.6.1/iceoryx2/service/index.html

Services are used by `iceoryx2` to establish communication between applications. 
The application we are working on in this PR, the "discovery service", uses this "service" infrastructure to provide discovery information. 

The structs returned when opening or creating a service are used to manage the lifetime of the "service" in `iceoryx2`:

1. As long as one application holds this struct for a given service, `iceoryx2` ensures that the service is discoverable and can be used to establish communication between applications
2. When no applications hold this service object, `iceoryx2` can choose to clean up the service thus making it not discoverable to other applications

This is what this comment was talking about: https://github.com/eclipse-iceoryx/iceoryx2/pull/805#discussion_r2140930137

> Now should I add resquest_response object in the Service like this?

Yes, the returned object should be stored to prevent it being cleaned up.

---

Let me know if you have more questions.

FYI: We have some more documentation coming with the next release to make these concepts more visible.

RTK-jangid:
@orecham  
I have the change to push service object of server to be stored.
Please check!

examples/rust/discovery_service/discovery_service_client.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    let service = node
+        .service_builder(service_name().try_into()?)
+        .request_response::<(), [StaticConfig]>()
+        .open_or_create()?;
+
+    let client = service.client_builder().create()?;
+
+    let waitset = WaitSetBuilder::new().create::<ipc::Service>()?;
+    let guard_1 = waitset.attach_interval(CYCLE_TIME)?;
```
orecham:
```suggestion
    let guard = waitset.attach_interval(CYCLE_TIME)?;
```

orecham (commented):
Just one style nitpick and then we are good to merge :)

RTK-jangid:
@orecham @elfenpiff  
Thank you for your time and support I learned a lot. 
Looking forward to contribute more :)

elfenpiff:
> Thank you for your time and support I learned a lot.
> Looking forward to contribute more :)

@RTK-jangid thank you for your contribution and for pushing the PR through - especially when it is a bit more cumbersome when it comes to mission-critical software! We would be happy to receive more contributions from your side.

PR #844, opened 2025-06-28
`iox2 config explain` command for config descriptions

### LINKED ISSUE #832: Extend 'iox2-config' to describe the config parameter
## Refactoring

The `config/README.md` has a detailed description of the config parameter. It would be nice to get the same info with `iox2 config explain`.

## Tasks

* [ ] create a `iox2 config explain` command which output the same content as `config/README.md`
* [ ] add a CI check to enforce that the output of `iox2 config explain` is in sync with `config/README.md`

RTK-jangid:
## Notes for Reviewer

- Adds a CLI command `iox2 config explain` to display all config parameter descriptions.
- Also includes a test to ensure each config has a corresponding description.

* [ ] Assign PR to reviewer

## References

Closes #832

### DIFF
--- a/iceoryx2-cli/iox2-config/src/commands.rs
+++ b/iceoryx2-cli/iox2-config/src/commands.rs
@@ -23,6 +23,8 @@ use iceoryx2_bb_system_types::file_path::FilePath;
 use iceoryx2_bb_system_types::path::Path;
 use std::panic::catch_unwind;
 
+use iceoryx2_cli::config_descriptions::get_sections;
+
 /// Prints the whole system configuration with all limits, features and details to the console.
 pub fn print_system_configuration() {
     println!(
@@ -198,5 +200,22 @@ fn generate(config_dir: Path, filepath: FilePath) -> Result<()> {
 
     println!("Default configuration is generated at {}", filepath);
 
+    Ok(())
+}
+
+pub fn print_config_description() -> Result<()> {
+    // Get a vector here that consists of all the configuration parameters and descriptions
+    let sections = get_sections();
+    for section in sections {
+        println!("\n== {} ==\n", section.name.bright_green());
+        for entry in section.entries {
+            println!(
+                "{:<50} {:<15} {}",
+                entry.key.bright_blue(),
+                entry.value_type.bright_yellow(),
+                entry.description.italic()
+            );
+        }
+    }
     Ok(())
 }
--- a/iceoryx2-cli/lib/src/config_descriptions.rs
+++ b/iceoryx2-cli/lib/src/config_descriptions.rs
@@ -1,0 +1,307 @@
+// Copyright (c) 2025 Contributors to the Eclipse Foundation
+//
+// See the NOTICE file(s) distributed with this work for additional
+// information regarding copyright ownership.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Apache Software License 2.0 which is available at
+// https://www.apache.org/licenses/LICENSE-2.0, or the MIT license
+// which is available at https://opensource.org/licenses/MIT.
+//
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+pub struct CliEntry {
+    pub key: &'static str,
+    pub value_type: &'static str,
+    pub description: &'static str,
+}
+
+pub struct CliSection {
+    pub name: &'static str,
+    pub entries: Vec<CliEntry>,
+}
+
+pub fn get_sections() -> Vec<CliSection> {
+    vec![
+        CliSection {
+            name: "Global",
+            entries: vec![
+                CliEntry {
+                    key: "global.root-path-unix",
+                    value_type: "[string]",
+                    description: "Path used as root for IPC-related files on Unix-based systems.",
+                },
+                CliEntry {
+                    key: "global.root-path-windows",
+                    value_type: "[string]",
+                    description: "Path used as root for IPC-related files on Windows systems.",
+                },
+                CliEntry {
+                    key: "global.prefix",
+                    value_type: "[string]",
+                    description: "Prefix prepended to all internal IPC resource names.",
+                },
+            ],
+        },
+        CliSection {
+            name: "Global: Service",
+            entries: vec![
+                CliEntry {
+                    key: "global.service.directory",
+                    value_type: "[string]",
+                    description: "Specifies the path for service-related files under `global.root-path`.",
+                },
+                CliEntry {
+                    key: "global.service.data-segment-suffix",
+                    value_type: "[string]",
+                    description: "Suffix added to the ports's data segment.",
+                },
+                CliEntry {
+                    key: "global.service.static-config-storage-suffix",
+                    value_type: "[string]",
+                    description: "Suffix for static service configuration files.",
+                },
+                CliEntry {
+                    key: "global.service.dynamic-config-storage-suffix",
+                    value_type: "[string]",
+                    description: "Suffix for dynamic service configuration files.",
+                },
+                CliEntry {
+                    key: "global.service.event-connection-suffix",
+                    value_type: "[string]",
+                    description: "Suffix for event channel.",
+                },
+                CliEntry {
+                    key: "global.service.connection-suffix",
+                    value_type: "[string]",
+                    description: "Suffix for one-to-one connections.",
+                },
+            ],
+        },
+                CliSection {
+            name: "Global: Node",
+            entries: vec![
+                CliEntry {
+                    key: "global.node.directory",
+                    value_type: "[string]",
+                    description: "Specifies the path for node-related files under `global.root-path`.",
+                },
+                CliEntry {
+                    key: "global.node.monitor-suffix",
+                    value_type: "[string]",
+                    description: "Suffix added to the node monitor.",
+                },
+                CliEntry {
+                    key: "global.node.static-config-suffix",
+                    value_type: "[string]",
+                    description: "Suffix added to the static config of the node.",
+                },
+                CliEntry {
+                    key: "global.node.service-tag-suffix",
+                    value_type: "[string]",
+                    description: "Suffix added to the service tag of the node.",
+                },
+                CliEntry {
+                    key: "global.node.cleanup-dead-nodes-on-creation",
+                    value_type: "[`true`|`false`]",
+                    description: "Defines if there shall be a scan for dead nodes with a following stale resource cleanup whenever a new node is created.",
+                },
+                CliEntry {
+                    key: "global.node.cleanup-dead-nodes-on-destruction",
+                    value_type: "[`true`|`false`]",
+                    description: "Defines if there shall be a scan for dead nodes with a following stale resource cleanup whenever a node is going out-of-scope.",
+                },
+            ],
+        },
+        CliSection{
+            name: "Global: Service Creation Timeout",
+            entries: vec![
+                CliEntry {
+                    key: "global.service.creation-timeout.secs",
+                    value_type: "[int]",
+                    description: "Maximum time for service setup in seconds.",
+                },
+                CliEntry {
+                    key: "global.service.creation-timeout.nanos",
+                    value_type: "[int]",
+                    description: "Additional nanoseconds for service setup timeout.",
+                },
+            ],
+        },
+        CliSection {
+            name: "Defaults: Publish Subscribe Messaging Pattern",
+            entries: vec![
+                CliEntry {
+                    key: "defaults.publish-subscribe.max-subscribers",
+                    value_type: "[int]",
+                    description: "Maximum number of subscribers.",
+                },
+                CliEntry {
+                    key: "defaults.publish-subscribe.max-publishers",
+                    value_type: "[int]",
+                    description: "Maximum number of publishers.",
+                },
+                CliEntry {
+                    key: "defaults.publish-subscribe.max-nodes",
+                    value_type: "[int]",
+                    description: "Maximum number of nodes.",
+                },
+                CliEntry {
+                    key: "defaults.publish-subscribe.subscriber-max-buffer-size",
+                    value_type: "[int]",
+                    description: "Maximum buffer size of a subscriber.",
+                },
+                CliEntry {
+                    key: "defaults.publish-subscribe.subscriber-max-borrowed-samples",
+                    value_type: "[int]",
+                    description: "Maximum samples a subscriber can hold.",
+                },
+                CliEntry {
+                    key: "defaults.publish-subscribe.publisher-max-loaned-samples",
+                    value_type: "[int]",
+                    description: "Maximum samples a publisher can loan.",
+                },
+                CliEntry {
+                    key: "defaults.publish-subscribe.publisher-history-size",
+                    value_type: "[int]",
+                    description: "Maximum history size a subscriber can request.",
+                },
+                CliEntry {
+                    key: "defaults.publish-subscribe.enable-safe-overflow",
+                    value_type: "[`true`|`false`]",
+                    description: "Default overflow behavior.",
+                },
+                CliEntry {
+                    key: "defaults.publish-subscribe.unable-to-deliver-strategy",
+                    value_type: "[`Block`|`DiscardSample`]",
+                    description: "Default strategy for non-overflowing setups when delivery fails.",
+                },
+                CliEntry {
+                    key: "defaults.publish-subscribe.subscriber-expired-connection-buffer",
+                    value_type: "[int]",
+                    description: "Expired connection buffer size of the subscriber. Connections to publishers are expired when the publisher disconnected from the service and the connection contains unconsumed samples.",
+                },
+            ],
+        },
+        CliSection {
+            name: "Defaults: Event Messaging Pattern",
+            entries: vec![
+                CliEntry {
+                    key: "defaults.event.max-listeners",
+                    value_type: "[int]",
+                    description: "Maximum number of listeners.",
+                },
+                CliEntry {
+                    key: "defaults.event.max-notifiers",
+                    value_type: "[int]",
+                    description: "Maximum number of notifiers.",
+                },
+                CliEntry {
+                    key: "defaults.event.max-nodes",
+                    value_type: "[int]",
+                    description: "Maximum number of nodes.",
+                },
+                CliEntry {
+                    key: "defaults.event.event-id-max-value",
+                    value_type: "[int]",
+                    description: "Greatest value an [`EventId`] can have.",
+                },
+            ],
+        },
+        CliSection {
+            name: "Defaults: Request Response Messaging Pattern",
+            entries: vec![
+                CliEntry {
+                    key: "defaults.request-response.enable-safe-overflow-for-requests",
+                    value_type: "[`true`|`false`]",
+                    description: "Defines if the request buffer of the service safely overflows.",
+                },
+                CliEntry {
+                    key: "defaults.request-response.enable-safe-overflow-for-responses",
+                    value_type: "[`true`|`false`]",
+                    description: "Defines if the request buffer of the service safely overflows.",
+                },
+                CliEntry {
+                    key: "defaults.request-response.max-active-requests-per-client",
+                    value_type: "[int]",
+                    description: "The maximum of active requests a server can hold per client.",
+                },
+                CliEntry {
+                    key: "defaults.request-response.max-response-buffer-size",
+                    value_type: "[int]",
+                    description: "The maximum buffer size for responses for an active request.",
+                },
+                CliEntry {
+                    key: "defaults.request-response.max-servers",
+                    value_type: "[int]",
+                    description: "The maximum amount of supported servers.",
+                },
+                CliEntry {
+                    key: "defaults.request-response.max-clients",
+                    value_type: "[int]",
+                    description: "The maximum amount of supported clients.",
+                },
+                CliEntry {
+                    key: "defaults.request-response.max-nodes",
+                    value_type: "[int]",
+                    description: "The maximum amount of supported nodes. Defines indirectly how many processes can open the service at the same time.",
+                },
+                CliEntry {
+                    key: "defaults.request-response.max-borrowed-responses-per-pending-response",
+                    value_type: "[int]",
+                    description: "The maximum number of borrowed responses a client can hold in parallel per pending response.",
+                },
+                CliEntry {
+                    key: "defaults.request-response.max-loaned-requests",
+                    value_type: "[int]",
+                    description: "Maximum number of requests a client can loan.",
+                },
+                CliEntry {
+                    key: "defaults.request-response.server-max-loaned-responses-per-request",
+                    value_type: "[int]",
+                    description: "Maximum number of responses a server can loan per request.",
+                },
+                CliEntry {
+                    key: "defaults.request-response.client-unable-to-deliver-strategy",
+                    value_type: "[`Block`|`DiscardSample`]",
+                    description: "Default strategy for non-overflowing setups when delivery fails.",
+                },
+                CliEntry {
+                    key: "defaults.request-response.server-unable-to-deliver-strategy",
+                    value_type: "[`Block`|`DiscardSample`]",
+                    description: "Default strategy for non-overflowing setups when delivery fails.",
+                },
+                CliEntry {
+                    key: "defaults.request-response.client-expired-connection-buffer",
+                    value_type: "[int]",
+                    description: "Expired connection buffer size of the client. Connections to servers are expired when the server disconnected from the service and the connection contains unconsumed responses.",
+                },
+                CliEntry {
+                    key: "defaults.request-response.enable-fire-and-forget-requests",
+                    value_type: "[`true`|`false`]",
+                    description: "Enables the client to send requests without expecting a response.",
+                },
+                CliEntry {
+                    key: "defaults.request-response.server-expired-connection-buffer",
+                    value_type: "[int]",
+                    description: "Expired connection buffer size of the server. Connections to clients are expired when the client disconnected from the service and the connection contains unconsumed active requests.",
+                },
+            ],
+        },
+        CliSection {
+            name: "Defaults: Blackboard",
+            entries: vec![
+                CliEntry {
+                    key: "defaults.blackboard.max-readers",
+                    value_type: "[int]",
+                    description: "Maximum number of readers on the blackboard.",
+                },
+                CliEntry {
+                    key: "defaults.blackboard.max-nodes",
+                    value_type: "[int]",
+                    description: "Maximum number of nodes that can access the blackboard.",
+                },
+            ],
+        },
+    ]
+}
--- a/iceoryx2-cli/tests/config_description_tests.rs
+++ b/iceoryx2-cli/tests/config_description_tests.rs
@@ -1,0 +1,67 @@
+// Copyright (c) 2025 Contributors to the Eclipse Foundation
+//
+// See the NOTICE file(s) distributed with this work for additional
+// information regarding copyright ownership.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Apache Software License 2.0 which is available at
+// https://www.apache.org/licenses/LICENSE-2.0, or the MIT license
+// which is available at https://opensource.org/licenses/MIT.
+//
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+mod config_descriptions_tests {
+    use std::collections::HashSet;
+    use toml::Value;
+
+    use iceoryx2::config::Config;
+    use iceoryx2_bb_testing::assert_that;
+    use iceoryx2_cli::config_descriptions::get_sections;
+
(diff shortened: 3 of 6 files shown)


RTK-jangid:
Hey @elBoberido 
I have created the draft PR with `iox2 config explain`. Now I just want to have a CI between README.md and
`iox2 config explain`. So Do we have have some docs how we are building github actions/workflows.

elBoberido:
@RTK-jangid no, there are currently no docs how to add workflows. I think you can start by adding a small script to `internal/scripts`. Once you have something that works and can detect a discrepancy between the README.md and the output from the tool, I can guide you on the CI setup.

iceoryx2-cli/tests/config_description_tests.rs
```diff
@@ -0,0 +1,67 @@
...
+            _ => {
+                keys.insert(prefix);
+            }
+        }
+    }
+
+    #[test]
+    fn check_config_description_is_present_for_all_config_parameters() {
+        let config = Config::default();
+        let toml_string = toml::to_string(&config).expect("Failed to serialize config to TOML");
```
elBoberido: Maybe TOML is not the best format for this check. Type like `Option<Foo>` will not show up if the value is `None`. Maybe [RON](https://crates.io/crates/ron) is an alternative

RTK-jangid:
@elBoberido I have the pushed the updates please review. 

Apologies for late updates. University exams :sob:

elBoberido:
@RTK-jangid no problem. We are also quite busy. I hope you performed good at your exams 🙂

elfenpiff:
@RTK-jangid I finally found the time to try out `iox2 config explain` and I love it - nothing beats a nice colorful terminal output.
But I would recommend to structure it a bit differently since most of the time the terminal width is not wide enough and then the nice layout gets broken. What do you think about something like this here:

```
some.config.entry [TYPE]
explanation of some.config.entry

another.config.entry [ANOTHER_TYPE] 
some more config entry explanation
```

But please keep the colors as they are - they are great!

iceoryx2-cli/lib/src/config_descriptions.rs
```diff
@@ -0,0 +1,327 @@
...
+                    description: "Greatest value an [`EventId`] can have.",
+                },
+                CliEntry {
+                    key: "defaults.event.deadline",
+                    value_type: "[int]",
+                    description: "Maximum allowed time between two consecutive notifications. If not sent after this time, all listeners attached to a WaitSet will be notified.",
+                },
+                CliEntry {
+                    key: "defaults.event.notifier-created-event",
+                    value_type: "[Option<int>]",
```
elBoberido: Food for thought. As a user unfamiliar with the project I would be wondering how the value would look like. Just an `123` or maybe `Some(123)` or something else?

What do you think about adding an additional field like
```rs
key: ...
required: false // alternatively: 'optional: true'
value_type: "int"
```

iceoryx2-cli/tests/config_description_tests.rs
```diff
@@ -0,0 +1,72 @@
...
+            _ => {
+                keys.insert(prefix);
+            }
+        }
+    }
+
+    #[test]
+    fn check_config_description_is_present_for_all_config_parameters() {
+        // Replace this with a version of Config that populates all Option<T> fields
+        let config = Config::default();
```
elBoberido: Does this mean that `RON` also does not show `Option<T>` with a value of `None`?

internal/scripts/ci_check_description_vector_and_README.sh
```diff
@@ -0,0 +1,35 @@
...
+missing=0
+
+echo "🔍 Checking if each config key from Rust exists in README.md ..."
+
+while IFS= read -r key; do
+    if grep -qF "$key" "$README_FILE"; then
+        echo "✅ Found: $key"
+    else
+        echo "❌ Missing: $key"
+        ((missing++))
```
elBoberido: I have to check how this looks on the terminal. In general, we are using only colors, just in case a user does not have a font with emojis or the emojis look differently in the font the user have chosen.

@elfenpiff what is your stance on this. To emoji or not to emoji?

elBoberido (commented):
Locks good in general

RTK-jangid:
> @RTK-jangid I finally found the time to try out `iox2 config explain` and I love it - nothing beats a nice colorful terminal output. But I would recommend to structure it a bit differently since most of the time the terminal width is not wide enough and then the nice layout gets broken. What do you think about something like this here:
> 
> ```
> some.config.entry [TYPE]
> explanation of some.config.entry
> 
> another.config.entry [ANOTHER_TYPE] 
> some more config entry explanation
> ```
> 
> But please keep the colors as they are - they are great!

@elfenpiff Yes, I thank you for the idea I was thinking the same about how to solve for changing window size.  :+1:

internal/scripts/ci_check_description_vector_and_README.sh
elfenpiff: Some of our users are on some very old systems like Ubuntu 18.04 - I have no idea if this already supports emojis but I would tend to beautiful pure ascii arts with colors and without emojis.

RTK-jangid:
I have the pushed changes. Please review

@elBoberido @elfenpiff  this is the new output style for `iox2 config explain`

Screenshot from 2025-07-19 13-09-11

.github/workflows/build-test.yml
```diff
@@ -82,6 +82,14 @@ jobs:
           GITHUB_BEFORE: ${{ github.event.before }}
           GITHUB_AFTER: ${{ github.event.after }}
         run: ./internal/scripts/ci_check_markdown_linting.sh
+      
+      - name: Check config key documentation
+        env:
+          GH_TOKEN: ${{ github.token }}
```
elBoberido: Can you please remove the entire `env` section from here. It seems it is not needed by your script.

Can you also remove the `GH_TOKEN` from line 80. It is not required and since the job does not have set any permissions, it is also useless but nevertheless, should be removed.

.github/workflows/build-test.yml
```diff
@@ -82,6 +82,14 @@ jobs:
...
+          GITHUB_PR_NUMBER: ${{ github.event.pull_request.number }}
+          GITHUB_BEFORE: ${{ github.event.before }}
+          GITHUB_AFTER: ${{ github.event.after }}
+        run: chmod +x ./internal/scripts/ci_check_description_vector_and_README.sh && ./internal/scripts/ci_check_description_vector_and_README.sh
```
elBoberido:
```suggestion
        run: ./internal/scripts/ci_check_description_vector_and_README.sh
```

This can also be done locally and committed to the repo
```
chmod +x ./internal/scripts/ci_check_description_vector_and_README.sh
```

internal/scripts/ci_check_description_vector_and_README.sh
```diff
@@ -0,0 +1,49 @@
...
+echo -e "${CYAN}${BOLD}=== Checking if each config key from Rust exists in README.md ===${RESET}"
+
+while IFS= read -r key; do
+    if grep -qF "$key" "$README_FILE"; then
+        echo -e "${GREEN}[ OK ]${RESET} Found: $key"
+    else
+        echo -e "${RED}[FAIL]${RESET} Missing: $key"
+        ((missing++))
+    fi
+done < <(grep -oP 'key:\s*"\K[^"]+' "$RUST_FILE")
```
elBoberido, line 36: This is beyond my shell scripting expertise. @elfenpiff can you check this :sweat_smile:

iceoryx2-cli/iox2-config/src/commands.rs
```diff
@@ -196,3 +198,22 @@ fn generate(config_dir: Path, filepath: FilePath) -> Result<()> {
...
+
+pub fn print_config_description() -> Result<()> {
+    let sections = get_sections();
+
+    for section in sections {
+        println!("\n{}", format!("== {} ==", section.name).bright_green());
+
+        for entry in section.entries {
+            println!(
+                "{} [{}]",
```
elBoberido: This results in having two brackets for the value type, e.g. `[[int]]` for `int`. This might confuse users to believe a slice (`[int]`) is the actual type. Maybe it would be a good idea to have the `value_type` in `config_description.rs` just as `int` instead of `[int]`?

iceoryx2-cli/lib/src/config_descriptions.rs
elBoberido: What do you think about this proposal?

iceoryx2-cli/tests/config_description_tests.rs
elBoberido: Can the comment be removed?

elBoberido:
> I have the pushed changes. Please review
> 
> @elBoberido @elfenpiff this is the new output style for `iox2 config explain`
> Screenshot from 2025-07-19 13-09-11

Looks greate in general. Just the double brackets are a bit confusing.

Just realizing. Would it be possible to print the default value of the key, e.g. `max-subscribers` would be `8` ... maybe in a greyisch color?

So it would be something like
```
defaults.publish-subscribe.max-subscribers [int] (required; default value is 8)
Maximum number...
...
defaults.event.notifier-created-event [int] (optional; defaults to false)
```

@elfenpiff ... oh no, writing this made me just realize that we need to add default values to all keys in the config file, else with each new entry in the config struct, we break the configuration of our users since they have to update their config with the new keys ... and I don't mean the `Default` trait but we might need to use `#[serde(default)]` and have a more fine grained implementation of the `Default` trait.

elfenpiff:
@elBoberido It is morning and I didn't have yet my full cup of Earl Grey Hot (with milk) but I am lost here:

> oh no, writing this made me just realize that we need to add default values to all keys in the config file, else with each new entry in the config struct, we break the configuration of our users since they have to update their config with the new keys ... and I don't mean the Default trait but we might need to use #[serde(default)] and have a more fine grained implementation of the Default trait.

RTK-jangid:
image


@elBoberido @elfenpiff Pushed the new changes with default values and the requested changes

elBoberido:
> @elBoberido It is morning and I didn't have yet my full cup of Earl Grey Hot (with milk) but I am lost here:
> 
> > oh no, writing this made me just realize that we need to add default values to all keys in the config file, else with each new entry in the config struct, we break the configuration of our users since they have to update their config with the new keys ... and I don't mean the Default trait but we might need to use #[serde(default)] and have a more fine grained implementation of the Default trait.

Step by step instructions:
1. Grab a cup
2. Put some hot water into the cup
3. Put some Earl Grey into the water
4. While waiting
    1. Checkout the v0.6.1 tag
    2. run `cargo run --bin iox2-config generate local`
    3. Checkout `main` and run `cargo run --example publish-subscribe-publisher`
    4. Notice the error message that the config file cannot be parsed ... there are two issues: renaming of `root-path-unix` to `root-path` and all the `blackboard` config parameter are missing -> every time we add new config parameter the user config will break -> solution: use `#[serde(default)]`
5. Drink tea

elBoberido:
> image
> 
> @elBoberido @elfenpiff Pushed the new changes with default values and the requested changes

I think having single `[...]` like in the first screenshot is fine, maybe even in a different color than the actual type, like in the second screenshot. It's just the `[[...]]` that might be confusing.

Regarding the `Required` part, depending on the issue with newly added config values, we might end up having each key optional with a default value. I also recall that the keys with `Option<T>` have a special semantic meaning, so the idea to just label them `Required` was probably also not my brightest :sweat_smile: 

Oh, can you place quotes around values which are strings like `".service_tag"`

RTK-jangid:
image

@elBoberido Output changes done. Please check

.github/workflows/build-test.yml
elBoberido: @RTK-jangid can you take care of this?

iceoryx2-cli/lib/src/config_descriptions.rs
```diff
@@ -0,0 +1,389 @@
...
+
+    vec![
+        CliSection {
+            name: "Global",
+            entries: vec![
+                CliEntry {
+                    key: "global.root-path",
+                    value_type: "string",
+                    default: format!("\"{}\"", config.global.root_path()),
+                    description: "Defines the path for all iceoryx2 files and directories.",
```
elBoberido, line 37: Do you have an idea how to check if this string is also in the README at the corresponding key?

iceoryx2-cli/lib/src/config_descriptions.rs
```diff
@@ -0,0 +1,389 @@
...
+                    key: "global.service.blackboard-data-suffix",
+                    value_type: "string",
+                    default: format!("\"{}\"", config.global.service.blackboard_data_suffix),
+                    description: "The suffix of the blackboard payload data segment.",
+                },
+            ],
+        },
+        CliSection {
+            name: "Global: Node",
+            entries: vec![
```
elBoberido: Can you please move this above `Global: Service`.

Semantically, the node will be created before the service and right now, the `Global: Node` is placed right in between `Global: Service` and `Global: Service Creation Timeout`.

iceoryx2-cli/lib/src/config_descriptions.rs
```diff
@@ -0,0 +1,389 @@
...
+                    value_type: "int",
+                    default: config.defaults.event.event_id_max_value.to_string(),
+                    description: "Greatest value an [`EventId`] can have.",
+                },
+                CliEntry {
+                    key: "defaults.event.deadline",
+                    value_type: "Option<Duration>",
+                    default: config.defaults.event.deadline.map_or("None".to_string(), |e| format!("{e:?}")),
+                    description: "Maximum allowed time between two consecutive notifications. If not sent after this time, all listeners attached to a WaitSet will be notified.",
+                },
```
elBoberido, line 263: @elfenpiff @RTK-jangid AFAIK, if this key is not set, then the default value will be `None`, essentially deactivating the feature. But the user is also not supposed to use this like `deadline = Some(Duration)` but just `deadline = Duration`. I think the user needs to be made aware of this but I'm not sure what the best option would be. Maybe having `value_type: "Duration"` and mentioning either in `default` and/or `description` that the `deadline` will not be used if the key is not present in the config file. The same would apply to the other three options below.

RTK-jangid:
image

@elBoberido  I have added a `**` mark in the description of these keys and during iox2 config explain command we will put up a note that tells that keys marked with `**` are optional configs and if not declared and assigned values then these configs will be deactivated.


> Do you have an idea how to check if this string is also in the README at the corresponding key?

I am not sure how to deal with this as bash script only compares the keys and the strings in README.md file to check if the keys are present . As there are different characters in README as `Option\<int\>` for value_type it becomes hard to match only string. And brain is not braining what else I could do. :dizzy_face:

iceoryx2-cli/lib/src/config_descriptions.rs
```diff
@@ -251,25 +251,25 @@ pub fn get_sections() -> Vec<CliSection> {
                     key: "defaults.event.deadline",
                     value_type: "Option<Duration>",
                     default: config.defaults.event.deadline.map_or("None".to_string(), |e| format!("{e:?}")),
-                    description: "Maximum allowed time between two consecutive notifications. If not sent after this time, all listeners attached to a WaitSet will be notified.",
+                    description: "** Maximum allowed time between two consecutive notifications. If not sent after this time, all listeners attached to a WaitSet will be notified.",
```
elBoberido: I still think that having the value type as `Option<T>` is confusing but I also don't have a good suggestion on how to improve it.

The default value of `None` is also a bit misleading, since is suggest to use `None` to deactivate the option and to use `Some(42)` to set a value.

Do you have an idea on how to make this more ambiguous to the user?

cc @elfenpiff

iceoryx2-cli/lib/src/config_descriptions.rs
```diff
@@ -0,0 +1,389 @@
...
+                    description: "Maximum number of nodes.",
+                },
+                CliEntry {
+                    key: "defaults.event.event-id-max-value",
+                    value_type: "int",
+                    default: config.defaults.event.event_id_max_value.to_string(),
+                    description: "Greatest value an [`EventId`] can have.",
+                },
+                CliEntry {
+                    key: "defaults.event.deadline",
```
elBoberido, line 253: I think this will be split up into `deadline.secs` and `deadline.nanos`. Similar to `creation-timeout`, which is also a `Duration`

iceoryx2-cli/lib/src/config_descriptions.rs
elfenpiff: @elBoberido I think the points you are mentioning are completely valid but should not be part of this PR. I would recommend creating a new issue. Add all insights from this PR there and close all comments related to this here.

I think we should get this PR merged with the scope it has and all side discussions should be happening somewhere else - otherwise either this PR explodes or everyone is confused since issues are raised that have nothing to do with `iox2 config explain`

iceoryx2-cli/lib/src/config_descriptions.rs
elfenpiff: @elBoberido lets create a new issue and discuss it there. For now I would remove the `**` and keep the documentation has it was. This PR shall just print the documentation as is with `iox2 config explain` and not improve the doc interface or solve documentation problems with the config itself.

iceoryx2-cli/lib/src/config_descriptions.rs
elBoberido: @RTK-jangid just wanted to be sure you are not waiting on input from our side?

If you are currently busy with other stuff, it is totally fine.

RTK-jangid:
> @elBoberido lets create a new issue and discuss it there. For now I would remove the `**` and keep the documentation has it was. This PR shall just print the documentation as is with `iox2 config explain` and not improve the doc interface or solve documentation problems with the config itself.

@elBoberido Apologies, So After this comment I was expecting a confirmation message and got confused what to do next and then I got engaged myself somewhere else. So Should I keep the bash file or remove it? 

So please provide me What to do with above mentioned issue and then I would update my PR with current main branch then we can merge it.

elBoberido:
@RTK-jangid oh, sorry. The idea was to not delay this PR because of that question and just merge it like it is, regarding that question.

I think the `deadline.secs` and `deadline.nanos` is the last remaining issue. Can you check if that is indeed the case that the `Duration` will be flattened like that.

RTK-jangid:
@elBoberido So I tried spliting `deadline.secs` and `deadline.nanos` but they are sub fields of Data type `Option<Duration>` and `deadline` is set to `None` by default So `Ron` it doesn't generate the sub fields like `deadline.secs` and `deadline.nanos` and the unit test case starts failing.

RTK-jangid:
Any ideas or workarounds ?

iceoryx2-cli/lib/src/config_descriptions.rs
```diff
@@ -0,0 +1,389 @@
...
+            ],
+        },
+        CliSection{
+            name: "Global: Service Creation Timeout",
+            entries: vec![
+                CliEntry {
+                    key: "global.service.creation-timeout.secs",
+                    value_type: "int",
+                    default: config.global.service.creation_timeout.as_secs().to_string(),
+                    description: "Maximum time for service setup in seconds. Uncreated services after this are marked as stalled.",
```
elBoberido: It seems this works only when `nanos` is also set and vice versa. This should be visible in the description. Could you please add something like `Attention: if 'secs' is set, 'nanos' must also be set`. Feel free to choose something else if you have a good idea.

elBoberido:
@RTK-jangid okay, I just checked what is required to have a valid `deadline` in the config file. It seems that both, `secs` and `nanos` must be present in order for serde to pick them up. If only one is defined, an error is emitted.

Now to the problem. The only thing I could come up was to change `default: String` to `default: Option<String>`. I'm not quite sure what other changes this would require but with that, one could do a special handling if the the value of `default` would be `None`, e.g. when printing the default value `default.unwrap_or("ignored if not set")`. This would then mean that the entry needs to be split into a `secs` and `nanos` part and that `value_type` would need to become just `int`, similar to `global.service.creation-timeout`. Does this make sense? Please also check if this would result in big changes. If this would be the case, we might find a different solution.

Btw, while writing this, I noticed that my IDE showed `default` in bold. I guess this is due to the `Default` trait. What do you think about naming the field `default_value`?

elBoberido:
@RTK-jangid I had a closer look and I think something like this might work

```rs
CliEntry {
    key: "defaults.event.deadline.secs",
    value_type: "secs",
    default: config.defaults.event.deadline.map(|d| d.as_secs().to_string()), // with 'subsec_nanos' for the nanos part
    description: "Maximum allowed time between two consecutive notifications. If not sent after this time, all listeners attached to a WaitSet will be notified."
};
``

**Comment by elBoberido** [comment]
... but I'm not quite sure if this would solve the problem with RON :sweat_smile:

**Comment by elBoberido** [comment]
... okay, I had a brief look at the check and my proposal would not fix the issue. I need to think a bit about the problem. Maybe RON is also not the right format here. I think if we would have a format which contains the type instead of the value, one could look for `Option<Duration>` and then apply a workaround.

@elfenpiff you already looked for a IDL format that can be generated with serde. Which crate did you find?

**Comment by RTK-jangid** [comment]
>Btw, while writing this, I noticed that my IDE showed default in bold. I guess this is due to the Default trait. What do you think about naming the field default_value?

Yes will do this.

**Comment by RTK-jangid** [comment]
> when printing the default value default.unwrap_or("ignored if not set"). This would then mean that the entry needs to be split into a secs and nanos part and that value_type would need to become just int, similar to global.service.creation-timeout

So the unit test checks if all the keys are present in the description or not, even if they are `None` but for to generate.

So in `creation-timeout` this field value is always `Duration`. So, RON picks it up and then splits the fields.

But `deadline` is `Option<Duration>` which can be set to None. So, RON can't determine the type and the subfields from the value as it can be `None`.

**Comment by RTK-jangid** [comment]
>I think if we would have a format which contains the type instead of the value, one could look for Option<Duration> and then apply a workaround.

Yes I am also searching for how to the get fields Data_type in the runtime then I can just put some dummy value according to data type. I want to make a a generic solution for `Option<T>` so that if`T` has sub fields then it gets automatically detected. The test case shouldn't need to be modified in future.

**Comment by RTK-jangid** [comment]
@elBoberido  If you agree, For Right Now I don't believe it is creating much confusion for user. We can merge this PR as it is and then I can open a new issue with this and then we can try to solve this problem.


>It seems this works only when nanos is also set and vice versa. This should be visible in the description. Could you please add something like Attention: if 'secs' is set, 'nanos' must also be set. Feel free to choose something else if you have a good idea.

After doing this obviously.

**Comment by elBoberido** [comment]
@RTK-jangid agreed. Can you also add to `defaults.event.deadline` that due to a current limitation, the keys are actually `defaults.event.deadline.secs` and `defaults.event.deadline.nanos`.

This might be interesting https://crates.io/crates/serde-reflection in order to get the actual type.

**Comment by RTK-jangid** [comment]
@elBoberido I have updated the descriptions and README.md with warnings and Note.

>This might be interesting https://crates.io/crates/serde-reflection in order to get the actual type.

I am looking into this crate. This also requires to add custom fields to registry manually. Using this also solution won't be very small so I will try to resolve all these issues in the next PR.

File: config/README.md [hunk 15]
```diff
@@ -85,9 +85,17 @@ Adjusting `global` settings ensures a non-interfering setup.
...
-* `global.service.creation-timeout.secs` &
-  `global.service.creation-timeout.nanos` - [int]: Maximum time for service
-  setup. Uncreated services after this are marked as stalled.
+* `global.service.creation-timeout.secs` - [int]: Maximum time for service setup
+  in seconds. Uncreated services after this are marked as stalled.
+  Attention: Both 'secs' and 'nanos' must be set together; leaving one
+  unset will cause the configuration to be invalid.
+* `global.service.creation-timeout.nanos` - [int]: Additional nanoseconds for
+  service setup timeout.Maximum time for service
+  setup.
```
elBoberido: It is sufficient to add this in the follow up PR, but can you also add the `Attention` to the `nanos` part?
I'm the prime example of a user who would only read the `nanos` part if I would need a timeout below 1 second.

config/README.md
```diff
@@ -98,6 +106,13 @@ Adjusting `global` settings ensures a non-interfering setup.
 * `defaults.event.max-nodes` - [int]: Maximum number of nodes.
 * `defaults.event.event-id-max-value` - [int]: Greatest value an [`EventId`] can
   have.
+* `defaults.event.deadline` - [Option\<Duration\>]: Maximum allowed time between
+  two consecutive notifications. If not sent after this time, all listeners attached
+  to a WaitSet will be notified.
+  Due to a current limitation, the keys are actually
+  `defaults.event.deadline.secs` and `defaults.event.deadline.nanos`
+  Attention: Both 'secs' and 'nanos' must be set together; leaving one
+  unset will cause the configuration to be invalid.
```
elBoberido, line 116: I just had another idea. I haven't thought it fully through, so take it as food for thought.

In the toml config file, the values actually look like this
```
[defaults.event.deadline]
secs = 0
nanos = 0
```

This means the actual entry is just `defaults.event.deadline` and the `secs` and `nanos` are then deserialized into a `Duration`. Could maybe the test be rewritten to just check for the `Duration` type or `Option<Duration>` and not for the fields of `Duration`? The documentation could then just add a hint that the config option is a `Duration` and that is serialized into a `secs` and `nanos`.

elBoberido:
@RTK-jangid can you try to rebase to main. I hope fixes the CI issue.

RTK-jangid:
@elBoberido I already did rebased last time but I don't know why CI is getting cancelled. Its not failing.

elBoberido:
@RTK-jangid I'm also not fully sure why this happens. It seems there is at least one test on FreeBSD that hangs and it seems macOS is also partially affected

> SLOW [>3180.000s] iceoryx2-bb-posix::thread_tests thread_affinity_is_set_to_all_existing_cores_when_nothing_was_configured

After one hour, the CI job runs into a timeout

RTK-jangid:
@elBoberido But I can't think of any file in this PR that can affect the FreeBSD part

Finding new Mysteries Everyday 🥲

elBoberido:
@RTK-jangid me neither :sweat_smile: 

I'll try to setup a FreeBSD environment and try to reproduce the issue next week.

If you are curious, you could also give it a try. I usually use [quickemu](https://github.com/quickemu-project/quickemu) to get a FreeBSD setup.

elBoberido:
@RTK-jangid can you try to increase the memory for FreeBSD in `build-test.yaml` from `16384` to `32768`

RTK-jangid:
>  VM_VBOX="./vbox.sh"
  VM_VBOX_LINK="https://raw.githubusercontent.com/vmactions/vbox/v1.0.4/vbox.sh"===================Debug Info in VM=============
  ssh: Could not resolve hostname freebsd: Name or service not known
  Error: The process '/usr/bin/bash' failed with exit code 1
  
  @elBoberido CI is failing by Increasing Memory. I will set up FreeBSD and try to learn something atleast.

elBoberido:
Maybe one last try with the memory. Can you try 20480?

elBoberido:
@RTK-jangid can you rebase to main once again? @elfenpiff made a commit that made the problem disappear on his branch, which is now merged to main.

RTK-jangid:
@elBoberido no luck

RTK-jangid:
Should I check by reverting each commit where it fails because I know this PR has passed checks previously ?

elBoberido:
@RTK-jangid my excitement was a bit premature. I had the same problem on my branch and an initial rebase to elfenpiffs branch seemed to fix it. But it came back when I rebased to main after his merge. I updated the github actions for FreeBSD and now it seems to build fine. Once my branch is merged, I'll ping you and you can give it another try.

I also tried to reproduce the issue locally, but was not able to do so.

elBoberido:
@RTK-jangid okay, can you try another rebase?

RTK-jangid:
@elBoberido nope 😢

elBoberido:
@RTK-jangid I need to try to figure out how to reproduce the issue. It's weird and it also worries me that this happens. I try to allocate some time but it will probably only be next week.

elBoberido:
@RTK-jangid I just tried to build your branch on my local FreeBSD setup. Unfortunately, it built and the tests also ran fine. I have another idea. The github runner have only 16GB RAM and we configured to VM to also use 16GB. Maybe the VM is just trying to get more memory and does not succeed. My VM uses only 8GB and I could build iceoryx2. Can you try to reduce the `mem` from `16384` to `8192`? It's just a wild guess, but I'm running out of ideas.

elBoberido:
@RTK-jangid it seems it did not work but thanks for trying. I try to find some time for debugging next week.

elBoberido:
@RTK-jangid we were not able to reproduce the issue locally and it became more and more apparent that it is an issue with the github action. We now deactivated the FreeBSD jobs. Can you please rebase again?

elBoberido:
@RTK-jangid okay, it seems I could do the rebase via the github UI. I think this change was introduced with some changes regarding the branch protection we overhauled last week. Need to investigate an move back to the old behavior.

Issue #845, opened 2025-06-29
Generalize tunnel implementation


orecham:
## Brief Description

The current tunnel implementation is tightly coupled to the mechanism used to tunnel the data, i.e. zenoh.
Decoupling the mechanism from the tunnel implementation will make it easier to add tunnels for different mechanisms.

Additionally, the initial implementation was a proof-of-concept, prioritizing functionality over cleanliness and maintainability. This should be addressed during the refactoring.

Issue #841, opened 2025-06-26
Cross language c++ -> Rust type incompatability


matthewashton-k:
I am trying to implement a c++ publisher for point clouds, and a rust subscriber but I run into this issue:
`PublishSubscribeOpenOrCreateError::PublishSubscribeOpenError(IncompatibleTypes)`

My c++ types are here:

```c++
struct PointXYZIR {
    static constexpr const char* IOX2_TYPE_NAME = "PointXYZIR";
    float x;
    float y;
    float z;
    float intensity;
    float time;
    std::uint16_t ring;
};

struct IceoryxPointCloud {
    static constexpr const char* IOX2_TYPE_NAME = "IceoryxPointCloud";
    size_t publish_count;
    PointXYZIR points[6000];
};
```

And my rust types:

```rust
#[repr(C)]
#[derive(Clone, Copy, Debug, Default, Encode, Decode, ZeroCopySend)]
#[type_name("PointXYZIR")]
pub struct PointXYZIR {
    pub x: f32,
    pub y: f32,
    pub z: f32,
    pub intensity: f32,
    pub time: f32,
    pub ring: u16,
}

pub const MAX_POINT_CLOUD_POINTS: usize = 6000;


#[repr(C)]
#[derive(Clone, Debug, ZeroCopySend)]
#[type_name("IceoryxPointCloud")]
pub struct IceoryxPointCloud {
    pub publish_count: usize,
    pub points: [PointXYZIR; MAX_POINT_CLOUD_POINTS],
}
```
I suspect it might be an alignment issue or something, but any help here would be greatly appreciated.

FerdinandSpitzschnueffler:
@matthewashton-k When I replace `TransmissionData` in our [cross language example](https://github.com/eclipse-iceoryx/iceoryx2/tree/main/examples/rust/publish_subscribe_cross_language) with the types you provided, the pub-sub communication is established without problems. Can you please share the log output you receive or a minimal example where the error occurs so that we can check what is going wrong? Please set the log level to `Trace` with `set_log_level(LogLevel::Trace);`.
The problem might be `size_t`/`usize` as compilers are allowed to optimize for <4GB systems. Can you try to use only fixed-size data types?

matthewashton-k:
here is what the trace is:

```
        0 [W] "Config::global_config()"
              | No config file was loaded, a config with default values will be used.
[...]

PR #846, opened 2025-06-29
[#845] Code Cleanliness in Zenoh Tunnel

orecham:
## Notes for Reviewer

No functional changes, only refactoring to make the implementation a little cleaner. Probably more follow-ups to come.

* [ ] Assign PR to reviewer

## References

Relates #845

### DIFF
--- a/iceoryx2-tunnels/zenoh/src/channel/notifier.rs
+++ b/iceoryx2-tunnels/zenoh/src/channel/notifier.rs
@@ -64,7 +64,7 @@ impl<ServiceType: iceoryx2::service::Service> NotifierChannel<'_, ServiceType> {
             from "NotifierChannel::create()",
             when middleware::zenoh::create_notifier(z_session, iox_service_config),
             with CreationError::Error,
-            "failed to create zenoh notifier to propgate local notificaitons to remote hosts"
+            "failed to create zenoh notifier to propagate local notificaitons to remote hosts"
         );
 
         Ok(Self {
--- a/iceoryx2-tunnels/zenoh/src/channel/publisher.rs
+++ b/iceoryx2-tunnels/zenoh/src/channel/publisher.rs
@@ -20,8 +20,8 @@ use iceoryx2::service::builder::CustomHeaderMarker;
 use iceoryx2::service::builder::CustomPayloadMarker;
 use iceoryx2::service::port_factory::publish_subscribe::PortFactory as IceoryxPublishSubscribeService;
 use iceoryx2::service::static_config::StaticConfig as IceoryxServiceConfig;
-use iceoryx2_bb_log::error;
 use iceoryx2_bb_log::fail;
+use iceoryx2_bb_log::fatal_panic;
 use iceoryx2_bb_log::info;
 
 use zenoh::bytes::ZBytes;
@@ -125,8 +125,7 @@ impl<ServiceType: iceoryx2::service::Service> Channel for PublisherChannel<'_, S
                 }
                 Ok(None) => break, // No more samples available
                 Err(e) => {
-                    error!("Failed to receive custom payload from iceoryx: {}", e);
-                    return Err(PropagationError::IceoryxPort);
+                    fatal_panic!("failed to receive custom payload: {e}");
                 }
             }
         }
--- a/iceoryx2-tunnels/zenoh/src/channel/subscriber.rs
+++ b/iceoryx2-tunnels/zenoh/src/channel/subscriber.rs
@@ -19,8 +19,8 @@ use iceoryx2::service::builder::CustomHeaderMarker;
 use iceoryx2::service::builder::CustomPayloadMarker;
 use iceoryx2::service::port_factory::publish_subscribe::PortFactory as IceoryxPublishSubscribeService;
 use iceoryx2::service::static_config::StaticConfig as IceoryxServiceConfig;
-use iceoryx2_bb_log::error;
 use iceoryx2_bb_log::fail;
+use iceoryx2_bb_log::fatal_panic;
 use iceoryx2_bb_log::info;
 
 use zenoh::handlers::FifoChannelHandler;
@@ -128,12 +128,7 @@ impl<ServiceType: iceoryx2::service::Service> Channel for SubscriberChannel<Serv
                         );
                     }
                     Err(e) => {
-                        error!(
-                            "Failed to loan sample ({}): {}",
-                            self.iox_service_config.name(),
-                            e
-                        );
-                        return Err(PropagationError::IceoryxPort);
+                        fatal_panic!("failed to loan custom payload: {e}");
                     }
                 }
             }


iceoryx2-tunnels/zenoh/src/channel/listener.rs
```diff
@@ -0,0 +1,103 @@
+// Copyright (c) 2025 Contributors to the Eclipse Foundation
```
orecham, line 1: Moved from connection/event.rs

iceoryx2-tunnels/zenoh/src/channel/publisher.rs
```diff
@@ -0,0 +1,124 @@
+// Copyright (c) 2025 Contributors to the Eclipse Foundation
```
orecham, line 1: Moved from connection/publish_subscribe.rs

iceoryx2-tunnels/zenoh/src/channel/subscriber.rs
```diff
@@ -0,0 +1,135 @@
+// Copyright (c) 2025 Contributors to the Eclipse Foundation
```
orecham, line 1: Moved from connection/publish_subscribe.rs

iceoryx2-tunnels/zenoh/src/channel/notifier.rs
```diff
@@ -0,0 +1,96 @@
+// Copyright (c) 2025 Contributors to the Eclipse Foundation
```
orecham, line 1: Moved from connection/event.rs

iceoryx2-tunnels/zenoh/src/middleware/iceoryx.rs
```diff
@@ -0,0 +1,108 @@
+// Copyright (c) 2025 Contributors to the Eclipse Foundation
```
orecham, line 1: Moved from to top-level lib.rs

iceoryx2-tunnels/zenoh/src/middleware/zenoh.rs
```diff
@@ -0,0 +1,126 @@
+// Copyright (c) 2025 Contributors to the Eclipse Foundation
```
orecham, line 1: Moved from to top-level lib.rs

iceoryx2-cli/iox2-tunnel/src/main.rs
```diff
@@ -96,9 +97,14 @@ mod supported_platform {
                         let on_event = |id: WaitSetAttachmentId<ipc::Service>| {
                             if id == tick {
                                 if let Err(e) = tunnel.discover(Scope::Both) {
-                                    error!("Failure in discovery: {}", e);
+                                    warn!("Error encountered whilst discoverying services: {}", e);
                                 };
-                                tunnel.propagate();
+                                if let Err(e) = tunnel.propagate() {
```
elfenpiff: I think the rust idiomatic approach is to use `.inspect_err()`

```rust
tunnel.propagate().inspect_err(|e| {
  warn!("Error encountered whilst propagating between hosts: {e}");
});

```

elfenpiff: Could you also adjust the code from above.

iceoryx2-tunnels/zenoh/src/middleware/iceoryx.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+/// Creates an iceoryx event service matching the provided service configuration.
+pub(crate) fn create_event_service<ServiceType: iceoryx2::service::Service>(
+    node: &Node<ServiceType>,
+    service_config: &ServiceConfig,
+) -> Result<EventService<ServiceType>, EventOpenOrCreateError> {
+    let service = node
+        .service_builder(service_config.name())
+        .event()
```
elfenpiff: No setting of `max_listeners` or `max_notifiers` or `max_nodes` or `event_id_max_value` here from the service config?

iceoryx2-tunnels/zenoh/src/middleware/iceoryx.rs
```diff
@@ -0,0 +1,108 @@
...
+    let port_config = service_config.publish_subscribe();
+    let service = unsafe {
+        node.service_builder(service_config.name())
+            .publish_subscribe::<[CustomPayloadMarker]>()
+            .user_header::<CustomHeaderMarker>()
+            .__internal_set_user_header_type_details(
+                &port_config.message_type_details().user_header,
+            )
+            .__internal_set_payload_type_details(&port_config.message_type_details().payload)
+            .enable_safe_overflow(port_config.has_safe_overflow())
```
elfenpiff: What about `max_nodes` or `payload_alignment` or `subscriber_max_buffer_size`?

iceoryx2-tunnels/zenoh/src/tunnel.rs
```diff
@@ -159,134 +138,231 @@ impl<Service: iceoryx2::service::Service> Tunnel<'_, Service> {
...
-    pub fn propagate(&self) {
-        // TODO(correctioness): consolidate and forward errors
-        for (id, connection) in &self.publish_subscribe_connectons {
-            if let Err(e) = connection.propagate() {
-                error!("Failed to propagate ({:?}): {}", id, e);
+    pub fn propagate(&self) -> Result<(), PropagationError> {
+        // Attempted to propagate all channels. Continue to next channel if error encountered.
+        let mut success = true;
+        for (id, channel) in &self.subscriber_channels {
+            if let Err(e) = channel.propagate() {
```
elfenpiff: Please use `channel.propagate().inspect_err(|e|{ //...` here and in all other occurrences in this file.

iceoryx2-tunnels/zenoh/src/tunnel.rs
```diff
@@ -159,134 +138,231 @@ impl<Service: iceoryx2::service::Service> Tunnel<'_, Service> {
...
+                error!(
+                    "Failed to propagate data through listener channel with id {:?}: {}",
+                    id, e
+                );
+                success = false;
             }
         }
+
+        if !success {
+            return Err(PropagationError::Incomplete);
```
elfenpiff: In iceoryx2 we never return an error directly, we always use the `fail!` macro like:
```rust
fail!(from self /* requires that self has implemented debug */, with PropagationError::Incomplete,
   "some meaningful message");
```

Could you adjust this here and all other occurrences in this PR.

iceoryx2-tunnels/zenoh/src/tunnel.rs
```diff
@@ -159,134 +138,231 @@ impl<Service: iceoryx2::service::Service> Tunnel<'_, Service> {
...
+    let iox_service_id = iox_service_config.service_id();
+    let needs_publisher = !publisher_channels.contains_key(iox_service_id);
+    let needs_subscriber = !subscriber_channels.contains_key(iox_service_id);
+
+    if needs_publisher || needs_subscriber {
+        let iox_service = middleware::iceoryx::create_publish_subscribe_service::<ServiceType>(
+            iox_node,
+            iox_service_config,
+        )
+        .map_err(|_e| DiscoveryError::ServiceCreation)?;
```
elfenpiff: Same here, please use fail instead of returning the error directly.

```rust
let iox_service = middleware::iceoryx::create_publish_subscribe_service::<ServiceType>(
            iox_node,
            iox_service_config,
        );

let iox_service = fail!(from self, when iox_service, with DiscoveryError::ServiceCreation,
                                      "meaningful error message");
```

elfenpiff: Please also adjust all other occurrences of `map_err`

elfenpiff (commented):
Please do not return errors directly with `?` or `return Err(Whatever)` and always use the `fail!` macro.

The `fail!` macro enforces an additional `debug!` message and it made debugging bugs in the past already a lot easier. With a consequent use of this we just have to tell the user to provide a log message and in 90% of the cases we see directly where something goes wrong and fix the issue without even reproducing it.

One good example is this bug report from last week: https://github.com/eclipse-iceoryx/iceoryx2/issues/841 where the log showed that the user is trying to send pubsub headers with different size.

iceoryx2-tunnels/zenoh/src/middleware/iceoryx.rs
orecham: @elfenpiff The docs on docs.rs are out of sync 😟: 
1. https://docs.rs/iceoryx2/0.6.1/iceoryx2/#quality-of-services
2. https://docs.rs/iceoryx2/0.6.1/iceoryx2/service/index.html#example

Are the examples in the documentation supposed to be comprehensive or is the expectation that users look at the API docs for the builder to see all options? If the case is the latter, I will adapt the comment in those docs.

orecham: How can I get the `payload_alignment` from the `StaticConfig` ? Is it necessary to set this too when the payload type details are also being set?

iceoryx2-tunnels/zenoh/src/channel/notifier.rs
```diff
@@ -0,0 +1,108 @@
...
+            from "NotifierChannel::create()",
+            when middleware::iceoryx::create_listener(iox_service),
+            with CreationError::Error,
+            "failed to create iceoryx listener for local notifications"
+        );
+        let z_notifier = fail!(
+            from "NotifierChannel::create()",
+            when middleware::zenoh::create_notifier(z_session, iox_service_config),
+            with CreationError::Error,
+            "failed to create zenoh notifier to propgate local notificaitons to remote hosts"
```
elfenpiff: ... to propagate ...

iceoryx2-tunnels/zenoh/src/channel/publisher.rs
```diff
@@ -0,0 +1,136 @@
...
+                    );
+
+                    info!(
+                        "PROPAGATE PublisherChannel {} [{}]",
+                        self.iox_service_config.service_id().as_str(),
+                        self.iox_service_config.name()
+                    );
+                }
+                Ok(None) => break, // No more samples available
+                Err(e) => {
```
elfenpiff, line 127: I was thinking that this should be a `fatal_panic!` and then dug a bit deeper into the code. Currently, a corrupted publisher might corrupt the connection and then `Subscriber::receive()` would always return a `ReceiveError::ConnectionFailure` and would be unable to receive any other data from any other publisher - which would violate the freedom of interference.

We have to refactor this inside the `Subscriber` and potentially other ports and call in this case the degradation callback and continue with the other healthy publisher connections.

So in your case, when you end up here, you can raise a `fatal_panic!` since you should never reach this path. Can only happen when you borrowed more samples than allowed which would mean you have a problem in the tunnel implementation and the other connection failure should actually never happen or propagated here to the user.

elfenpiff: I created this issue to track this: https://github.com/eclipse-iceoryx/iceoryx2/issues/854

iceoryx2-tunnels/zenoh/src/channel/subscriber.rs
```diff
@@ -0,0 +1,144 @@
...
+                        );
+
+                        info!(
+                            "PROPAGATE SubscriberChannel {} [{}]",
+                            self.iox_service_config.service_id().as_str(),
+                            self.iox_service_config.name()
+                        );
+                    }
+                    Err(e) => {
+                        error!(
```
elfenpiff: Please use here `fatal_panic!` instead since a error in this path means always that the tunnel implementation has a bug.

iceoryx2-tunnels/zenoh/src/discovery/iceoryx.rs
```diff
@@ -10,132 +10,161 @@
...
                         }
-                        None => break,
-                    },
-                    Err(_e) => {
-                        return Err(DiscoveryError::Error);
                     }
                 }
             }
+            Ok(None) => break,
+            Err(_) => return Err(DiscoveryError::UpdateFromPort),
```
elfenpiff: Please use here the `fail!` macro.

iceoryx2-tunnels/zenoh/src/discovery/zenoh.rs
```diff
@@ -59,36 +69,46 @@ impl<ServiceType: iceoryx2::service::Service> ZenohDiscovery<'_, ServiceType> {
...
+                                "skipping discovered service config, unable to deserialize: {}",
+                                e
+                            );
                         }
-                        Err(_e) => todo!(),
                     }
                 }
-                Err(_e) => { /* Ignore and process other requests */ }
+                Err(e) => {
+                    warn!("skipping errorneous reply to zenoh discovery query: {}", e)
```
elfenpiff: Would it make sense to also introduce a degradation callback for the tunnel and call it here. Then the user is informed, could in theory adjust their algorithm and can define a reaction from printing out a warning, returning an error result to a panic.

iceoryx2-tunnels/zenoh/src/middleware/iceoryx.rs
elfenpiff: @orecham 

> builder to see all options? If the case is the latter, I will adapt the comment in those docs.

It is the latter, could you please adjust the comment in the docs.

elfenpiff: @orecham 

> How can I get the payload_alignment from the StaticConfig ? 

```rust
 service
     .static_config()
     .message_type_details()
     .payload
     .alignment;
```

> Is it necessary to set this too when the payload type details are also being set?

I think you are right and this does not need to be set in your use case since you already read out the `message_type_details` and configured the service explicitly with them.

iceoryx2-tunnels/zenoh/src/discovery/zenoh.rs
orecham: @elfenpiff 
Not sure if completely see the usage pattern end-to-end here. Maybe worth discussing this, but nevertheless would be out-of-scope for this refactoring PR.

I did notice however, that I was not propagating the error, so the user would not even be able to react to it. I will fix that.

iceoryx2-tunnels/zenoh/src/discovery/zenoh.rs
```diff
@@ -59,36 +69,49 @@ impl<ServiceType: iceoryx2::service::Service> ZenohDiscovery<'_, ServiceType> {
...
+                                e
+                            );
                         }
-                        Err(_e) => todo!(),
                     }
                 }
-                Err(_e) => { /* Ignore and process other requests */ }
+                Err(e) => fail!(
+                    from "discovery_via_subscriber()",
+                    when Err(e),
```
elfenpiff, line 98: You can remove this. But it would be nice to print out the error value in the error message.
```
"errorneous reply received from zenoh discovery query ({e:?})"
```

elfenpiff (approved):

iceoryx2-tunnels/zenoh/src/discovery/zenoh.rs
orecham: Added to: https://github.com/eclipse-iceoryx/iceoryx2/issues/845

Issue #817, opened 2025-06-13
Blackboard messaging pattern


FerdinandSpitzschnueffler:
The blackboard messaging pattern is based on a key-value store. All keys have the same shared-memory-compatible type, like a fixed-size string or an integer, and every value can have a different shared-memory-compatible type.

The pattern shines when:

- there is a large number of readers,
- the update frequency is either low
  - or only a few values are updated at a high frequency,
  - or individual consumers are only interested in a small subset of key-value pairs,
- and the readers do not require a history or previous values.

PR #872, opened 2025-07-14
[#817] blackboard reader and writer implementation

FerdinandSpitzschnueffler:
## Notes for Reviewer

This PR adds the communication endpoints (readers/writers) of the blackboard messaging pattern. The provided examples are for illustrative purposes only and are not yet finished and therefore not documented. I will finish them in a follow-up PR. The cleanup when a node crashes will also be done in a follow-up PR.

* [ ] ~Changelog updated [in the unreleased section][changelog] including API breaking changes~
* [x] Assign PR to reviewer

## References

Relates to #817

### DIFF
--- a/examples/rust/blackboard/creator.rs
+++ b/examples/rust/blackboard/creator.rs
@@ -20,9 +20,10 @@ fn main() -> Result<(), Box<dyn core::error::Error>> {
     set_log_level_from_env_or(LogLevel::Info);
     let node = NodeBuilder::new().create::<ipc::Service>()?;
 
+    type KeyType = u32;
     let service = node
         .service_builder(&"My/Funk/ServiceName".try_into()?)
-        .blackboard_creator::<u32>()
+        .blackboard_creator::<KeyType>()
         .add::<u64>(0, 0)
         .add::<FixedSizeByteString<100>>(5, "Groovy".try_into()?)
         .add::<f32>(9, 0.0)
--- a/examples/rust/blackboard/creator.rs
+++ b/examples/rust/blackboard/creator.rs
@@ -1,0 +1,60 @@
+// Copyright (c) 2025 Contributors to the Eclipse Foundation
+//
+// See the NOTICE file(s) distributed with this work for additional
+// information regarding copyright ownership.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Apache Software License 2.0 which is available at
+// https://www.apache.org/licenses/LICENSE-2.0, or the MIT license
+// which is available at https://opensource.org/licenses/MIT.
+//
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use core::time::Duration;
+use iceoryx2::prelude::*;
+use iceoryx2_bb_container::byte_string::FixedSizeByteString;
+
+const CYCLE_TIME: Duration = Duration::from_secs(1);
+
+fn main() -> Result<(), Box<dyn core::error::Error>> {
+    set_log_level_from_env_or(LogLevel::Info);
+    let node = NodeBuilder::new().create::<ipc::Service>()?;
+
+    let service = node
+        .service_builder(&"My/Funk/ServiceName".try_into()?)
+        .blackboard_creator::<u32>()
+        .add::<u64>(0, 0)
+        .add::<FixedSizeByteString<100>>(5, "Groovy".try_into()?)
+        .add::<f32>(9, 0.0)
+        .create()?;
+
+    let writer = service.writer_builder().create()?;
+
+    let writer_handle_0 = writer.entry::<u64>(&0)?;
+    let mut writer_handle_5 = writer.entry::<FixedSizeByteString<100>>(&5)?;
+    let writer_handle_9 = writer.entry::<f32>(&9)?;
+
+    let mut counter = 0;
+
+    while node.wait(CYCLE_TIME).is_ok() {
+        counter += 1;
+
+        writer_handle_0.update_with_copy(counter);
+        println!("Write new value for key 0: {counter}");
+
+        let entry_value_uninit = writer_handle_5.loan_uninit();
+        let value = format!("Funky {}", counter);
+        let entry_value =
+            entry_value_uninit.write(FixedSizeByteString::<100>::from_bytes(value.as_bytes())?);
+        writer_handle_5 = entry_value.update();
+        println!("Write new value for key 5: {}", value);
+
+        let value = counter as f32 * 7.7;
+        writer_handle_9.update_with_copy(value);
+        println!("Write new value for key 9: {value}");
+    }
+
+    println!("exit");
+
+    Ok(())
+}
--- a/examples/rust/blackboard_event_based_communication/creator.rs
+++ b/examples/rust/blackboard_event_based_communication/creator.rs
@@ -18,11 +18,12 @@ const CYCLE_TIME: Duration = Duration::from_secs(1);
 fn main() -> Result<(), Box<dyn core::error::Error>> {
     set_log_level_from_env_or(LogLevel::Trace);
     let node = NodeBuilder::new().create::<ipc::Service>()?;
+    type KeyType = u32;
     let key = 99;
 
     let service = node
         .service_builder(&"My/Funk/ServiceName".try_into()?)
-        .blackboard_creator::<u32>()
+        .blackboard_creator::<KeyType>()
         .add::<u64>(key, 2023)
         .create()?;
 
--- a/examples/rust/blackboard_event_based_communication/creator.rs
+++ b/examples/rust/blackboard_event_based_communication/creator.rs
@@ -1,0 +1,50 @@
+// Copyright (c) 2025 Contributors to the Eclipse Foundation
+//
+// See the NOTICE file(s) distributed with this work for additional
+// information regarding copyright ownership.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Apache Software License 2.0 which is available at
+// https://www.apache.org/licenses/LICENSE-2.0, or the MIT license
+// which is available at https://opensource.org/licenses/MIT.
+//
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use core::time::Duration;
+use iceoryx2::prelude::*;
+
+const CYCLE_TIME: Duration = Duration::from_secs(1);
+
+fn main() -> Result<(), Box<dyn core::error::Error>> {
+    set_log_level_from_env_or(LogLevel::Trace);
+    let node = NodeBuilder::new().create::<ipc::Service>()?;
+    let key = 99;
+
+    let service = node
+        .service_builder(&"My/Funk/ServiceName".try_into()?)
+        .blackboard_creator::<u32>()
+        .add::<u64>(key, 2023)
+        .create()?;
+
+    let event_service = node
+        .service_builder(&"My/Funk/ServiceName".try_into()?)
+        .event()
+        .open_or_create()?;
+    let listener = event_service.listener_builder().create()?;
+
+    let reader = service.reader_builder().create()?;
+    let reader_handle = reader.entry::<u64>(&key)?;
+
+    // wait for entry id
+    while node.wait(Duration::ZERO).is_ok() {
+        if let Ok(Some(id)) = listener.timed_wait_one(CYCLE_TIME) {
+            if id == reader_handle.entry_id() {
+                println!("read u64: {}", reader_handle.get());
+            }
+        }
+    }
+
+    println!("exit");
+
+    Ok(())
+}
--- a/iceoryx2-bb/lock-free/src/spmc/unrestricted_atomic.rs
+++ b/iceoryx2-bb/lock-free/src/spmc/unrestricted_atomic.rs
@@ -62,7 +62,7 @@ impl<T: Copy> Drop for Producer<'_, T> {
 unsafe impl<T: Copy> Send for Producer<'_, T> {}
 unsafe impl<T: Copy> Sync for Producer<'_, T> {}
 
-/// An atomic implementation where the underlying type has to by copyable but is otherwise
+/// An atomic implementation where the underlying type has to be copyable but is otherwise
 /// unrestricted.
 #[repr(C)]
 pub struct UnrestrictedAtomic<T: Copy> {
@@ -122,7 +122,7 @@ impl<T: Copy> UnrestrictedAtomic<T> {
         /////////////////////////
         // SYNC POINT - write
         // prevent reordering of `data` after advancing `write_cell` which would signal
-        // the completion of the store operation an would result in a data race when
+        // the completion of the store operation and would result in a data race when
         // `data` would be written after the `write_cell` operation
         /////////////////////////
         self.write_cell.fetch_add(1, Ordering::Release);
--- a/iceoryx2-bb/lock-free/src/spmc/unrestricted_atomic.rs
+++ b/iceoryx2-bb/lock-free/src/spmc/unrestricted_atomic.rs
@@ -53,14 +53,27 @@ impl<T: Copy> Producer<'_, T> {
     }
 
     #[doc(hidden)]
-    pub unsafe fn get_ptr_to_write_cell(&self) -> *mut T {
+    // # Safety
+    //
+    //   * the memory position must not be modified after __internal_update_write_cell has been
+    //     called
+    pub unsafe fn __internal_get_ptr_to_write_cell(&self) -> *mut T {
         let write_cell = self.atomic.write_cell.load(Ordering::Relaxed);
         unsafe { (*self.atomic.data[write_cell as usize % NUMBER_OF_CELLS].get()).as_mut_ptr() }
     }
 
     #[doc(hidden)]
-    pub unsafe fn update_write_cell(&self) {
-        self.atomic.write_cell.fetch_add(1, Ordering::Relaxed);
+    // # Safety
+    //
+    //   * the method must not be called without first writing to the memory position returned by
+    //     __internal_get_ptr_to_write_cell
+    pub unsafe fn __internal_update_write_cell(&self) {
+        /////////////////////////
+        // SYNC POINT - write
+        // After writing the content of the write_cell, the content needs to be synced with the
+        // reader.
+        /////////////////////////
+        self.atomic.write_cell.fetch_add(1, Ordering::Release);
     }
 }
 
--- a/iceoryx2-bb/lock-free/src/spmc/unrestricted_atomic.rs
+++ b/iceoryx2-bb/lock-free/src/spmc/unrestricted_atomic.rs
@@ -51,6 +51,17 @@ impl<T: Copy> Producer<'_, T> {
     pub fn store(&self, new_value: T) {
         self.atomic.store(new_value);
     }
+
+    #[doc(hidden)]
+    pub unsafe fn get_ptr_to_write_cell(&self) -> *mut T {
+        let write_cell = self.atomic.write_cell.load(Ordering::Relaxed);
+        unsafe { (*self.atomic.data[write_cell as usize % NUMBER_OF_CELLS].get()).as_mut_ptr() }
+    }
+
+    #[doc(hidden)]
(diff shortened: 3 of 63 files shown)


examples/rust/blackboard/creator.rs
FerdinandSpitzschnueffler, line 1: The provided examples are for illustrative purposes only and are not yet finished and therefore not documented. This will be done in a follow-up.

examples/rust/blackboard_event_based_communication/creator.rs
```diff
@@ -0,0 +1,50 @@
...
+        .add::<u64>(key, 2023)
+        .create()?;
+
+    let event_service = node
+        .service_builder(&"My/Funk/ServiceName".try_into()?)
+        .event()
+        .open_or_create()?;
+    let listener = event_service.listener_builder().create()?;
+
+    let reader = service.reader_builder().create()?;
```
elfenpiff, line 36: I know it is just for demonstration and not yet the final example but I would move the reader into the opener and the writer into the creator. Otherwise, it somehow is counter-intuitive and I start to wonder why it is the other way around.

But this can be done in the blackboard example pr

iceoryx2-bb/lock-free/src/spmc/unrestricted_atomic.rs
```diff
@@ -51,6 +51,17 @@ impl<T: Copy> Producer<'_, T> {
...
+
+    #[doc(hidden)]
+    pub unsafe fn get_ptr_to_write_cell(&self) -> *mut T {
+        let write_cell = self.atomic.write_cell.load(Ordering::Relaxed);
+        unsafe { (*self.atomic.data[write_cell as usize % NUMBER_OF_CELLS].get()).as_mut_ptr() }
+    }
+
+    #[doc(hidden)]
+    pub unsafe fn update_write_cell(&self) {
+        self.atomic.write_cell.fetch_add(1, Ordering::Relaxed);
```
elfenpiff: I think the memory order here must be `Release`. I assume that one first writes the content of the write cell and when the update of the pointer was finished `update_write_cell` is called. When now the other side, the reader, consumes the new memory location the underlying payload needs to be synced - therefore it needs to be `Release`.

iceoryx2-bb/lock-free/src/spmc/unrestricted_atomic.rs
```diff
@@ -51,6 +51,17 @@ impl<T: Copy> Producer<'_, T> {
     pub fn store(&self, new_value: T) {
         self.atomic.store(new_value);
     }
+
+    #[doc(hidden)]
+    pub unsafe fn get_ptr_to_write_cell(&self) -> *mut T {
```
elfenpiff: Even when the method is hidden, could you please add a `// # Safety` section here and define all contracts that the user must ensure that they are fulfilled.
I assume that this memory position shall not be modified after `update_write_cell` was called.
Maybe there is even more.

The same goes for `update_write_cell`. Could you maybe even add the prefix `__internal_` to those two methods to make it really clear that those shall not be used.
When reviewing some piece of code one does not see directly the `#[doc(hidden)]` annotation, just the pure function call and then an `__internal_` prefix becomes another signaling flag.

iceoryx2-bb/lock-free/src/spmc/unrestricted_atomic.rs
elfenpiff: Can you add here the `// SYNC POINT - write` comment and explain the synchronization like it is done in the other parts of the code.

iceoryx2-bb/lock-free/tests/spmc_unrestricted_atomic_tests.rs
```diff
@@ -108,3 +108,19 @@ fn spmc_unrestricted_atomic_load_store_works_concurrently() {
         });
     });
 }
+
+#[test]
+fn spmc_unrestricted_atomic_get_ptr_write_and_update_works() {
```
elfenpiff, line 113: Can you add another concurrent tests. Two threads, one thread reads the value in an infinite busy loop while the other one is using this API to write an incrementing large `u128` number in there.

When some memory order is wrong miri (whenever we activate her) should be able to detect it.

iceoryx2/src/lib.rs
```diff
@@ -370,6 +433,30 @@
...
+//! For a detailed documentation see the
+//! [`blackboard::Creator`](crate::service::builder::blackboard::Creator)
+//!
+//! ```
+//! use iceoryx2::prelude::*;
+//!
+//! # fn main() -> Result<(), Box<dyn core::error::Error>> {
+//! let node = NodeBuilder::new().create::<ipc::Service>()?;
+//!
+//! let service = node.service_builder(&"My/Funk/ServiceName".try_into()?)
```
elfenpiff, line 450: It is not so intuitively clear that `u64` is here the key type. Could you please add a comment here and mention this.

Or even better, create a type alias `type KeyType = u64` and use it here, then you do not need a comment. Could you adjust also the other examples accordingly.

iceoryx2/src/port/reader.rs
```diff
@@ -10,5 +10,305 @@
...
+
+extern crate alloc;
+use alloc::sync::Arc;
+
+use super::port_identifiers::UniqueReaderId;
+
+#[derive(Debug)]
+struct ReaderSharedState<
+    Service: service::Service,
+    KeyType: Send + Sync + Eq + Clone + Debug + 'static,
```
elfenpiff: I think we should add `Hash` here as a requirement from the begin with. This is something required when we would implement a shared memory compatible map (in rust HashMap). It would hash the entry first and then add it into the tree structure.

If we add this later we would restrict the API later which could cause some trouble on the user side. If we never use it, we can lift the restriction later and it would cause no trouble on our users side.

iceoryx2/src/port/reader.rs
```diff
@@ -10,5 +10,305 @@
...
+                .dynamic_storage
+                .get()
+                .blackboard()
+                .release_reader_handle(handle)
+        }
+    }
+}
+
+/// Defines a failure that can occur when a [`Reader`] is created with
+/// [`crate::service::port_factory::reader::PortFactoryReader`].
```
elfenpiff: Can you please hide the full link path like ``[`PortFactoryReader`](crate::service::port_factory::reader::PortFactoryReader)`` to improve readability.

iceoryx2/src/port/reader.rs
```diff
@@ -10,5 +10,305 @@
...
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum ReaderCreateError {
+    /// The maximum amount of [`Reader`]s that can connect to a
+    /// [`Service`](crate::service::Service) is
+    /// defined in [`crate::config::Config`]. When this is exceeded no more [`Reader`]s
```
elfenpiff: Please hide the full link path.

iceoryx2/src/port/reader.rs
```diff
@@ -10,5 +10,305 @@
...
+    /// can be created for a specific [`Service`](crate::service::Service).
```
elfenpiff, line 88: Please hide the full link path.

iceoryx2/src/port/reader.rs
```diff
@@ -10,5 +10,305 @@
...
+            None => {
+                fail!(from origin, with ReaderCreateError::ExceedsMaxSupportedReaders,
+                            "{} since it would exceed the maximum supported amount of readers of {}.",
+                            msg, service.static_config.blackboard().max_readers);
+            }
+        };
+
+        match Arc::get_mut(&mut new_self.shared_state) {
+            None => {
+                fail!(from origin, with ReaderCreateError::InternalFailure,
```
elfenpiff: Could you please use `fatal_panic!` here since it would be clearly an implementation bug when the arc would have multiple references already at this point.

iceoryx2/src/port/reader.rs
```diff
@@ -10,5 +10,305 @@
...
+                "{} since no entry with the given key exists.", msg);
+        }
+
+        let entry = &self
+            .shared_state
+            .service_state
+            .additional_resource
+            .mgmt
+            .get()
+            .entries[index.unwrap()];
```
elfenpiff: Please never use unwrap in production code. You could write it with an `if let` expression or a match statement.

For instance:

```rust
let index = match unsafe { self.shared_state.service_state.additional_resource.mgmt.get().map.get(key) } {
  Some(v) = v,
  None => {
    fail!(from self, with ReaderHandleError::EntryDoesNotExist,
                "{} since no entry with the given key exists.", msg);
  }
}
```

elfenpiff: If you have an unwrap somewhere else, please change it as well.

iceoryx2/src/port/writer.rs
```diff
@@ -10,5 +10,532 @@
...
+
+extern crate alloc;
+use alloc::sync::Arc;
+
+use super::port_identifiers::UniqueWriterId;
+
+#[derive(Debug)]
+struct WriterSharedState<
+    Service: service::Service,
+    KeyType: Send + Sync + Eq + Clone + Debug + 'static,
```
elfenpiff: Same thing here, could we add `Hash` as an additional restriction.

iceoryx2/src/port/writer.rs
```diff
@@ -10,5 +10,532 @@
...
+    /// # }
+    /// ```
+    pub fn entry<ValueType: Copy + ZeroCopySend>(
+        &self,
+        key: &KeyType,
+    ) -> Result<WriterHandle<Service, KeyType, ValueType>, WriterHandleError> {
+        let msg = "Unable to create writer handle";
+
+        // check if key exists
+        let index = unsafe {
```
elfenpiff: Please pack this into a match statement and do not use `index.unwrap()` when acquiring the entry.

iceoryx2/src/service/builder/blackboard.rs
```diff
@@ -140,65 +151,73 @@ impl From<ServiceAvailabilityState> for BlackboardCreateError {
...
 }
 
-impl core::error::Error for BlackboardOpenOrCreateError {}
+#[derive(Debug)]
+pub(crate) struct Entry {
+    pub(crate) type_details: TypeDetail,
+    // offset is sufficient for static single-writer case (no insert after creation, no remove)
+    // dynamic single-writer case (additional inserts after service creation and remove): offset (32 bit) + aba counter (32 bit)
+    // dynamic multi-writer case: offset (32 bit) + writer segment id (8 bit) + aba counter (24 bit)
+    pub(crate) offset: IoxAtomicU64,
```
elfenpiff, line 172: In a follow up PR we could create a separate type for this. So that this does not have to be documented.

Something which does all the bitshifting for us, is testable and the contract is carved into the API.

```rust
struct BlackboardOffset(IoxAtomicU64); // so it is essentially an atomic

impl BlackboardOffset {
  pub fn new(offset: u32, segment_id: u8, aba_counter: u32) -> Self;
  pub fn offset(&self) -> u32;
  pub fn writer_segment_id(&self) -> u8;
  pub fn aba_counter(&self) -> u32;
}
```

Could you track this in some issue please.

iceoryx2/src/service/builder/blackboard.rs
```diff
@@ -246,33 +265,341 @@ impl<KeyType: ZeroCopySend + Debug, ServiceType: service::Service> Builder<KeyTy
...
     }
 
-    fn prepare_config_details(&mut self) {
-        self.config_details_mut().type_details =
-            TypeDetail::__internal_new::<KeyType>(message_type_details::TypeVariant::FixedSize);
+    /// Adds key-value pairs to the blackboard.
+    pub fn add<ValueType: ZeroCopySend + Copy + 'static>(
+        mut self,
+        key: KeyType,
+        value: ValueType,
```
elfenpiff, line 330: Not for this PR! 

I think we should be able to extend this easily to `RelocatableTypes` and dynamic slices. The `TypeDetail` would in this case be `DynamicSize` and the rest stays the same. At the end, where we sum up the memory we would need to be able to ask the `RelocatableType` how much it requires but this could be done also here with `internal_value_size` where we would add for relocatable types or slices the len and required memory.

Could you please track this in some issue.

iceoryx2/src/service/builder/blackboard.rs
```diff
@@ -246,33 +265,341 @@ impl<KeyType: ZeroCopySend + Debug, ServiceType: service::Service> Builder<KeyTy
...
+                        .supplementary_size(RelocatableFlatMap::<KeyType, usize>::const_memory_size(capacity)+RelocatableVec::<Entry>::const_memory_size(capacity))
+                        .initializer(|entry: &mut Mgmt<KeyType>, allocator: &mut BumpAllocator| {
+                            if unsafe {entry.map.init(allocator)}.is_err() || unsafe {entry.entries.init(allocator).is_err()} {
+                                return false
+                            }
+                            for i in 0..capacity {
+                                // write value passed to add() to payload_shm
+                                let mem = payload_shm.allocate(unsafe { Layout::from_size_align_unchecked(self.builder.internals[i].internal_value_size, self.builder.internals[i].internal_value_alignment) });
+                                if mem.is_err() {
+                                    error!("Writing the value to the blackboard data segment failed.");
```
elfenpiff: `error!` also supports the `from self` syntax. Could you please use it here for all error logs?

So adjust it to `error!(from self, "Writing ....`

iceoryx2/src/service/builder/mod.rs
```diff
@@ -177,15 +177,34 @@ impl<S: Service> Builder<S> {
 
     /// Create a new builder to create a
     /// [`MessagingPattern::Blackboard`](crate::service::messaging_pattern::MessagingPattern::Blackboard) [`Service`].
-    pub fn blackboard<KeyType: ZeroCopySend + Debug>(self) -> blackboard::Builder<KeyType, S> {
+    pub fn blackboard_creator<
+        KeyType: Send + Sync + Eq + Clone + Debug + 'static + ZeroCopySend,
```
elfenpiff: Please add `Hash` here.

iceoryx2/src/service/builder/mod.rs
```diff
@@ -177,15 +177,34 @@ impl<S: Service> Builder<S> {
...
+                self.shared_node.config(),
+            ),
+            self.shared_node,
+        )
+        .blackboard_creator()
+    }
+
+    /// Create a new builder to open a
+    /// [`MessagingPattern::Blackboard`](crate::service::messaging_pattern::MessagingPattern::Blackboard) [`Service`].
+    pub fn blackboard_opener<KeyType: Send + Sync + Eq + Clone + Debug + 'static + ZeroCopySend>(
```
elfenpiff: Please add `Hash` here

iceoryx2/src/service/dynamic_config/event.rs
```diff
@@ -103,12 +103,12 @@ impl DynamicConfig {
             + Container::<NotifierDetails>::memory_size(config.number_of_notifiers)
     }
 
-    /// Returns the how many [`crate::port::listener::Listener`] ports are currently connected.
+    /// Returns how many [`crate::port::listener::Listener`] ports are currently connected.
```
elfenpiff: Can you please hide the link for better readability.

iceoryx2/src/service/dynamic_config/event.rs
```diff
@@ -103,12 +103,12 @@ impl DynamicConfig {
...
     pub fn number_of_listeners(&self) -> usize {
         self.listeners.len()
     }
 
-    /// Returns the how many [`crate::port::notifier::Notifier`] ports are currently connected.
+    /// Returns how many [`crate::port::notifier::Notifier`] ports are currently connected.
```
elfenpiff: Can you please hide the link for better readability.

iceoryx2/src/service/mod.rs
```diff
@@ -695,11 +740,17 @@ pub trait Service: Debug + Sized + internal::ServiceInternal<Self> + Clone {
...
     /// [`MutexProtected`](iceoryx2_cal::arc_sync_policy::mutex_protected::MutexProtected), the
     /// [`Service`]s ports are threadsafe and the payload can be moved into threads. If it is set
-    /// to to [`SingleThreaded`](iceoryx2_cal::arc_sync_policy::single_threaded::SingleThreaded),
+    /// to [`SingleThreaded`](iceoryx2_cal::arc_sync_policy::single_threaded::SingleThreaded),
     /// the [`Service`]s ports and payload cannot be shared ([`Sync`]) between threads or moved
     /// ([`Send`]) into other threads.
     type ArcThreadSafetyPolicy<T: Send + Debug>: ArcSyncPolicy<T>;
 
+    /// Defines the construct used to store the management data of the blackboard service.
+    type BlackboardMgmt<KeyType: Send + Sync + Debug + 'static>: DynamicStorage<KeyType>;
```
elfenpiff, line 748: Not in this PR.

Could we adjust the `DynamicStorage` type alias to be a generic alias like `trait DynamicStorage<T>: DynamicStorage<T>` and use it here like `type BlackboardMgmt<...>: Self::DynamicStorage<KeyType>` or even better - remove the type alias for `BlackboardMgmt` and `BlackboardPayload` (when we generalize the SharedMemory type alias as well).

It would have the big advantage that an inconsistent service variant would not be creatable. An inconsistent variant would be for instance a local variant that uses by accident posix shared memory for the blackboard.
The idea is to define the abstract concept implementation ones and then use it everywhere else.

Could you please track this in some issue.

examples/rust/blackboard_event_based_communication/creator.rs
FerdinandSpitzschnueffler: I'll change this in the follow-up.

iceoryx2-bb/lock-free/src/spmc/unrestricted_atomic.rs
FerdinandSpitzschnueffler: Agree

iceoryx2/src/service/builder/blackboard.rs
FerdinandSpitzschnueffler: #880

iceoryx2/src/service/builder/blackboard.rs
FerdinandSpitzschnueffler: Please see #881 and add information if I missed something.

iceoryx2/src/service/mod.rs
FerdinandSpitzschnueffler: I created #882. I'm not sure if I fully understand your idea, so please have a look and adjust the issue if necessary.

iceoryx2/src/service/dynamic_config/blackboard.rs
```diff
@@ -37,30 +56,78 @@ pub struct ReaderDetails {
...
+            when self.writers.init(allocator),
+            "This should never happen! Unable to initialize writer port id container.");
     }
 
     pub(crate) fn memory_size(config: &DynamicConfigSettings) -> usize {
         Container::<ReaderDetails>::memory_size(config.number_of_readers)
+            + Container::<WriterDetails>::memory_size(config.number_of_writers)
+    }
+
+    /// Returns how many [`crate::port::reader::Reader`] ports are currently connected.
```
elfenpiff, line 103: Can be done in another PR when you work on this. Could you please hide the link.

iceoryx2/src/service/dynamic_config/blackboard.rs
```diff
@@ -37,30 +56,78 @@ pub struct ReaderDetails {
...
+    pub fn number_of_readers(&self) -> usize {
+        self.readers.len()
+    }
+
+    /// Returns how many [`crate::port::writer::Writer`] ports are currently connected.
```
elfenpiff, line 108: Can be done in another PR when you work on this. Could you please hide the link.

iceoryx2/src/service/dynamic_config/blackboard.rs
FerdinandSpitzschnueffler: Sure, I will change this in the next PR.

Issue #884, opened 2025-07-19
CLI tooling for pubsub


elfenpiff:
## Brief feature description

It shall be possible to send and receive pubsub messages via `iox2`.

orecham:
Duplicate of https://github.com/eclipse-iceoryx/iceoryx2/issues/413

PR #892, opened 2025-07-22
[#884] pubsub `iox2` with simple record & replay

elfenpiff:
## Notes for Reviewer
This PR allows to send and receive samples from a pubsub service. Additionally, it adds minimalistic support for record & replay. The record & replay feature is intended for development and simple debugging use cases and not for full-blown deterministic use-cases.

Example:
1. run any example, for instance `cargo run --example publish_subscribe_publisher` and `cargo run --example publish_subscribe_subscriber`
2. receive messages and record the payload, for 3000 milliseconds, in hex format to `record.out` `cargo run --bin iox2 -- service-debug subscribe -t 3000 "My/Funk/ServiceName" -o record.out -d HEX`
3. replay messages repeatedly for 9999 times and send a message every 850 milliseconds `cargo run --bin iox2 -- service-debug publish "My/Funk/ServiceName" --type-name TransmissionData --type-size16 --type-alignment 8 --type-variant FIXEDSIZE -d HEX -r 9999 -i record.out -t 850`

* [ ] Assign PR to reviewer

## References

Closes #

### DIFF
--- a/iceoryx2-bb/posix/src/file_descriptor.rs
+++ b/iceoryx2-bb/posix/src/file_descriptor.rs
@@ -137,9 +137,9 @@ impl FileDescriptor {
             return None;
         }
 
-        Some(FileDescriptor {
-            value,
-            is_owned: false,
+        Self::new(value).map(|mut fd| {
+            fd.is_owned = false;
+            fd
         })
     }
 
--- a/iceoryx2-cli/iox2-service/src/cli.rs
+++ b/iceoryx2-cli/iox2-service/src/cli.rs
@@ -183,11 +183,20 @@ pub struct PublishOptions {
     pub time_between_messages: usize,
 
     #[clap(
+        short,
         long,
         default_value = "TEXT",
         help = "Defines how the provided data is encoded."
     )]
     pub data_representation: DataRepresentation,
+
+    #[clap(
+        short,
+        long,
+        default_value = "1",
+        help = "How often shall the messages be sent."
+    )]
+    pub repetitions: usize,
 
     #[clap(
         long,
@@ -250,6 +259,7 @@ pub struct SubscribeOptions {
     pub node_name: String,
 
     #[clap(
+        short,
         long,
         default_value = "HEX",
         help = "Defines how the data shall be displayed."
--- a/iceoryx2-cli/iox2-service/src/cli.rs
+++ b/iceoryx2-cli/iox2-service/src/cli.rs
@@ -262,6 +262,13 @@ pub struct SubscribeOptions {
         help = "Maximum number of messages to be received before the process stops."
     )]
     pub max_messages: Option<u64>,
+
+    #[clap(
+        short,
+        long,
+        help = "When set, the received data is additionally added to the provided file."
+    )]
+    pub output_file: Option<String>,
 }
 
 #[derive(Subcommand)]
--- a/iceoryx2-cli/iox2-service/src/cli.rs
+++ b/iceoryx2-cli/iox2-service/src/cli.rs
@@ -153,7 +153,7 @@ pub enum CliTypeVariant {
 #[derive(Clone, Copy, ValueEnum)]
 #[value(rename_all = "UPPERCASE")]
 pub enum DataRepresentation {
-    Text,
+    Raw,
     Hex,
 }
 
@@ -185,7 +185,7 @@ pub struct PublishOptions {
     #[clap(
         short,
         long,
-        default_value = "TEXT",
+        default_value = "HEX",
         help = "Defines how the provided data is encoded."
     )]
     pub data_representation: DataRepresentation,
@@ -286,6 +286,9 @@ pub struct SubscribeOptions {
         help = "When set, the received data is additionally added to the provided file."
     )]
     pub output_file: Option<String>,
+
+    #[clap(action, short, long, help = "Do not show the received data.")]
+    pub quiet: bool,
 }
 
 #[derive(Subcommand)]
--- a/iceoryx2-cli/iox2-service/src/cli.rs
+++ b/iceoryx2-cli/iox2-service/src/cli.rs
@@ -153,7 +153,7 @@ pub enum CliTypeVariant {
 #[derive(Clone, Copy, ValueEnum)]
 #[value(rename_all = "UPPERCASE")]
 pub enum DataRepresentation {
-    Raw,
+    Iox2Dump,
     Hex,
 }
 
@@ -174,6 +174,8 @@ pub struct PublishOptions {
         help = "The messages that shall be sent. Can be multiple messages. If no messages are given stdin is read."
     )]
     pub message: Vec<String>,
+    #[clap(short, long, help = "When set, the data from this file is published.")]
+    pub input_file: Option<String>,
     #[clap(
         short,
         long,
--- a/iceoryx2-cli/iox2-service/src/cli.rs
+++ b/iceoryx2-cli/iox2-service/src/cli.rs
@@ -13,6 +13,7 @@
 use clap::Args;
 use clap::Parser;
 use clap::Subcommand;
+use clap::ValueEnum;
 
 use iceoryx2_cli::filter::MessagingPatternFilter;
 use iceoryx2_cli::help_template;
@@ -142,6 +143,95 @@ pub struct ListenOptions {
     pub repetitions: Option<u64>,
 }
 
+#[derive(Clone, Copy, ValueEnum)]
+#[value(rename_all = "UPPERCASE")]
+pub enum CliTypeVariant {
+    Dynamic,
+    FixedSize,
+}
+
+#[derive(Parser)]
+pub struct PublishOptions {
+    #[clap(help = "Name of the service which shall the message be sent to.")]
+    pub service: String,
+    #[clap(
+        short,
+        long,
+        default_value = "shell_node",
+        help = "Defines the node name of the publish endpoint."
+    )]
+    pub node_name: String,
+    #[clap(short, long, help = "The messages that shall be sent.")]
+    pub message: Vec<String>,
+    #[clap(
+        short,
+        long,
+        default_value = "1000",
+        help = "Time between the messages in milliseconds."
+    )]
+    pub time_between_messages: usize,
+
+    #[clap(
+        long,
+        default_value = "4096",
+        help = "It defines the initial payload size for dynamic type variants."
+    )]
+    pub initial_payload_size: usize,
+
+    #[clap(
+        long,
+        default_value = "u8",
+        help = "Defines the unique type identifier of the services type."
+    )]
+    pub type_name: String,
+    #[clap(
+        long,
+        default_value = "1",
+        help = "Defines the type size of the services type."
+    )]
+    pub type_size: usize,
+    #[clap(
+        long,
+        default_value = "1",
+        help = "Defines the type alignment of the services type."
+    )]
+    pub type_alignment: usize,
+    #[clap(long, default_value = "DYNAMIC", help = "Defines variant.")]
+    pub type_variant: CliTypeVariant,
+
+    #[clap(
+        long,
+        default_value = "()",
+        help = "Defines the unique type identifier of the services user header type."
+    )]
+    pub header_type_name: String,
+    #[clap(
+        long,
+        default_value = "0",
+        help = "Defines the type size of the services user header type."
+    )]
+    pub header_type_size: usize,
+    #[clap(
+        long,
+        default_value = "1",
+        help = "Defines the type alignment of the services user header type."
+    )]
+    pub header_type_alignment: usize,
+}
+
+#[derive(Parser)]
+pub struct SubscribeOptions {
+    #[clap(help = "Name of the service which shall be waited on for a message.")]
+    pub service: String,
+    #[clap(
+        short,
+        long,
+        default_value = "shell_node",
+        help = "Defines the node name of the subscriber endpoint."
+    )]
+    pub node_name: String,
+}
+
 #[derive(Subcommand)]
 pub enum Action {
     #[clap(
@@ -169,4 +259,14 @@ pub enum Action {
         help_template = help_template(HelpOptions::DontPrintCommandSection)
     )]
     Listen(ListenOptions),
+    #[clap(
+        about = "Publish a message",
+        help_template = help_template(HelpOptions::DontPrintCommandSection)
+    )]
+    Publish(PublishOptions),
+    #[clap(
+        about = "Receive messages",
+        help_template = help_template(HelpOptions::DontPrintCommandSection)
+    )]
+    Subscribe(SubscribeOptions),
 }
--- a/iceoryx2-cli/iox2-service/src/cli.rs
+++ b/iceoryx2-cli/iox2-service/src/cli.rs
@@ -171,7 +171,7 @@ pub struct PublishOptions {
     #[clap(
         short,
         long,
-        help = "The messages that shall be sent. Can be multiple messages."
+        help = "The messages that shall be sent. Can be multiple messages. If no messages are given stdin is read."
     )]
     pub message: Vec<String>,
     #[clap(
@@ -181,6 +181,13 @@ pub struct PublishOptions {
         help = "Time between the messages in milliseconds."
     )]
     pub time_between_messages: usize,
+
+    #[clap(
+        long,
+        default_value = "TEXT",
+        help = "Defines how the provided data is encoded."
+    )]
+    pub data_representation: DataRepresentation,
 
     #[clap(
         long,
--- a/iceoryx2-cli/iox2-service/src/cli.rs
+++ b/iceoryx2-cli/iox2-service/src/cli.rs
@@ -248,6 +248,20 @@ pub struct SubscribeOptions {
         help = "Defines how the data shall be displayed."
     )]
     pub data_representation: DataRepresentation,
+
+    #[clap(
+        short,
+        long,
+        help = "Maximum runtime in milliseconds. When the timeout has passed the process stops."
+    )]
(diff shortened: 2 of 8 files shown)


iceoryx2-bb/posix/src/file_descriptor.rs
```diff
@@ -137,9 +137,9 @@ impl FileDescriptor {
             return None;
         }
 
-        Some(FileDescriptor {
-            value,
-            is_owned: false,
+        Self::new(value).map(|mut fd| {
+            fd.is_owned = false;
+            fd
```
elBoberido, line 142: This might break the VxWorks MVP. I have to check.

iceoryx2-cli/iox2-service/src/command_publish.rs
```diff
@@ -0,0 +1,219 @@
...
+fn hex_string_to_raw_data(hex_string: &str) -> Result<Vec<u8>> {
+    let mut hex_string = hex_string.to_string();
+    hex_string.retain(|c| !c.is_whitespace());
+    (0..hex_string.len())
+        .step_by(2)
+        .map(|n| {
+            u8::from_str_radix(&hex_string[n..n + 2], 16)
+                .map_err(|e| anyhow::anyhow!("Invalid hex input at position {}: {}", n, e))
+        })
+        .collect::<Result<Vec<u8>>>()
```
elBoberido: This is slightly shorter and also more robust.
```suggestion
    hex_string
        .split_ascii_whitespace()
        .map(|hex| {
            let hex = hex.trim_start_matches("0x");
            u8::from_str_radix(&hex, 16)
                .map_err(|e| anyhow::anyhow!("Invalid hex input at position {}: {}", hex, e))
        })
        .collect::<Result<Vec<u8>>>()
```

iceoryx2-cli/iox2-service/src/command_publish.rs
```diff
@@ -0,0 +1,219 @@
...
+        CliTypeVariant::Dynamic => service
+            .publisher_builder()
+            .initial_max_slice_len(4096)
+            .allocation_strategy(AllocationStrategy::PowerOfTwo)
+            .create()?,
+    };
+
+    let mut message_buffer = vec![];
+    read_file_into_buffer(&mut message_buffer, &options)?;
+    read_cli_msg_into_buffer(&mut message_buffer, &options)?;
```
elBoberido: Can you elaborate why you want to support file and cli messages at the same time?

iceoryx2-cli/iox2-service/src/command_publish.rs
```diff
@@ -0,0 +1,219 @@
...
+
+fn read_cli_msg_into_buffer(
+    message_buffer: &mut Vec<(Vec<u8>, Vec<u8>)>,
+    options: &PublishOptions,
+) -> Result<()> {
+    for message in &options.message {
+        match options.data_representation {
+            DataRepresentation::Iox2Dump => {
+                message_buffer.push((vec![], message.as_bytes().to_vec()))
+            }
```
elBoberido, line 103: I think `Iox2Dump` does not make sense for cli msg since there will be non printable characters and also characters representing `enter` on the keyboard.

iceoryx2-cli/iox2-service/src/command_publish.rs
```diff
@@ -0,0 +1,219 @@
...
+                        None => header = Some(hex_string_to_raw_data(line)?),
+                        Some(ref h) => {
+                            message_buffer.push((h.clone(), hex_string_to_raw_data(line)?));
+                            header = None;
+                        }
+                    }
+                }
+            }
+            DataRepresentation::Iox2Dump => {
+                let mut file = File::open(file)?;
```
elBoberido: Did you consider to use a `BufReader`?

There are a lot potentially small reads which can be slow without a `BufReader`.

iceoryx2-cli/iox2-service/src/command_publish.rs
```diff
@@ -0,0 +1,219 @@
...
+                    file.read_exact(&mut header)?;
+
+                    file.read_exact(&mut buffer)?;
+                    let payload_len = u64::from_le_bytes(buffer);
+                    let mut payload = vec![0u8; payload_len as usize];
+                    file.read_exact(&mut payload)?;
+
+                    message_buffer.push((header, payload));
+                    Ok(())
+                };
```
elBoberido: Did you consider to create a minimal file format?

For example, adding magic bytes to the file like `iox2`, `iceoryx2` or `iox2dump` and a version number. Having meta information like the header and payload type, size and alignment would also make the recording more robust. The messaging pattern and the service name would also be nice. This would make it extensible and prevent us from headaches when we update the file format and our users are still using the old files.
For the data stream I would also suggest to use a small custom header for each stored sample with sequence number, maybe timestamp and size of header and payload.

What do you think about this suggestions?

iceoryx2-cli/iox2-service/src/command_subscribe.rs
```diff
@@ -0,0 +1,206 @@
...
+use std::time::Duration;
+use std::time::Instant;
+
+fn raw_data_to_hex_string(raw_data: &[u8]) -> String {
+    let mut ret_val = String::with_capacity(2 * raw_data.len());
+    for byte in raw_data {
+        ret_val.push_str(&format!("{byte:0>2x} "));
+    }
+
+    ret_val
```
elBoberido: This is a bit more efficient. The string will already have the right size since 3 ascii chars will be added, due to the whitespace and also the `write` macro writes directly into the buffer without allocating, unlike the `format` macro.
```suggestion
// somewhere at the top
use std::fmt::Write as _;

// ...
    let mut ret_val = String::with_capacity(raw_data.len() * 3);
    for byte in raw_data {
        let _ = write!(&mut ret_val, "{byte:02x} ");
    }

    ret_val
```

One could also think of getting rid of the whitespace at the end by using `" {byte:02x}"` and not adding a whitespace at the place where the string is used

elBoberido (commented):
Good first start. Can you please also add end-to-end tests for the new functionality?

This will easily break if not tested from the beginning.

iceoryx2-cli/iox2-service/src/cli.rs
```diff
@@ -169,4 +320,14 @@ pub enum Action {
         help_template = help_template(HelpOptions::DontPrintCommandSection)
     )]
     Listen(ListenOptions),
+    #[clap(
+        about = "Publish a message",
+        help_template = help_template(HelpOptions::DontPrintCommandSection)
+    )]
+    Publish(PublishOptions),
+    #[clap(
+        about = "Receive messages",
```
elBoberido: Subscribe to any service and introspect its messages.

iceoryx2-cli/iox2-service/src/cli.rs
```diff
@@ -169,4 +320,14 @@ pub enum Action {
         help_template = help_template(HelpOptions::DontPrintCommandSection)
     )]
     Listen(ListenOptions),
+    #[clap(
+        about = "Publish a message",
```
elBoberido: Publish a message to any service.

iceoryx2-cli/iox2-service/src/cli.rs
```diff
@@ -142,6 +143,156 @@ pub struct ListenOptions {
...
+    )]
+    pub data_representation: DataRepresentation,
+
+    #[clap(
+        short,
+        long,
+        default_value = "1",
+        help = "How often shall the messages be sent."
+    )]
+    pub repetitions: usize,
```
elBoberido, line 231: I think it should be not too difficult to have something like this

```rs
enum Repetitions {
    Count(u64),
    Infinite,
}

File: iceoryx2-bb/posix/src/file_descriptor.rs (hunk 1 above)
Comment (elfenpiff) [reply iceoryx2-bb/posix/src/file_descriptor.rs:142]: I realized that this was broken here so this fix is valid.

File: iceoryx2-cli/iox2-service/src/command_publish.rs (hunk 2 above)
Comment (elfenpiff) [reply iceoryx2-cli/iox2-service/src/command_publish.rs]: I would not go this route with the "0x" trim. The file that shall be used for replay shall come from the CLI tooling. It is a little debugging/introspection helper and not a generic record & replay tool.

**Comment by elBoberido** [comment]
Oh, and the amout of bytes dumped to the terminal schould be limited, e.g. 40 bytes by defaut and the user schould be able to override it. Just imagine someone subscribes to a service which publishes some 100 MB

File: iceoryx2-cli/iox2-service/src/command_publish.rs (hunk 3 above)
Comment (elfenpiff) [reply iceoryx2-cli/iox2-service/src/command_publish.rs]: Adjusted

File: iceoryx2-cli/iox2-service/src/command_publish.rs (hunk 4 above)
Comment (elfenpiff) [reply iceoryx2-cli/iox2-service/src/command_publish.rs:103]: In my opinion it makes absolutely sense when you send for instance json/yaml/xml/toml files around or to see what data my little chatbot narrator receives.

So whenever the user knows that their service sends text it makes sense - up for the user to decide.

File: iceoryx2-cli/iox2-service/src/command_publish.rs (hunk 6 above)
Comment (elfenpiff) [reply iceoryx2-cli/iox2-service/src/command_publish.rs]: I would decline all of your suggestions. I agree that these are must have features for a solid record and replay system - but this feature is intended for simple debugging purposes and shall support developement. It is comparable to `tcpdump`.

Furthermore, use cases like a changed service name or type name (with the same type layout) maybe useful at this low level.

It is not a use case to store the recordings in an archive and replay them later. Therefore, version number and other meta information is not required.

File: iceoryx2-cli/iox2-service/src/cli.rs (hunk 10 above)
Comment (elfenpiff) [reply iceoryx2-cli/iox2-service/src/cli.rs:231]: Clap supports only unit variants. I will add `0` for infinite repetition.

File: iceoryx2-cli/iox2-service/src/command_publish.rs (hunk 2 above)
Comment (elBoberido) [reply iceoryx2-cli/iox2-service/src/command_publish.rs]: The `0x` was just a nice to have for use cases when one manually specifies the data for the publisher in the CLI

File: iceoryx2-cli/iox2-service/src/command_publish.rs (hunk 4 above)
Comment (elBoberido) [reply iceoryx2-cli/iox2-service/src/command_publish.rs:103]: Right. Totally overlooked that. There is still the issue that these data is not send alone. A string, for example, still has a size component which would result in the same issue as described above.

It doesn't hurt to have this option, though.

File: iceoryx2-cli/iox2-service/src/command_publish.rs (hunk 3 above)
Comment (elfenpiff) [reply iceoryx2-cli/iox2-service/src/command_publish.rs]: I reverted it. It just adds error handling and restricts the usage somehow arbitrarily. If someone wants to send a mixture of messages and something from a recorded file - why not. Maybe you have a record and want to inject something additional to see if your algorithm is robust enough.

It works and I don't see why one should restrict it.

File: iceoryx2-cli/iox2-service/src/command_publish.rs (hunk 6 above)
Comment (elBoberido) [reply iceoryx2-cli/iox2-service/src/command_publish.rs]: Well, `tcpdump` is using `pcap` as data format and it has the features I described above :sweat_smile: 

https://datatracker.ietf.org/doc/id/draft-gharris-opsawg-pcap-00.html

File: iceoryx2-cli/iox2-service/src/cli.rs (hunk 10 above)
Comment (elBoberido) [reply iceoryx2-cli/iox2-service/src/cli.rs:231]: No, please not. Then leave it rather as it is. Surrogates are a bad idea and should not be use.

File: iceoryx2-cli/iox2-service/src/command_subscribe.rs [hunk 11]
```diff
@@ -24,9 +24,11 @@ use std::time::Duration;
 use std::time::Instant;
 
 fn raw_data_to_hex_string(raw_data: &[u8]) -> String {
+    use std::fmt::Write;
+
     let mut ret_val = String::with_capacity(2 * raw_data.len());
```
elBoberido:
```suggestion
    let mut ret_val = String::with_capacity(3 * raw_data.len());
```

it is two characters per byte and a whitespace :)

iceoryx2-cli/iox2-service/src/cli.rs
elBoberido: I know you will hate me for this comment, but please remove this. Using surrogates is an anti feature and makes it harder to solve the problem later in a proper way.

iceoryx2-userland/record-and-replay/src/testing.rs
```diff
@@ -0,0 +1,23 @@
...
+// which is available at https://opensource.org/licenses/MIT.
+//
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use crate::{
+    record::RawRecord,
+    recorder::{Recorder, RecorderWriteError},
+};
+
+pub fn recorder_write_unchecked(
```
elBoberido: The `_unchecked` screams for `unsafe`. Can you describe what this function does and why it is not unsafe?

iceoryx2-cli/iox2-service/src/cli.rs
```diff
@@ -142,6 +143,283 @@ pub struct ListenOptions {
...
+}
+
+#[derive(Parser)]
+pub struct PublishOptions {
+    #[clap(help = "Name of the service which shall the message be sent to.")]
+    pub service: String,
+    #[clap(
+        short,
+        long,
+        default_value = "shell_node",
```
elBoberido: Food for thought. Shall we prefix all iceoryx provided nodes with iox2?

Also, what do you think giving the user more context when observing this node in the introspection, e.g. iox2-cli-service-publisher?

iceoryx2-cli/iox2-service/src/cli.rs
```diff
@@ -142,6 +143,283 @@ pub struct ListenOptions {
...
+        help = "Defines the node name of the subscriber endpoint."
+    )]
+    pub node_name: String,
+
+    #[clap(
+        short,
+        long,
+        default_value = "HEX",
```
elBoberido: Is this correct? It's HUMANREADABLE for the publisher

iceoryx2-userland/record-and-replay/tests/hex_conversion_tests.rs
```diff
@@ -0,0 +1,51 @@
...
+    use iceoryx2_userland_record_and_replay::hex_conversion::{
+        bytes_to_hex_string, hex_string_to_bytes, HexToBytesConversionError,
+    };
+
+    #[test]
+    fn back_and_forth_conversion_works() {
+        let sut = "don't stop me now";
+        let hex_sut = bytes_to_hex_string(sut.as_bytes());
+
+        let original_sut = hex_string_to_bytes(&hex_sut).unwrap();
```
elBoberido: Sorry, couldn't resist. This is more of a roundtrip_sut than an original_sut ;)

iceoryx2-userland/record-and-replay/tests/recorder_tests.rs
```diff
@@ -0,0 +1,81 @@
...
+        let result = recorder.write(RawRecord {
+            timestamp: Duration::ZERO,
+            system_header: &[0u8; 8],
+            user_header: &[0u8; 0],
+            payload: &[0u8; 8],
+        });
+
+        assert_that!(result, is_ok);
+
+        File::remove(&file_name).unwrap();
```
elBoberido, line 79: Just to prevent the file not being deleted when the test fails
```suggestion
        File::remove(&file_name).unwrap();
        
        assert_that!(result, is_ok);
```

Btw. would it make sense to also check the content of the file?

iceoryx2-cli/iox2-service/src/command_replay.rs
```diff
@@ -0,0 +1,98 @@
...
+                    );
+                }
+                sample.assume_init()
+            };
+
+            let elapsed = start.elapsed().as_millis() as f64 * options.time_factor as f64;
+            let timestamp = data.timestamp.as_millis() as f64 * options.time_factor as f64;
+            if elapsed < timestamp {
+                std::thread::sleep(Duration::from_millis((timestamp - elapsed) as u64));
+            }
```
elBoberido, line 85: Not for now but maybe a `good first issue`. Would this be something the `WaitSet` with the timer trigger could be used?

iceoryx2-userland/record-and-replay/src/record.rs
```diff
@@ -0,0 +1,292 @@
...
+        Ok(())
+    }
+
+    pub(crate) fn read(self, file: &File) -> Result<Option<Record>, ReplayerOpenError> {
+        let msg = "Unable to read next record";
+        match self.data_representation {
+            DataRepresentation::HumanReadable => {
+                let mut timestamp = None;
+                let mut system_header = None;
+                let mut header = None;
```
elBoberido: Have you considered splitting this function up. It's quite large.

iceoryx2-userland/record-and-replay/tests/recorder_tests.rs
elBoberido: Okay, after reviewing the last file, ignore the question regarding the content ... and also the comment regarding moving the assert below the file removal. The same issue exists in other tests but since it is a temp file with a random name, it should not be an issue and does not warrant spending time to refactor the tests.

elBoberido (commented):
Finished the review. Now new nitpicks. I'll try to run the CLI tool tomorrow.

iceoryx2-cli/iox2-service/src/command_replay.rs
elfenpiff: Actually, we do not yet have this kind of functionality in our WaitSet. It is possible to get woken up with a specific interval but here, in theory, every waiting time could be different.
So we would need to add with every cycle a new wakeup timer and remove the old one.

So I wouldn't use the waitset here.

PR #902, opened 2025-07-29
[#900] Removed is_initialized field from MetaVec

### LINKED ISSUE #900: Remove MetaVec::is_initialized field
I think the `MetaVec::is_initialized` field is superfluous. As far as I can see, it's only ever set to `false` in `RelocatableVec::new_uninit()` to indicate that `self.data_ptr` isn't yet initialized (i.e., has `distance == 0`). But then why not check whether `self.data_ptr.distance == 0` instead?

Another problem with this field is that it's an `IoxAtomicBool`, even though that's unnecessary: All writes to `is_initialized` are done through a `&mut` reference. Writes via mutable references never have to be atomic, because by Rust's rules, a `&mut` reference guarantees

1. that no other reference to this value is accessible, and
2. that memory consistency before and after the lifetime of the `&mut` reference is handled by other means.

See the [`AtomicBool::get_mut()`](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicBool.html#method.get_mut) method for reference, which takes a `&mut self` and returns a `&mut bool` – no atomic accesses required. This means that `is_initialized: IoxAtomicBool` could be replaced with `...: bool`, all loads and stores replaced by regular reads and writes, and the semantics would be the same.

A third problem is that the memory ordering is wrong for what `is_initialized` is ostensibly supposed to do. All loads and stores are `Relaxed`, so no memory consistency with the initialization of `self.data_ptr` is introduced. Instead, the store in `init()` should be `Release`, and the loads in `drop()` and `verify_init()` should be `Acquire`.

To summarize:

1. `MetaVec::is_initialized` uses the wrong atomic memory ordering for loads and stores.
2. It doesn't have to be atomic at all, a regular `bool` would suffice.
3. It isn't needed anyway and could be replaced by a check for `self.data_ptr.distance == 0` (or an accessor to that effect).

awillenbuecher-xq-tec:
## Notes for Reviewer

The `MetaVec::is_initialized` field is redundant with `RelocatablePointer::distance` for `RelocatableVec<T>`, and always `true` for `Vec<T>`. This PR removes the field and replaces it with a new method on `PointerTrait`.

* [ ] Assign PR to reviewer

## References

Closes #900


iceoryx2-bb/elementary-traits/src/pointer_trait.rs
```diff
@@ -31,4 +31,10 @@ pub trait PointerTrait<T> {
     ///  * Do not call this method when the pointer contains a null pointer.
     ///
     unsafe fn as_mut_ptr(&mut self) -> *mut T;
+
+    /// Indicates whether the pointer has been initialized.
+    ///
+    /// *Note:* This method should not be used when the pointer can be initialized to
+    /// point to itself; it is allowed to report false negatives in this case.
+    fn is_initialized(&self) -> bool;
```
elfenpiff: Since this is effectively used only by the `RelocatablePointer` I would add here a default implementation that always returns `true` and is only specialized for the `RelocatablePointer` and returns there the state of initialization.

So change this here to:

```rust
fn is_initialized(&self) -> bool {
    true
}
```

iceoryx2-bb/elementary-traits/src/owning_pointer.rs
```diff
@@ -67,6 +67,11 @@ impl<T> PointerTrait<T> for OwningPointer<T> {
     unsafe fn as_mut_ptr(&mut self) -> *mut T {
         self.ptr
     }
+
+    fn is_initialized(&self) -> bool {
```
elfenpiff: This can be removed when the trait as a default impl.

elfenpiff (commented):
@awillenbuecher-xq-tec Thanks for the contribution!

I would provide a default implementation in the pointer trait that is only specialized by the relocatable ptr.

iceoryx2-bb/elementary-traits/src/pointer_trait.rs
awillenbuecher-xq-tec: I've added a default implementation.

iceoryx2-bb/elementary-traits/src/owning_pointer.rs
awillenbuecher-xq-tec: Done

elfenpiff (approved):

Issue #865, opened 2025-07-10
Support `nostd` builds


orecham:
## Brief Description

Support use of `iceoryx2` on `nostd` targets

orecham:
Related:

* https://github.com/eclipse-iceoryx/iceoryx2/issues/731
* https://github.com/eclipse-iceoryx/iceoryx2/issues/853

PR #907, opened 2025-07-29
[#865] Add bincode implementation of the `Serialize` concept

orecham:
## Notes for Reviewer

Adds a `bincode` implementation of the `Serialize` concept for use in `nostd` deployments.

* [ ] Assign PR to reviewer

## References

Closes #


Cargo.toml
```diff
@@ -86,6 +86,7 @@ iceoryx2-tunnels-end-to-end-testing = { version = "0.6.1", path = "iceoryx2-tunn
 iceoryx2-tunnels-zenoh = { version = "0.6.1", path = "iceoryx2-tunnels/zenoh" }
 
 anyhow = { version = "1.0.86" }
+bincode = { version = "2.0.1", default-features = false, features = ["alloc","serde"] }
```
elfenpiff: According to their readme, bincode has a minimum required rust version of 1.85 (see: https://github.com/bincode-org/bincode?tab=readme-ov-file#what-is-bincodes-msrv-minimum-supported-rust-version), iceoryx2 has currently 1.81.

From my side it is not a problem to increment the MSRV on our side but there I might user out there which have a hard requirement on a lower version. @elBoberido do you have any insights here?

elBoberido: VxWorks currently only support Rust 1.82. But AFAIK the serializer can be changed for VxWorks via the `recommended.rs` file, so it should not be a big deal for VxWorks.

But I'm not sure what else will be affected with a larger MSRV. Would it be possible to use the bincode crate but not update our MSRV? If bincode is either hidden behind a feature flag, e.g. `nostd`, then it could work.

Generally, especially for embedded use cases it would be good to support an older rust compiler.

There is also a "nuklear" option. I just tried to build bincode with Rust 1.81 and it builds fine. We could carry a small patch to downgrade the MSRV for bincode. It seems the main reason to go with 1.85 was to be able to use the Rust Edition 2024 at some point.

elfenpiff: I would vote against a feature flag for now. I would prefer if we would take the time and address the feature flags in a separate issue cleanly.

My idea is, that iceoryx2 is by default stripped so much, that it has nearly no dependencies (or only dependencies that are `no_std`) and with added feature flags, dependencies are added.

Do you want human readable files, use the toml feature flag. Want to have bincode serialization, fine use this flag. Want to have CLI tooling with fancy colors, add the ascii-arts flag ... something like that.

elBoberido: @orecham @elfenpiff what do you think about using [postcard](https://crates.io/crates/postcard) instead of `bincode`?

I think it supports older compiler, has only serde as hard dependency and also meant for embedded nostd use cases.

orecham: @elBoberido I honestly don't mind either way, I only went with `bincode` because it was your previous suggestion.

That said, I have already tested `bincode` for some other targets.

elBoberido: @orecham yeah, I also use bincode in a personal project but with the requirement of Rust 1.85 it might be better to use an alternative.

orecham: The version of `bincode` that can be built with MSRV 1.8.1 is v1.3.1, but this version is not `nostd` (like the `2.x` version is).

I don't like the idea of hiding it behind a feature flag (worse UX), or patching the upstream version just to use it (annoying to maintain).

If we want to resist bumping our MSRV to 1.8.5, I would prefer to just switch to `postcard`, but I would wait to test it on the actual targets that needed the change in serializer before merging. My preferred option of the two is to bump the MSRV.

@elBoberido @elfenpiff Thoughts?

orecham: @elBoberido @elfenpiff I switched it to `postcard`.

elBoberido: @orecham once we we more reasons to update the compiler, we can also add a bincode implementation. This is interesting if someone already uses bincode in their project and wants to reduce the number of dependencies.

Issue #663, opened 2025-03-28
Provide Official Yocto Recipe


elfenpiff:
iceoryx2 is already used in yocto. Instead of that every user writes its own recipe we might should provide an official one and test it in our CI.

Some solution seem to exist here: https://github.com/user-attachments/files/19497699/iceoryx2.zip (thanks to @zachary0101)

See also this part of the discussion: https://github.com/eclipse-iceoryx/iceoryx2/issues/660#issuecomment-2759847346

PR #912, opened 2025-07-30
[#663] Recreate the cmake patch, used in the yocto layer

vansweej:
cmake variable BUILD_SHARED_LIBS is created. When ON shared libraries will be build and deployed, when OFF only static libraries will be build and deployed.
Default to ON, so without explicit setting this cmake variable, no change in build behavior.


doc/release-notes/iceoryx2-unreleased.md
```diff
@@ -122,6 +122,9 @@
     please refer to `config/README.md` on how to generate a default config file.
   [#831](https://github.com/eclipse-iceoryx/iceoryx2/issues/831)
 
+* Recreate the cmake patch, used in the yocto layer
+  [#663](https://github.com/eclipse-iceoryx/iceoryx2/issues/663)
+
```
elBoberido: I think this can be moved to the features section, e.g. line 20. It introduces a new option which is ON by default, so does not break existing builds.

I think when it is moved to the other section, the markdown linter will also stop complaining.

elBoberido: @vansweej there is a merge confligt from another entry in the changelog.

Could you please rebase.

Btw, we try to reduce merge conflicts in this file by sorting the entries by the issue number. If you move the entry up to line 20, there is a reduced risk that another PR will interfere with yours.

vansweej: we crossed. I synced my fork already

elBoberido: Okay, no problem. We just have to merge this PR quickly :)

elBoberido:
@vansweej closing this PR since the refactoring of the recipes in https://github.com/eclipse-iceoryx/meta-yocto-iceoryx2/pull/7 does not require the patch anymore. Thanks for bringing this up.

PR #946, opened 2025-08-13
[#936] CMake build for cxx container library

### LINKED ISSUE #936: C++ version of iceoryx-bb to house base library facilities
## Brief feature description

Establish a C++ build for iceoryx2-bb to hold base library facilities such as basic vocabulary types (optional, expected, variant) and containers (shared memory vector and string). The base library will build as its own library, independent of the current ffi project, but should be fully integrated with the main iceoryx build the same way that the ffi currently is. The build should support both CMake and Bazel.

## Tasks

- [x] cmake build
- [x] cmake install
- [x] bazel integration

ComicSansMS:
## Notes for Reviewer
This sets up the CMake build infrastructure for the C++ container library. Other build systems are still missing at the moment.

I also added CTest support for the ffi unit tests with this, as that only required a couple of lines of code, and I changed the method for setting the C++ standard version to use the more robust `target_compile_features`. There is further potential for cleaning up these scripts if we decide to bump the minimum required version in the future:

- 3.23 introduces file sets for `target_sources`
- 3.24 introduces the `COMPILE_WARNING_AS_ERROR` target property

There is some duplication necessary, most visible in the duplicated `googletest.cmake` file. This is due to the fact that iceoryx currently maintains ffi and its tests as a self-contained CMake projects. According to @elfenpiff this is intentional and should be supported going forward. As a consequence, supporting scripts like `googletest.cmake` have either be duplicated or placed outside the source tree of the project. Both of these are not ideal, but given the alternatives I find the former to be far more preferable, so I went with this approach here. Any input on the issue is highly appreciated.

* [x] Assign PR to reviewer

## References

Closes #936

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -133,10 +133,17 @@ if(WARNING_AS_ERROR)
     endif()
 endif()
 
+if(BUILD_TESTING)
+    enable_testing()
+endif()
+
 if(SANITIZERS)
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fsanitize=address -fsanitize=undefined")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address -fsanitize=undefined")
 endif()
+
+# C++ containers
+add_subdirectory(iceoryx2-bb/cxx)
 
 # C binding
 add_subdirectory(iceoryx2-ffi/c)
--- a/iceoryx2-bb/cxx/CMakeLists.txt
+++ b/iceoryx2-bb/cxx/CMakeLists.txt
@@ -12,25 +12,29 @@
 
 cmake_minimum_required(VERSION 3.22)
 
-project(iceoryx2-bb VERSION ${IOX2_VERSION_STRING} LANGUAGES CXX)
+project(iceoryx2-bb-cxx VERSION ${IOX2_VERSION_STRING} LANGUAGES CXX)
 
 set(PREFIX iceoryx2/v${CMAKE_PROJECT_VERSION})
 
-# include only lib -> includes are installed only once despite being used in static lib as well as shared lib
+add_library(iceoryx2-bb-containers-cxx INTERFACE)
+add_library(iceoryx2-bb-cxx::iceoryx2-bb-containers-cxx ALIAS iceoryx2-bb-containers-cxx)
 
-add_library(iceoryx2-bb-containers INTERFACE)
-add_library(iceoryx2-cxx::iceoryx2-bb-containers ALIAS iceoryx2-bb-containers)
-
-target_include_directories(iceoryx2-bb-containers
+target_include_directories(iceoryx2-bb-containers-cxx
     INTERFACE
     $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
     $<INSTALL_INTERFACE:include/${PREFIX}>
 )
 
-target_compile_features(iceoryx2-bb-containers INTERFACE cxx_std_14)
+target_compile_features(iceoryx2-bb-containers-cxx INTERFACE cxx_std_14)
 
 # include install setup
-# @todo
+# find_package from source tree
+set(${PROJECT_NAME}_DIR ${PROJECT_SOURCE_DIR}/cmake
+    CACHE FILEPATH
+    "${PROJECT_NAME}Config.cmake to make find_package(${PROJECT_NAME}) work in source tree!"
+    FORCE
+)
+# @todo - install, export and packaging
 
 # add tests
 
--- a/iceoryx2-bb/cxx/CMakeLists.txt
+++ b/iceoryx2-bb/cxx/CMakeLists.txt
@@ -1,0 +1,40 @@
+# Copyright (c) 2024 Contributors to the Eclipse Foundation
+#
+# See the NOTICE file(s) distributed with this work for additional
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache Software License 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0, or the MIT license
+# which is available at https://opensource.org/licenses/MIT.
+#
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+
+cmake_minimum_required(VERSION 3.22)
+
+project(iceoryx2-bb VERSION ${IOX2_VERSION_STRING} LANGUAGES CXX)
+
+set(PREFIX iceoryx2/v${CMAKE_PROJECT_VERSION})
+
+# include only lib -> includes are installed only once despite being used in static lib as well as shared lib
+
+add_library(iceoryx2-bb-containers INTERFACE)
+add_library(iceoryx2-cxx::iceoryx2-bb-containers ALIAS iceoryx2-bb-containers)
+
+target_include_directories(iceoryx2-bb-containers
+    INTERFACE
+    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
+    $<INSTALL_INTERFACE:include/${PREFIX}>
+)
+
+target_compile_features(iceoryx2-bb-containers INTERFACE cxx_std_14)
+
+# include install setup
+# @todo
+
+# add tests
+
+if(${BUILD_TESTING})
+    enable_testing()
+    add_subdirectory(tests)
+endif()
--- a/iceoryx2-bb/cxx/tests/CMakeLists.txt
+++ b/iceoryx2-bb/cxx/tests/CMakeLists.txt
@@ -16,7 +16,7 @@ project(iceoryx2-bb-cxx-tests VERSION ${IOX2_VERSION_STRING} LANGUAGES CXX)
 
 include(cmake/googletest.cmake)
 
-find_package(iceoryx2-cxx REQUIRED)
+find_package(iceoryx2-bb-cxx REQUIRED)
 
 add_executable(iceoryx2-bb-containers-tests ${TEST_FILES})
 target_sources(iceoryx2-bb-containers-tests
@@ -24,7 +24,7 @@ target_sources(iceoryx2-bb-containers-tests
     ${PROJECT_SOURCE_DIR}/src/main.cpp
 )
 
-target_link_libraries(iceoryx2-bb-containers-tests iceoryx2-cxx::iceoryx2-bb-containers GTest::gtest GTest::gmock)
+target_link_libraries(iceoryx2-bb-containers-tests iceoryx2-bb-cxx::iceoryx2-bb-containers-cxx GTest::gtest GTest::gmock)
 target_compile_options(iceoryx2-bb-containers-tests PRIVATE $<$<CXX_COMPILER_ID:MSVC>:/bigobj>)
 target_compile_features(iceoryx2-bb-containers-tests PRIVATE cxx_std_14)
 
--- a/iceoryx2-bb/cxx/tests/CMakeLists.txt
+++ b/iceoryx2-bb/cxx/tests/CMakeLists.txt
@@ -12,21 +12,23 @@
 
 cmake_minimum_required(VERSION 3.22)
 
-project(iceoryx2-cxx-tests VERSION ${IOX2_VERSION_STRING} LANGUAGES CXX)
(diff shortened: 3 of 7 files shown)


CMakeLists.txt
```diff
@@ -133,11 +133,18 @@ if(WARNING_AS_ERROR)
     endif()
 endif()
 
+if(BUILD_TESTING)
+    enable_testing()
+endif()
```
elBoberido, line 138: Nice. We used `BUILD_TEST` in iceoryx classic and I changed the flag for iceoryx2 to be more in line with the cmake ecosystem. It seems I forgot about this. Btw, we are in the process to make the C and C++ bindings build standalone for packagers. Do you know if it is a problem to cal `enable_testing()` multple times?

iceoryx2-bb/cxx/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+
+cmake_minimum_required(VERSION 3.22)
+
+project(iceoryx2-bb VERSION ${IOX2_VERSION_STRING} LANGUAGES CXX)
+
+set(PREFIX iceoryx2/v${CMAKE_PROJECT_VERSION})
+
+# include only lib -> includes are installed only once despite being used in static lib as well as shared lib
```
elBoberido: I think this one can be removed. The following code crates the final library directly and I guess this will also remain a header only lib.

We used this trick in the C++ bindings in order to create static as well as shared libraries with the same build and did not want to add the header directly to one of the binary targets but have a third, header only lib, which the binary targets can depend on.

iceoryx2-bb/cxx/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+
+add_library(iceoryx2-bb-containers INTERFACE)
+add_library(iceoryx2-cxx::iceoryx2-bb-containers ALIAS iceoryx2-bb-containers)
+
+target_include_directories(iceoryx2-bb-containers
+    INTERFACE
+    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
```
elBoberido, line 24: Out of curiosity. We used always `CMAKE_CURRENT_SOURCE_DIR` but it's good to get some feedback from outside of our bubble. Has `PROJECT_SOURCE_DIR` advantages over `CMAKE_CURRENT_SOURCE_DIR`? I'm happy to change it everywhere else to `PROJECT_SOURCE_DIR`

iceoryx2-bb/cxx/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+
+target_compile_features(iceoryx2-bb-containers INTERFACE cxx_std_14)
+
+# include install setup
+# @todo
+
+# add tests
+
+if(${BUILD_TESTING})
+    enable_testing()
```
elBoberido, line 42: Okay, I guess this makes my question from above obsolete.

iceoryx2-bb/cxx/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+# This program and the accompanying materials are made available under the
+# terms of the Apache Software License 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0, or the MIT license
+# which is available at https://opensource.org/licenses/MIT.
+#
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+
+cmake_minimum_required(VERSION 3.22)
+
+project(iceoryx2-bb VERSION ${IOX2_VERSION_STRING} LANGUAGES CXX)
```
elBoberido: Food for thought. Is there a chance that we offer some kind of C API for this container? In this case it would maybe make sense to use `iceoryx2-bb-cxx` or maybe `iceoryx2-cxx-bb`, which would have the nice side effect to have everything that is C++ related have the `iceoryx2-cxx` prefix

iceoryx2-bb/cxx/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+
+set(PREFIX iceoryx2/v${CMAKE_PROJECT_VERSION})
+
+# include only lib -> includes are installed only once despite being used in static lib as well as shared lib
+
+add_library(iceoryx2-bb-containers INTERFACE)
+add_library(iceoryx2-cxx::iceoryx2-bb-containers ALIAS iceoryx2-bb-containers)
```
elBoberido: The namespace should probably be the same like the project name, else it will be confusing since the user would need to do a `find_package(iceoryx2-bb)` but need to link towards `iceoryx2-cxx::*`. If we stick with just `iceoryx2-bb` as package name, I would suggest to use `iceoryx2-bb::container`.

iceoryx2-bb/cxx/tests/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+
+cmake_minimum_required(VERSION 3.22)
+
+project(iceoryx2-bb-cxx-tests VERSION ${IOX2_VERSION_STRING} LANGUAGES CXX)
+
+include(cmake/googletest.cmake)
+
+find_package(iceoryx2-cxx REQUIRED)
```
elBoberido: I guess this is just a typo and you meant this
```suggestion
find_package(iceoryx2-bb REQUIRED)
```

elBoberido (commented):
Looks good. Just some questions about naming

iceoryx2-bb/cxx/tests/cmake/googletest.cmake
```diff
@@ -0,0 +1,26 @@
+# Copyright (c) 2024 Contributors to the Eclipse Foundation
```
elfenpiff, line 1: I thought we create the following scheme:
* `iceoryx2-bb/container/cxx/`
* `iceoryx2-bb/container/rust/`
* `iceoryx2-bb/container/python/` (in the future)

All other crates most likely will not be ported to any other language.

iceoryx2-bb/cxx/CMakeLists.txt
elBoberido: ... well, calling it just `container` would require that the library is also called just `containers`. This should not be a big issue since it it a header only lib and if it becomes a binary lib at some point, we can always use `OUTPUT_NAME` to rename the binary to `iceoryx2_bb_containers.so`

iceoryx2-bb/cxx/tests/CMakeLists.txt
elfenpiff: Since we are only porting the containers to C++ I would call it `iceoryx2-bb-container-cxx` and add the `cxx` suffix at the end so that it is equal to `iceoryx2-bb-container` crate for the first part. It makes it clear that it is the C++ version of the crate/library.

If we every have other versions of that crate we can then follow the name scheme `iceoryx2-bb-container-LANGUAGE`

iceoryx2-bb/cxx/CMakeLists.txt
elfenpiff: Completely agree but please use the name `iceoryx2-bb-container-cxx`

iceoryx2-ffi/cxx/tests/CMakeLists.txt
```diff
@@ -24,12 +24,14 @@ add_executable(${PROJECT_NAME} ${TEST_FILES})
 
 target_link_libraries(${PROJECT_NAME} iceoryx2-cxx::static-lib-cxx GTest::gtest GTest::gmock)
 target_compile_options(${PROJECT_NAME} PRIVATE $<$<CXX_COMPILER_ID:MSVC>:/bigobj>)
+target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_17)
```
elfenpiff, line 27: We have C++14 as a strict requirement.

iceoryx2-ffi/cxx/CMakeLists.txt
```diff
@@ -100,10 +100,9 @@ target_link_libraries(iceoryx2-cxx-object-lib
 add_library(static-lib-cxx STATIC $<TARGET_OBJECTS:iceoryx2-cxx-object-lib>)
 add_library(iceoryx2-cxx::static-lib-cxx ALIAS static-lib-cxx)
 
+target_compile_features(static-lib-cxx PUBLIC cxx_std_17)
```
elfenpiff, line 103: Please use `cxx_std_14` here and everywhere else.

iceoryx2-ffi/cxx/tests/CMakeLists.txt
elBoberido: Sure, but it's still using iceoryx_hoofs, which are at the moment C++17

iceoryx2-ffi/cxx/CMakeLists.txt
elBoberido: Not possible at the moment. See above.

iceoryx2-bb/cxx/CMakeLists.txt
elBoberido: You would call the namespace `iceoryx2-bb-container-cxx`?

I would suggest to not got this route and just use `iceoryx2-bb` as namespace and `container-cxx` as package name. The user would then just have to do a `find_package(iceoryx2-bb) ` and link towards `iceoryx2-bb::container-cxx`.

This means there would also be just one cmake package for `iceoryx2-bb` and not multiple ones. There would be just additional libraries in the `iceoryx2-bb` package.

CMakeLists.txt
ComicSansMS: It should be fine to invoke it multiple times. Basically, you will want one invocation for every directory level where you want `ctest` to work, so usually the `CMakeLists.txt` with a `project` command will also end up with an `enable_testing`.

iceoryx2-bb/cxx/CMakeLists.txt
ComicSansMS: `CMAKE_CURRENT_SOURCE_DIR` is always the very top of the source tree that CMake is aware of (so the directory passed to `cmake -S ...`), `PROJECT_SOURCE_DIR` is the directory where the last `CMakeLists.txt` with a `project` command is located. If you have nested projects like iceoryx, you almost alwas want `PROJECT_SOURCE_DIR`, otherwise you will end up with a different path when building the projects standalone. In general the legitimate uses for `CMAKE_CURRENT_SOURCE_DIR` are very few.

iceoryx2-bb/cxx/CMakeLists.txt
ComicSansMS: Agreed a `-cxx` does make sense in the name here, even if we don't add other languages later on.

iceoryx2-ffi/cxx/tests/CMakeLists.txt
ComicSansMS: It should still work with 14, as this CMake treats this as a minimum requirement.

iceoryx2-bb/cxx/tests/cmake/googletest.cmake
ComicSansMS: I don't have strong feelings about the hierarchy here. I mostly tried to mirror the layout of `iceoryx2-ffi` here, but that may not be a good idea.

iceoryx2-bb/cxx/tests/CMakeLists.txt
ComicSansMS: There is two possibilities here, always have a prefix `iceoryx2-LANG-...` or always have the language as a suffix `iceoryx2-...-LANG`.

The ffi targets seem to favor the prefix variant at the moment.

iceoryx2-bb/cxx/CMakeLists.txt
ComicSansMS: I settled on `iceoryx2-bb-cxx::iceoryx2-bb-containers-cxx` for now. It's somewhat verbose, but it does seem sensible to me?

iceoryx2-bb/cxx/tests/CMakeLists.txt
elfenpiff: I prefer the suffix variant since then the closely related things are close together in a literal sort. Like in my IDE file explorer or with ls.

elfenpiff:
@ComicSansMS for me it is fine to merge but since @elBoberido understands our cmake thingies better I would leave the approval to him

Issue #939, opened 2025-08-11
C++ `optional` vocabulary type


ComicSansMS:
## Brief feature description

Provide as part of the C++ base library an implementation of a vocabulary type that can act as a drop-in replacement for `std::optional`. This targets deployments where no safety certified C++17 standard library is available to provide optional.

elfenpiff:
One side constraint. The `optional` must compile with C++14.

elBoberido:
@ComicSansMS can this be closed?

PR #949, opened 2025-08-13
C++ Optional type for iceoryx 2 container library

ComicSansMS:
## Notes for Reviewer
Initial implementation of iceoryx2 version of C++17's `std::optional`. The goal here is to provide a drop-in replacement for the standard library facility, so the interface follows the standard library and does not divert from the standard interface for safety.

Open points:
* Some member functions are still missing, in particular the emplacement API, the iterator interface and the monadic interface. These will be added in a later version.
* The current implementation does not handle certain cases of conditional explicit and conditional noexcept correctly. It also does not forward triviality of special member functions of the wrapped type correctly. This will be added in a later version.
* The current implementation calls `std::abort` as a placeholder for a currently unspecified precondition violation handling mechanism. This will be replaced in the future by a different mechanism, but I don't know yet what it will look like. These cases are also not covered by unit tests at the moment.
* This PR depends on the changes from #946 and replicates those commits to be buildable.
* The entire container library is considered unstable and subject to breaking changes at the moment and therefore not currently packaged as part of an install of iceoryx2.

* [x] Assign PR to reviewer

## References

Relates to #939


iceoryx2-bb/cxx/include/iox2/container/optional.hpp
```diff
@@ -0,0 +1,285 @@
...
+        :OptionalValueHolder()
+        {
+            if (!rhs.isEmpty) { set(std::move(rhs.value)); }
+        }
+
+#       if __cplusplus >= 202002L
+        constexpr
+#       endif
+        ~OptionalValueHolder() {
+            if (!isEmpty) {
```
elfenpiff: Why not call `reset()` here directly. Simplifies the code a bit.

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
```diff
@@ -0,0 +1,285 @@
...
+                if (rhs.isEmpty) {
+                    reset();
+                } else {
+                    set(std::move(rhs.value));
+                }
+            }
+            return *this;
+        }
+
+        constexpr void set(T const& v) {
```
elfenpiff: Instead of having two implementations, couldn't we use here perfect forwarding and write the code like this:

```cxx
template<typename TT>
constexpr void set(TT&& v) {
  // ...
  new(&value) T{std::forward<TT>(v)};
  // ...
}
```

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
```diff
@@ -0,0 +1,285 @@
...
+        union {
+            char null;
+            typename std::remove_cv<T>::type value;
+        };
+        bool isEmpty;
+
+        constexpr OptionalValueHolder() noexcept: null(), isEmpty(true) {}
+        constexpr OptionalValueHolder(T const& v): value(v), isEmpty(false) {}
+        constexpr OptionalValueHolder(T&& v): value(std::move(v)), isEmpty(false) {}
+        constexpr OptionalValueHolder(OptionalValueHolder const& rhs)
```
elfenpiff: A personal preference, not a must: What do you think of decoupling the interface and the implementation? So here you just declare the interface and at the bottom of the file the full implementation follows, like here: https://github.com/eclipse-iceoryx/iceoryx2/blob/main/iceoryx2-ffi/cxx/include/iox2/sample.hpp

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
```diff
@@ -0,0 +1,285 @@
...
+        if (m_value.isEmpty) {
+            return nullptr;
+        } else {
+            return &m_value.value;
+        }
+    }
+
+    constexpr const T& operator*() const & noexcept {
+        if (m_value.isEmpty) {
+            std::abort();
```
elfenpiff, line 254: Could we create a tiny abstraction around it, maybe an `iox2_abort()`. The idea is, that when we already use this all over the code base, we do not need to refactor it later when we have to provide a fatal error handler that is customizable.

For now it would be just a `void iox2_abort() { std::abort; }`.

This should then be adjusted everywhere here in the code. We should also add a short entry here: https://github.com/eclipse-iceoryx/iceoryx2/blob/main/FAQ_ICEORYX_DEVS.md just stating:
1. fatal error handling use `iox2_abort`
2. non-fatal error handling return `expected`

Maybe in a separate C++ section.

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
```diff
@@ -0,0 +1,867 @@
...
+    
+    Observable& operator=(Observable&& rhs) {
+        ++g_ObservableWasMoveAssigned;
+        id = rhs.id;
+        return *this;
+    }
+};
+}
+
+TEST(Optional, DefaultConstructor) {
```
elfenpiff: The tests look good and complete, but I have some stylistic remarks, even though it is not so important for the optional, but just for consistency.

1. Please use lower_snake_case_names for all test functions. This would then be nicely in line with all Rust, C-, C++-, Python- ffi tests.
2. Write the test names like little requirements. It should contain the
  * What does the test do
  * What is the expected outcome

So, in this cas,e for example
* `default_constructed_optional_is_empty`
* or in one of the test cases below `optional_constructed_by_value_contains_value`

For those tests, it seems a bit silly, but it helps reviewing the test code. I review always in two iterations:
1. I check just the names, see if they behavior is in line with my expections on the construct and check also for completeness of the tests from feature/behavior point of view
2. I review the test itself and check if the name of the test itself matches the test implementation.

For the test implementation, we try to follow the arrange, act, assert pattern - when it makes sense, see: https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/

Also, a good read is this: https://github.com/eclipse-iceoryx/iceoryx/blob/main/doc/website/concepts/best-practice-for-testing.md @elBoberido wrote it several years ago and describes very well our testing best practices.

We should copy this document to iceoryx2 and link it in the dev faq ...

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
```diff
@@ -0,0 +1,867 @@
...
+    // [optional.ctor] / 2
+    {
+        iox2::container::Optional<int> o;
+        ASSERT_FALSE(o.has_value());
+    }
+    // [optional.ctor] / 3
+    {
```
elfenpiff: Please split up every test that has more than one scope into multiple tests.

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
```diff
@@ -0,0 +1,867 @@
...
+        iox2::container::Optional<int> empty;
+        ASSERT_TRUE(o.has_value());
+        ASSERT_TRUE(!empty.has_value());
+        o = empty;
+        ASSERT_TRUE(!o.has_value());
+        ASSERT_TRUE(!empty.has_value());
+    }
+    {
+        iox2::container::Optional<testing::Observable> o{testing::Observable{}};
+        testing::g_ObservableWasInitialized = 0;
```
elfenpiff: Please introduce an `Observable::reset()` function that sets every value to 0?

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
```diff
@@ -0,0 +1,867 @@
...
+        ASSERT_TRUE(!empty.has_value());
+        ASSERT_EQ(testing::g_ObservableWasInitialized, 0);
+        ASSERT_EQ(testing::g_ObservableWasCopied, 0);
+        ASSERT_EQ(testing::g_ObservableWasCopyAssigned, 0);
+        ASSERT_EQ(testing::g_ObservableWasMoved, 0);
+        ASSERT_EQ(testing::g_ObservableWasMoveAssigned, 0);
+        ASSERT_EQ(testing::g_ObservableWasDestructed, 1);
+        testing::g_ObservableWasDestructed = 0;
+    }
+    ASSERT_EQ(testing::g_ObservableWasDestructed, 0);
```
elfenpiff: I once read a blog article that says every testable side-effect should be tested, since sometimes weird things happen that we accidentally just overlook.

So in this case, I would introduce several methods for the `Observable` that check, for instance, that the observable was destroyed, and nothing else happened to it. So whenever you call it, it checks if every variable is zero, except the one you expect not to be 0. After the call, the `Observerable` resets itself automatically.

For instance, due to a mistake, the observable could have been moved again or copied, but not destroyed.

So in the Observable, we would have this:
```cxx
class Observable {
  public:
    auto was_destructed(uint64_t times = 1) -> bool {
      auto result = g_ObserableWasDestructed == times;
      result &= g_ObservableWasInitialized == 0;
      result &= g_ObservableWasCopied == 0;
      // ...
      reset();
      return result;
    }
};
```

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
```diff
@@ -0,0 +1,867 @@
...
+#include <gtest/gtest.h>
+
+namespace {
+namespace testing {
+int g_ObservableWasInitialized = 0;
+int g_ObservableWasCopied = 0;
+int g_ObservableWasCopyAssigned = 0;
+int g_ObservableWasMoved = 0;
+int g_ObservableWasMoveAssigned = 0;
+int g_ObservableWasDestructed = 0;
```
elfenpiff: Could we also add here an instance counter? that is incremented in every constructor and decremented in the dtor.

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
elfenpiff: Or even better. We introduce a ObservableTestGuard. It has 2 tasks:
1. Provide access to the global variables to check what happened when
2. When it goes out of scope, it expects that the instance counter is 0, not that by any incident we leaked 

So it would look like this:

```cxx
class ObservableTestGuard {
  public:
    ObservableTestGuard() {
      g_Observable_instance_counter = 0;
      // sets also all other variables to zero
    }
    ~ObservableTestGuard() {
      EXPECT_EQ(g_Observable_instance_counter, 0);
    }

    auto was_destructed(uint64_t times = 1) -> bool {
      auto result = g_ObserableWasDestructed == times;
      result &= g_ObservableWasInitialized == 0;
      result &= g_ObservableWasCopied == 0;
      // ...
      reset();
      return result;
    }
};
```

Then the test could be written like this:

```cxx
ObservableTestGuard test_guard;
iox2::container::Optional<testing::Observable> o{testing::Observable{}};
ASSERT_TRUE(test_guard.was_initialized(1));

iox2::container::Optional<testing::Observable> empty;
ASSERT_TRUE(o.has_value());
ASSERT_TRUE(!empty.has_value());

o = empty;

ASSERT_TRUE(!o.has_value());
ASSERT_TRUE(!empty.has_value());

ASSERT_TRUE(test_guard.was_destructed(1));

// here the test_guard goes out of scope and checks that nothing else lives.
```

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
elfenpiff: Shouldn't this be 1?

elfenpiff (commented):
Mainly test comments:
1. Please use lower_snake_case for test names
2. Rewrite test names so that they explain what is being tested and what the outcome shall be
3. Add instance counter to observable and count all living instances
4. Add test guard that resets all values on instantiation and tests when going out-of-scope that living instances == 0
5. Add methods to test if a certain method was called N times and nothing else was happening (g_ObservableCOUNTER == N && g_Observable*** == 0)

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
ComicSansMS: That would work. 
Overload resolution surrounding forwarding references is slightly more complicated than doing it this way, so I usually prefer the split version. It also makes it easier to see which cases one forgot to test in coverage. Downside is code duplication, but for a class like this that is unlikely to change much in the future, that doesn't bother me too much.

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
ComicSansMS: I don't care much.

Having the implementation out of line gets somewhat annoying when there is a lot of metaprogramming in the signatures which will have to be duplicated. Optional will gain some of that with the missing features, but it is not the worst offender in that regard. Other containers might be much worse though.

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
ComicSansMS: Yes, sorry, I forgot to mention this in the review comments. I now added it to the reviewer notes. We will need an abstraction here, but I haven't decided yet what I want that to look like. Consider the `abort` a placeholder for now.

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
ComicSansMS: Thanks for the document, that is a very helpful read.

The naming convention for the tests is also useful to know, thanks!

The idea was to eventually trace each test case to the respective clause in the ISO 14882 document. That way we can show requirements coverage for the ISO standard which should be useful for certification. I started doing that but for the reduced feature set we are shipping now, it is difficult to do the tracing correctly, so I left that out of the PR for now (except for these two leftover comments in the `DefaultConstructor` test case).

I thought I already was following arrange-act-assert. Is the comment here that expected values should always be put in named variables? It is true though that some test cases assert more than one property of a function. I was planning to pull that apart during the requirements tracing phase. There is always a trade-off here between simplicity and number of test cases. I take it from the best practices document that you usually push things further in the direction of more and smaller test cases, which makes total sense.

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
ComicSansMS: I see what you are saying. This would bring us much closer to how most mocking frameworks handle testing. I was deliberately not doing that here so to not overconstrain the tests. I would also argue that this goes against the spirit of AAA, but that is a philosophical debate, which we should probably resolve offline.

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
ComicSansMS: I can add that for ease of understanding. The value of the instance counter would always be the sum of all the constructor counters though (`WasInitialized` + `WasMoved` + `WasCopied`), so it won't assert any properties that are not already covered.

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
ComicSansMS: It could be, but the idea here was to only observe the changes to the counter caused by the operation under test, not to have a valid count of all active instances at every point during execution.

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
elfenpiff: I added the arrange act assert comment only for completeness.

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
ComicSansMS: Done.

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
ComicSansMS: Done, and added an automated leak check to every test.

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
ComicSansMS: Done.

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
elBoberido: He have some abstraction in iceoryx classic -> https://github.com/eclipse-iceoryx/iceoryx/blob/main/iceoryx_hoofs/reporting/include/iox/assertions.hpp

It also contains the machinery to actually test the aborts with gTest. We might need to trim the implementation down, though. At least we should take some inspiration and not start from scratch.

iceoryx2-bb/cxx/include/iox2/container/config.hpp
elBoberido, line 1: In iceoryx classic, we decided to have a flat include structure and place everything into just `iox`, after we had some issues with moving headers around. Shall he have the same structure for iceoryx2?

An option would be to place everything into `iox2/bb` instead of just `iox2`.

iceoryx2-bb/cxx/include/iox2/container/config.hpp
```diff
@@ -0,0 +1,24 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+#ifndef INCLUDE_GUARD_IOX2_CONTAINER_CONFIG_HPP
+#define INCLUDE_GUARD_IOX2_CONTAINER_CONFIG_HPP
+
+
+#define IOX2_CONTAINER_CONFIG_USE_STD_OPTIONAL 0
+
+#define IOX2_CONTAINER_CONFIG_USE_STD_EXPECTED 1
```
elBoberido: The question is if we want to make this a compile time option?

This can be done with converting this file into a `config.hpp.in` similar to this -> https://github.com/eclipse-iceoryx/iceoryx/blob/main/iceoryx_platform/linux/cmake/platform_settings.hpp.in#L22

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
```diff
@@ -0,0 +1,296 @@
...
+#if __cplusplus >= 201703L
+template <class T>
+Optional(T) -> Optional<T>;
+#endif
+
+#endif
+
+#if __cplusplus >= 201703L
+inline constexpr NulloptT nullopt;
+#endif
```
elBoberido: I'm not quite sure why this is only enabled for C++17?

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
```diff
@@ -0,0 +1,296 @@
...
+
+namespace detail {
+/// Internal union implementation for Optional.
+/// @todo Proper handling of cond. explicit, cond. noexcept and triviality of special member functions.
+template <typename T>
+struct OptionalValueHolder {
+    union {
+        char null;
+        typename std::remove_cv<T>::type value;
+    };
```
elBoberido: From my point of view, this is a sane approach but 7 years ago I had a discussion with an automotive guy that was against using `union`. I think the justification he gave was wrong but we had no choice and use something like the `UninitializedArray`.

Since you are are also in favor of the `union`, I guess you also do not see an issue with certification when it is properly encapsulated?

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
```diff
@@ -0,0 +1,296 @@
...
+    bool isEmpty;
```
elBoberido: We always try to initialize a member, even if it is later on initialized in the ctor again. This is just a precaution in order to avoid problems when a new ctor is added and someone forgets to initialize all members.

Also, according our coding style, member shall be snake_case.

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
```diff
@@ -0,0 +1,296 @@
...
+            std::abort();
+        }
+        return std::move(m_value.value);
+    }
+
+    constexpr const T&& operator*() const&& noexcept {
+        if (m_value.isEmpty) {
+            std::abort();
+        }
+        return std::move(m_value.value);
```
elBoberido: Food for thought. What do you think about renaming `m_value` to `m_data`/`m_storage`/`m_something_else`? This would prevent the repeating `value`.

iceoryx2-bb/cxx/include/iox2/container/config.hpp
ComicSansMS: I don't mind having everything under `iox2`. The only reason not to do that would be name clashes, and I don't see that as a major issue for us, especially not if the base parts remain as tightly integrated with the rest of iceoryx2.

Another question is whether we want nested namespaces. There's [an argument against nested namespaces in C++](https://open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0816r0.pdf) as they make it more likely that people write weird name-lookup bugs.

iceoryx2-bb/cxx/include/iox2/container/config.hpp
ComicSansMS: Yes, that's also what I had in mind here, I just did not flesh it out yet.

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
ComicSansMS: `inline` variables are C++17.

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
ComicSansMS: This might have been due to the MISRA rule that says `union` must not be used. This implementation will require a deviation, but using something like `UninitializedArray` instead is certainly not a reasonable mitigation for that.

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
ComicSansMS: By "always initialize" you mean using class member initialization, like `bool is_empty = true;`?

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
ComicSansMS: Resolved by making the `value` field in the holder non-public.

iceoryx2-bb/cxx/include/iox2/container/config.hpp
elBoberido: Yeah, name clashes would be a problem.

We also overdid it at some point with nested namespaces and then we decided to have only one or at most two nested namespaces.

@elfenpiff what is you stance on this?

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
elBoberido: Oh, the `inline`. In iceoryx classic we used the `nullopt` constexpr without the inline keyword ... the compiler didn't complain but you are more familiar with the fine details about the standard :sweat_smile:

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
elBoberido: Yes, I meant class member initialization.

Are there good reasons not to do it?

From my understanding, it should not affect performance if the member is initialized via class member initialization as well as in the initializer list in the ctor.

ComicSansMS: Understood. That should be fine from a performance point of view, any modern SSA-based compiler will optimize the redundant writes away without breaking a sweat.

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
ComicSansMS: Without the `inline` it becomes a global variable definition, which you can't place in a header file. You can replace the `inline` with `extern` though and place the definition in a `.cpp` file. However, doing so will make it somewhat opaque to the compiler and could lead to worse codegen (although I'd hope that won't happen for an empty type like `Nullopt_T`).

Alternatively, I guess you could also make it `static` in the header file, but that is really ugly...

ComicSansMS: I added a proper pre-C++17 alternative implementation that is save from ODR-violations. I also adjusted the definition of NulloptT to make it consistent with C++17, mainly disallowing direct construction from the named type itself.

iceoryx2-bb/cxx/include/iox2/container/config.hpp
ComicSansMS: Tracked via #971.

iceoryx2-bb/cxx/include/iox2/container/config.hpp
ComicSansMS: Will be resolved offline.

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
ComicSansMS: Done.

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
ComicSansMS: Done.

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
ComicSansMS: Tracked via #972.

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
ComicSansMS: I did not do the separation for now, but I won't object to change this later if someone cares strongly about this.

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
ComicSansMS: Leaks are now explicitly tested through a fixture and tests should cover all relevant counters.

I added a fixture that allows enforces checking all counters always but after trying it out for a few tests, found that it created rather verbose test code that had a tendency to obscure what was actually being tested. The diagnostics from that fixture were also not great. Take another look at the tests and let me know what you think now. If you still feel it should be more strict, I can deploy a version using the strict fixture. For comparison, with a fully enforced check for all operations, the overall number of lines of code for the test suite would roughly double. Less verbose variations check only subsets of the operation. I felt that the current amount of checking to be a good sweet spot between verbosity and strictness.

iceoryx2-bb/cxx/iceoryx2-bb-containers-cxx.natvis
```diff
@@ -0,0 +1,14 @@
...
+
+  <Type Name="iox2::container::Optional&lt;*&gt;">
+    <Intrinsic Name="has_value" Expression="!m_value.m_is_empty"/>
+    <Intrinsic Name="value" Expression="m_value.m_u_value"/>
+    <DisplayString Condition="!has_value()">nullopt</DisplayString>
+    <DisplayString Condition="has_value()">{value()}</DisplayString>
+    <Expand>
+      <Item Name="[value]" Condition="has_value()">value()</Item>
+    </Expand>
+  </Type>
```
elBoberido, line 12: :star_struck:

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
```diff
@@ -0,0 +1,355 @@
...
+#include <type_traits>
+#include <utility>
+
+#if IOX2_CONTAINER_CONFIG_USE_STD_OPTIONAL && IOX2_CONTAINER_CONFIG_USE_CUSTOM_OPTIONAL
+#error Optional cannot be overriden to use both custom and std variants at the same time.
+#endif
+
+#if IOX2_CONTAINER_CONFIG_USE_STD_OPTIONAL
+#include <optional>
+#endif
```
elBoberido, line 29: Food for thought. Do you think there is a chance to have issues with some tooling that checks for usage of STL container and will complain here? Would it be an idea to hide the include behind another header which will be auto-generated/configured with the mechanism in #971?

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
```diff
@@ -0,0 +1,355 @@
...
+    }
+
+    // destructor
+#if __cplusplus >= 202002L
+    constexpr
+#endif
+        ~Optional() = default;
+
+    // assignment
+    constexpr auto operator=(NulloptT /* unused */) noexcept -> Optional& {
```
elBoberido, line 235: Fun fact when we did the certification for `iceoryx_hoofs`, there were rules which forbid the usage of C-style multiline comments and also required to have names for the arguments. This left us with a custom implementation of `maybe_unused`. I guess it is fine for now and we act if it is required again and we also have the `maybe_unused` again.

iceoryx2-bb/cxx/tests/CMakeLists.txt
```diff
@@ -26,12 +26,20 @@ include(Iceoryx2GTest)
 
 add_executable(iceoryx2-bb-containers-tests ${TEST_FILES})
 target_sources(iceoryx2-bb-containers-tests
+    PRIVATE
+    ${PROJECT_SOURCE_DIR}/src/testing/observable.hpp
+    ${PROJECT_SOURCE_DIR}/src/testing/test_utils.hpp
```
elBoberido, line 31: Out of curiosity. Are these header only for internal usage or is it planned that they are also provided to our users at some time?

In iceoryx classic we had a separate `*-testing` package for the latter case.

In any case, no further actions for this PR required.

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
ComicSansMS: Yes, that sounds very reasonable. I did not put too much thought into it when writing these lines, so there is certainly lots of room for improvement in this area.

iceoryx2-bb/cxx/tests/CMakeLists.txt
ComicSansMS: I intended them for internal use only, but if we think it's useful to provide them to users I am happy to provide packaging for them.

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
elBoberido: This can also be done in a follow up PR, together with #971. Maybe an action item can be added to that issue.

iceoryx2-bb/cxx/tests/CMakeLists.txt
elBoberido: No, it's fine like it is. If we change our mind, we can still move the header to a different cmake package. Was just curious what the current intention is.

iceoryx2-bb/cxx/CMakeLists.txt
```diff
@@ -37,6 +44,10 @@ target_include_directories(iceoryx2-bb-containers-cxx
 target_compile_features(iceoryx2-bb-containers-cxx INTERFACE cxx_std_14)
 # TODO check if this settings will be propagated to the user; maybe this should only be used for the tests
 target_compile_options(iceoryx2-bb-containers-cxx INTERFACE ${ICEORYX2_CXX_FLAGS} ${ICEORYX2_CXX_WARNINGS})
```
elBoberido: I've added this on a PR and it's still on my todo list before the release. It has been some time since I last had a closer look at how the flags are propagated for `INTERFACE`. In case you know it by chance, would be happy if you could shed some light on this.

iceoryx2-bb/cxx/cmake/GdbPrettyPrinter.cmake
```diff
@@ -0,0 +1,155 @@
+# Copyright (c) 2024 Contributors to the Eclipse Foundation
```
elBoberido: Nitpick. I also always forget to change the date when copying the header. Can you please change it to 2025 for all new files which do not have copied code.

iceoryx2-bb/cxx/cmake/GdbPrettyPrinter.cmake
```diff
@@ -0,0 +1,155 @@
...
+script at `<pretty-printer-file-path>` into `target`.
+
+It will create a custom target
+`<target>_pretty_printer_<pretty_printer_name>` that contains all build
+steps for producing the object file containing the gdb pretty printer
+data. That object file will then get added to `target` as a library link
+input.
+
+This function does nothing unless you are building natively on Linux and
+ICEORYX2_EMBED_GDB_PRETTY_PRINTER is set. This function may fail to perform
```
elBoberido, line 47: This option is quite hidden. Would it make sense to expose it in `options.cmake`?

iceoryx2-bb/cxx/iceoryx2-bb-containers-cxx-pretty-printer.py
elBoberido, line 1: :star_struck:

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
```diff
@@ -0,0 +1,355 @@
...
+/// Internal union implementation for Optional.
+/// @todo Proper handling of cond. explicit, cond. noexcept and triviality of special member functions.
+template <typename T>
+class OptionalValueHolder {
+  private:
+    union {
+        char m_u_null;
+        std::remove_cv_t<T> m_u_value;
+    };
+    bool m_is_empty = true;
```
elBoberido, line 75: I guess this could then later on be specialized for nullable types :)

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
```diff
@@ -0,0 +1,355 @@
...
+                               bool> = true>
+    // NOLINTNEXTLINE(hicpp-explicit-conversions), as specified in ISO14882:2017 [optional]
+    constexpr Optional(U&& value)
+        : m_value(std::forward<U>(value)) {
+    }
+
+    // destructor
+#if __cplusplus >= 202002L
+    constexpr
+#endif
```
elBoberido, line 231: Not for this PR, but what do you think about hiding these lines behind a define, e.g. `IOX2_CONSTEXPR_DTOR`? Would be a `good first issue`.

iceoryx2-bb/cxx/tests/src/testing/observable.hpp
```diff
@@ -0,0 +1,99 @@
...
+namespace iox2 {
+namespace container {
+namespace testing {
+
+/// A mock type that tracks invocations of all special member functions and maintains an id for tracking copies.
+class Observable {
+  public:
+    /// Counters keeping track of all operations performed on objects of type Observable.
+    static struct Counters {
+        int wasInitialized;     ///< Incremented for each invocation of a constructor other than copy or move.
```
elBoberido: Another unwritten rule. We try to only use fixed width integers, with few exceptions, like a POSIX API requiring it or when we want to cover all integer types. Int this case, it would be either `in32_t` or `int64_t`.

We also always do in-class member initialization to prevent accidentally ending up with an instance with uninitialized member.

I'm also wondering why clang-tidy does not complain about the members not being snake_case. Do we need to add `PublicMemberCase` to the clang-tidy file? @elfenpiff I guess your intention was to have all members snake_case, right?

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
```diff
@@ -0,0 +1,1008 @@
...
+#include <gtest/gtest.h>
+
+namespace {
+using iox2::container::testing::Observable;
+
+class OptionalFixture : public iox2::container::testing::DetectLeakedObservablesFixture { };
+
+TEST(Optional, default_constructor_initializes_empty_optional) {
+    // [optional.ctor] / 2
+    iox2::container::Optional<int> const sut;
```
elBoberido: Please also use in tests fixed-width integers. A few years ago, we spent quite some time to replace all the non fixed-width integers in iceoryx classic.

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
```diff
@@ -0,0 +1,1008 @@
...
+TEST(Optional, value_constructor_initializes_the_contained_value) {
+    int const contained_value = 42;
+    iox2::container::Optional<int> const sut(contained_value);
+    ASSERT_TRUE(sut.has_value());
+    EXPECT_EQ(*sut, contained_value);
+}
+
+TEST_F(OptionalFixture, value_constructor_move_constructs_for_rvalue) {
+    Observable::s_counter.wasInitialized = 0;
+    Observable::s_counter.wasMoveConstructed = 0;
```
elBoberido: Shouldn't this already be reset by the fixture?

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
```diff
@@ -0,0 +1,1008 @@
...
+        Observable const move_target = *std::move(sut);
+        ASSERT_EQ(Observable::s_counter.wasMoveConstructed, 1);
+        ASSERT_EQ(Observable::s_counter.wasMoveAssigned, 0);
+        ASSERT_EQ(Observable::s_counter.wasDestructed, 0);
+        ASSERT_EQ(move_target.id, tracking_id);
+    }
+    ASSERT_EQ(Observable::s_counter.wasDestructed, 2);
+}
+
+TEST_F(OptionalFixture, const_rvalue_operator_star_dereferences_to_const_rvalue_and_is_just_not_very_useful_overall) {
```
elBoberido, line 666: :laughing:

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
```diff
@@ -0,0 +1,1008 @@
...
+TEST(Optional, const_operator_arrow_should_bypass_overloaded_operator_ampersand) {
+    int const tracking_id = 54321;
+    iox2::container::testing::CustomAddressOperator obj;
+    obj.id = tracking_id;
+    iox2::container::Optional<iox2::container::testing::CustomAddressOperator> const sut { obj };
+    iox2::container::testing::CustomAddressOperator::s_count_address_operator = 0;
+    ASSERT_EQ(sut->id, tracking_id);
+    ASSERT_EQ(iox2::container::testing::CustomAddressOperator::s_count_address_operator, 0);
+}
+
```
elBoberido, line 1007: Maybe I missed it and it's also kind of a philosophical question. Should there be an explicit test that verifies that  a moved-from Optional is empty?

This would require to deactivate some warnings for the scope of the test.

elBoberido (commented):
Just some more nitpicks

iceoryx2-bb/cxx/CMakeLists.txt
ComicSansMS: `INTERFACE`  properties will indeed propagate to any target that depends on `iceoryx2-bb-containers-cxx` and will also become part of the respective exported target in the install tree. As such, we should only put flags in here that are required to build the iceoryx2 headers correctly.

iceoryx2-bb/cxx/cmake/GdbPrettyPrinter.cmake
ComicSansMS: What is the policy here? Every file has the copyright year in which it was created? Or the year when it was last changed? Or a range (2024-2025) from when it was created to when it was last changed?

iceoryx2-bb/cxx/cmake/GdbPrettyPrinter.cmake
ComicSansMS: I was thinking about having a generic `EMBED_DEBUG_SYMBOLS` option that guards this behavior for both MSVC and GNU-based builds. I can move that over to `options.cmake` if people think it is ready to be exposed publicly.

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
ComicSansMS: The idea was that the test always manually resets all relevant counters before the operation that it is being tested. This is redundant but I find that it makes it easier for a reader of the test to see what the test is actually being about. It does make the test less strict than it could be though. I am open to removing these kinds of redundant initializations if people they are distracting.

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
ComicSansMS: A moved-from optional is not empty, it contains an object in a moved-from state. This is rather unintuitive for casual users and probably not ideal for safety, but changing it would break API compatibility with `std::optional`. The behavior is already being tested implicitly by checking that the one of the move operations got invoked on ` Observable`, but could be made more explicit by a test that deactivates the respective `clang-tidy`  warning.

iceoryx2-bb/cxx/CMakeLists.txt
elBoberido: Okay, then please remove the flags and the warnings. Having them in the tests should hopefully be sufficient to catch the issues.

iceoryx2-bb/cxx/cmake/GdbPrettyPrinter.cmake
elBoberido: We used to update the file each year, but it was always cumbersome at the start of the year and it is also not required. Nowadays, we just have the file creation year in the copyright header.

iceoryx2-bb/cxx/cmake/GdbPrettyPrinter.cmake
elBoberido: No, if you think it is not yet ready for public use, then we can keep it as is. Was not aware of the state/intention.

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
elBoberido: Please create an issue or add it as action item to an existing issue so that it is not forgotten.

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
elBoberido: No, problem. Sounds reasonable. Might get out of sync when touched later on but it also does not break things. So lets keep it and try and check how it works out.

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
elBoberido: Yeah, the question is now how much do we want to deviate from the behavior of the `std::optional`. For iceoryx classic, we've chosen the path to have an empty `optional` after it is moved. Of course, one can always get a reference to the inner object and move it out, leaving a moved from object in the optional.

The idea was to try to not have nullable types in the iceoryx base library, except the `optional` and the `optional` needs to communicate that to the user. @elfenpiff shall we still follow the philosophy of iceoryx_hoofs?

elfenpiff: We need to write this down, but we should follow this philosophy:

1. Safety is the most crucial factor. It beats performance, API compatibility, and usability
2. Easy to use and hard to misuse is a safety factor. Also the API shall have no surprises.
3. Performance is essential, but not as crucial as 1 or 2

elfenpiff: Of course, one could now argue that it is surprising for the developer that a moved optional does no longer contain a value. On the other hand `clang` and `clang-tidy` will spit out a warning of `use after move` when accessing such an optional - therefore I would state that use after move is always a bad style, should be avoided at all costs and therefore the user is just surprised that a bad style is no longer "supported"

elBoberido: Point 2 will be difficult to accomplish. Depending on who is using it, it might be surprising that it does not behave like the unsafe STL. But I get what you mean.

iceoryx2-bb/cxx/cmake/GdbPrettyPrinter.cmake
ComicSansMS: Fixed in all newly added files.

iceoryx2-bb/cxx/CMakeLists.txt
ComicSansMS: Done. If this library becomes non-header only at some point in the future, I will add them back in as `PRIVATE`.

iceoryx2-bb/cxx/cmake/GdbPrettyPrinter.cmake
ComicSansMS: Then let's keep testing this internally for a while longer and make it public in a future PR.

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
ComicSansMS: Yes... no... maybe! 😅

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
ComicSansMS: Tracked as #999. I don't have the permissions yet to mark it as good first issue myself.

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
ComicSansMS: Added [a comment](https://github.com/eclipse-iceoryx/iceoryx2/issues/971#issuecomment-3242757875) to #971.

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
elBoberido: done

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
ComicSansMS: The problem is, if the use case is drop-in replacability for `std::optional`, nulling on move does not necesarily make it safer. Consider:

```cpp
std::optional<std::vector<int>> opt{ std::vector<int>{ 1, 2, 3, 4, 5 } };
std::vector<int> moved_out = std::move(opt).value();
*opt = std::vector<int>{ 6, 7, 8, 9 };
assert(opt->at(3) == 9);
```

This is valid for C++17 optional but would trigger language UB for an empties-on-move optional. Code that was written for C++17 `optional` would not be able to migrate to iox2's `optional` safely.

clang-tidy's bugprone-use-after-move would catch this though.


Also, the current API for `optional` has a bunch of other places that are unsafe. If I were to design an optional with a 'safe' API, as I attempted for static vector, it would look very different from what we have here. There may be room for such a type in iox2, but the goal here was compatibility with C++17.

elBoberido (approved):
PR looks good. Let's talk about the move semantics in a dedicated issue. No need to delay the merge for that.

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
elBoberido: Well, if one defines a termination as safe, then it would be safe ;)

Can you open a dedicated issue for this discussion or maybe a github discussion?

It has pros and cons and there might be things we did not consider. We have to weigh what is more important for us and also if there are silent errors if we deviate from the STL.

iceoryx2-bb/cxx/tests/src/testing/observable.hpp
ComicSansMS: > We also always do in-class member initialization to prevent accidentally ending up with an instance with uninitialized member.

Does always here really mean _always_? Specifically:

```cpp
struct Inner {
   int x = 0;
   int y = 0;
};
struct Outer {
  Inner i; // ?
};
```

How would you initialize `i` here? `={}`, or `={0}`, or `={0, 0};`. Or is it fine to omit the initializer in this case, since it is already been taken care off at the lower level?

I would also like to point out that since we are dealing with `static` here, the members of `s_counter` do get zero-initialized here, but any member initializers applied would only come into effect during dynamic initialization, making them susceptible to the static initialization order fiasco. So for types that are placed into `static` I would suggest to consider exempting them from this rule.

elBoberido: In that example we just used `Inner i {};` ... although, I think if we knew it was a class with a default ctor, which already did the initialization, we just used `Inner i;`

Aaargh, okay, the static initialization order fiasco is a good point. For the tests it is not that important anyway and the compiler should give us a hint. In order to prevent the static initialization fiasco, we usually also try to not have global static variables but hide them in (free) functions in order to initialize the variable when used the first time.

iceoryx2-bb/cxx/tests/src/optional_tests.cpp
ComicSansMS: Tracked as #1001.

Here is a variation of the example above that is not detected by clang-tidy and is not detectable at runtime by a check that would terminate the program:

```cpp
std::optional<std::vector<int>> opt{ std::vector<int>{ 1, 2, 3, 4, 5 } };
std::vector<int>& the_element = *opt;
std::vector<int> moved_out = std::move(opt).value();
the_element = std::vector<int>{ 6, 7, 8, 9 };
assert(the_element->at(3) == 9);
```
 This again works perfectly fine in C++17, but leads to language UB for an empties-on-move optional. What's worse is, depending on the implementation of `std::vector` in this case, this is likely to silently work on lower optimization levels, but break randomly on higher optimization levels.

Issue #951, opened 2025-08-14
Shared memory compatible static vector type with compile-time static maximum capacity


ComicSansMS:
## Brief feature description

Provide a C++ implementation of a static vector that can be shared between nodes. This container will need to be memory layout compatible with the Rust version of the same container. This container will roughly resemble `std::inplace_vector` in terms of functionality, but will not attempt to act as a drop-in replacement for the dynamic vector type. It will use an API that prioritizes safety.

PR #952, opened 2025-08-14
Static Vector Interface Review

ComicSansMS:
## Notes for Reviewer
This draft PR is used to discuss the interface of the static vector class for the base library (#951 ).

The goal is for now to discuss what we want the API to look like and how to evolve the container going forward.

This PR depends on the upstream changes from PR #949. For this review, focus on the changes in `static_vector.hpp`. I'm especially interested in feedback on names and semantics of the different methods, but feedback on implementation techniques is also welcome.

* [ ] Assign PR to reviewer

## References

Closes #


iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
```diff
@@ -0,0 +1,262 @@
...
+    static_assert(N > 0, "Static container with capacity 0 is not allowed.");
+    static_assert(std::is_standard_layout<T>::value, "Containers can only be used with standard layout types.");
+public:
+    using ValueType = T;
+    using SizeType = size_t;
+    using DifferenceType = ptrdiff_t;
+    using Reference = T&;
+    using ConstReference = T const&;
+    using Pointer = T*;
+    using ConstPointer = T const*;
```
ComicSansMS, line 45: I did not see a good way for a checked iterator API. The problem is that pretty much the entire iterator interface uses signatures that don't provide an error return path (other than exceptions).

So I chose here to leave the entire iterator API as unchecked, which can be achieved exposing just raw pointers. This can be made more robust by introducing iterator classes that detect preconditon violations and treat them as hard errors, but it can't be a "safe" API with the iterator model from the C++ standard library.

iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
```diff
@@ -0,0 +1,262 @@
...
+    using Iterator = Pointer;
+    using ConstIterator = ConstPointer;
+    using OptionalReference = Optional<std::reference_wrapper<T>>;
+    using OptionalConstReference = Optional<std::reference_wrapper<T const>>;
+
+    // Unchecked element access
```
ComicSansMS, line 51: The idea here is to contain all unchecked container operations to a proxy object. The usage would then be that instead of `container.begin()` a user has to write `container.unchecked_access().begin()`.

iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
```diff
@@ -0,0 +1,262 @@
...
+public:
+    // constructors
+    constexpr StaticVector() noexcept = default;
+    constexpr StaticVector(StaticVector const&) = default;
+    constexpr StaticVector(StaticVector&&) = default;
+    
+    template<uint64_t M, typename std::enable_if<(N >= M), bool>::type = true>
+    constexpr StaticVector(StaticVector<T, M> const& rhs)
+    :m_storage(rhs.m_storage)
+    {}
```
ComicSansMS: The idea here is that the capacity is not a salient part of the container's value, so it should be possible for vectors of different capacities to interact. This will also expand to things like comparison operators.

iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
```diff
@@ -0,0 +1,262 @@
...
+        if (m_storage.size < N) {
+            new (m_storage.pointer_from_index(m_storage.size)) T(std::forward<Args>(args)...);
+            ++m_storage.size;
+            return true;
+        } else {
+            return false;
+        }
+    }
+    
+    constexpr bool try_push_back(T const& v) {
```
ComicSansMS: The name `try_push_back` was chosen to distinguish it from `push_back` on standard library containers. The intention is not to mandate `try_` as a prefix for all operations that can fail.

The return type right now is `bool`, but could just as well be `expected<void, StaticVectorError>`. I don't feel strongly about this, but `bool` seemed fine for me here.

iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
```diff
@@ -0,0 +1,262 @@
...
+
+    constexpr SizeType size() const noexcept {
+        return m_storage.size;
+    }
+
+    bool empty() const {
+        return size() == 0;
+    }
+
+    OptionalReference element_at(SizeType index) {
```
ComicSansMS: Element access is currently via `Optional<std::reference_wrapper<T>>`. This is slightly less ergonomic than C++26's `optional<T&>`. If we can build proper support for reference types into iox2's `Optional` that would be an easy fix for that. Otherwise, a custom reference wrapper type that provides an `operator->` might improve usability for class member access, as the compiler will automatically chain nested calls to `operator->`.

iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
```diff
@@ -0,0 +1,262 @@
...
+        static_assert(std::is_standard_layout<T>::value, "Storage is only valid for standard layout types.");
+
+        alignas(T) char bytes[sizeof(T) * N];
+        uint64_t size;
+
+        constexpr StaticVectorStorage() noexcept
+        :bytes{}, size(0)
+        {}
+
+        template<size_t M, typename std::enable_if<(N >= M), bool>::type = true>
```
elfenpiff: This is really nice and something we didn't have for the classic iceoryx version.

iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
elfenpiff: When having a `try_` prefix then a bool return makes absolutely sense since the `try_` already conveys that it is alright if the try is not successful.

elBoberido: I think it boils down if we want that the result should be evaluated. The `expected` has a `nodiscard` attribute by default, so I also think it would be a bit of an overkill for this function.

iceoryx2-bb/cxx/include/iox2/container/config.hpp
elBoberido, line 1: In iceoryx classic, we decided to have a flat include structure and place everything into just `iox`, after we had some issues with moving headers around. Shall he have the same structure for iceoryx2?

An option would be to place everything into `iox2/bb` instead of just `iox2`.

@elfenpiff @ComicSansMS what do you think?

iceoryx2-bb/cxx/include/iox2/container/config.hpp
```diff
@@ -0,0 +1,24 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+#ifndef INCLUDE_GUARD_IOX2_CONTAINER_CONFIG_HPP
+#define INCLUDE_GUARD_IOX2_CONTAINER_CONFIG_HPP
+
+
+#define IOX2_CONTAINER_CONFIG_USE_STD_OPTIONAL 0
+
+#define IOX2_CONTAINER_CONFIG_USE_STD_EXPECTED 1
```
elBoberido: The question is if we want to make this a compile time option?

This can be done with converting this file into a `config.hpp.in` similar to this -> https://github.com/eclipse-iceoryx/iceoryx/blob/main/iceoryx_platform/linux/cmake/platform_settings.hpp.in#L22

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
```diff
@@ -0,0 +1,285 @@
...
+#if __cplusplus >= 201703L
+template<class T>
+Optional(T) -> Optional<T>;
+#endif
+
+#endif
+
+#if __cplusplus >= 201703L
+inline constexpr NulloptT nullopt;
+#endif
```
elBoberido: I'm not quite sure why this is only enabled for C++17?

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
```diff
@@ -0,0 +1,285 @@
...
+
+namespace detail {
+    /// Internal union implementation for Optional.
+    /// @todo Proper handling of cond. explicit, cond. noexcept and triviality of special member functions.
+    template<typename T>
+    struct OptionalValueHolder {
+        union {
+            char null;
+            typename std::remove_cv<T>::type value;
+        };
```
elBoberido: From my point of view, this is a sane approach but 7 years ago I had a discussion with an automotive guy that was against using `union`. I think the justification he gave was wrong but we had no choice and use something like the `UninitializedArray`.

Since you are are also in favor of the `union`, I guess you also do not see an issue with certification when it is properly encapsulated?

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
```diff
@@ -0,0 +1,285 @@
...
+        bool isEmpty;
```
elBoberido: We always try to initialize a member, even if it is later on initialized in the ctor again. This is just a precaution in order to avoid problems when a new ctor is added and someone forgets to initialize all members.

Also, according our coding style, member shall be snake_case.

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
```diff
@@ -0,0 +1,285 @@
...
+            std::abort();
+        }
+        return m_value.value;
+    }
+
+    constexpr T&& operator*() && noexcept {
+        if (m_value.isEmpty) {
+            std::abort();
+        }
+        return std::move(m_value.value);
```
elBoberido: Food for thought. What do you think about renaming `m_value` to `m_data`/`m_storage`/`m_something_else`? This would prevent the repeating `value`.

iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
```diff
@@ -0,0 +1,262 @@
...
+            }
+        }
+
+#       if __cplusplus >= 202002L
+        constexpr
+#       endif
+        ~StaticVectorStorage() {
+            for (uint64_t i = size; i != 0; --i) {
+                uint64_t const index = i - 1;
+                pointer_from_index(index)->~T();
```
elBoberido: Opportunity for micro-optimization :)
```suggestion
            for (uint64_t i = size; i != 0;) {
                i -= 1;
                pointer_from_index(i)->~T();
```

iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
elBoberido: Shouldn't this member type definitions use the same naming convention like the STL container and be lower_snake_case?

I know that we have a different naming convention but for this type of member type definitions, I think we should closely follow the STL for compatibility reasons.

I think it would be worthwhile to do the same with type traits.

@elfenpiff @ComicSansMS what do you think?

iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
ComicSansMS: > The `expected` has a `nodiscard` attribute by default, so I also think it would be a bit of an overkill for this function.

Although 1. C++14 will not support `nodiscard` and 2. MISRA will force you to explicitly drop even a `bool` return value. 😅

iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
elBoberido: Agree. But since I don't like to type that much, what do you think about just `unchecked`?

iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
elBoberido: In iceoryx classic, we had the option to convert a std::string into a fixed size string. Since the size of the std::string could exceed the capacity of the fixed string, we used `iox<into<lossy<FixedString>>(std_string)`.

Not for this PR but what do you think about a similar functionality also for the fixed vector. There could also be a `try_into<FixedVec>(std_vec)`.

And finally, this could also be used to convert a fixed vector with a larger capacity to one with a smaller capacity.

iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
elBoberido: @ComicSansMS C++14 not having `nodiscard` did not stop us on iceoryx classic

```C++
#if __cplusplus >= 201703L
#define IOX_NO_DISCARD [[nodiscard]]
#else
#define IOX_NO_DISCARD
#endif
```

Okay, technically we did not have it on C++14 :)

Maybe we should re-introduce the `IOX_NO_DISCARD`, `IOX_FALLTHROUGH` and `IOX_MAYBE_UNUSED` macros from iceoryx classic if we need to support C++14. The latter two had a working implementation for C++14 -> https://github.com/eclipse-iceoryx/iceoryx/blob/6eb6fb80c879f457e3a40d3fab5417ee2423e607/iceoryx_platform/linux/include/iceoryx_platform/attributes.hpp

iceoryx2-bb/cxx/include/iox2/container/config.hpp
elBoberido: moved to #949

iceoryx2-bb/cxx/include/iox2/container/config.hpp
elBoberido: moved to #949

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
elBoberido: moved to #949

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
elBoberido: moved to #949

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
elBoberido: moved to #949

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
elBoberido: moved to #949

iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
ComicSansMS: Unfortunately, MISRA forbids modifying loop counters in the body of a `for`-loop. 🥲

The reason I wrote it this way is that it mirrors STL-style reverse iterators, where `rbegin ≙ end` and `rend ≙ begin` and you apply an off-by-one before dereferencing. I'm happy to change that if that's too confusing though.

iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
ComicSansMS: Yes, that's a good point. I guess this boils down to: "To what extent do we expect these containers to be used in generic code that was written for STL containers".

For Optional I stuck with the STL-style naming, as that was intended to be a drop-in replacement for `std::optional`. For the containers, the relation is less strict, so I went with iceoryx-style names.

Strictly speaking, to support this use case we need to conform with the named container requirements, for `StaticVector` that would be [*SequenceContainer*](https://en.cppreference.com/w/cpp/named_req/SequenceContainer). Note that this also requires a sizable amount of unchecked operations. If we ever decide that we need proxy iterators (which is a design option for the more complicated containers), we will not be able to fulfill any of the named container requirements, as all of them expect `T* ptr = &container[0];` to work.

In practice, pre-C++20 most people are not aware that the named requirements even exist and just "wing it" with regards to what properties of the container they depend on.

For compatibility with Ranges, things are a lot more flexible. The container itself basically [only needs `begin` and `end`](https://www.en.cppreference.com/w/cpp/ranges/range.html) (which I fear will have to remain part of the unchecked API) and the majority of the requirements falls on the [iterator type](https://www.en.cppreference.com/w/cpp/iterator/forward_iterator.html). Fortunately, with Ranges all interfacing with the iterator happens through [`std::iterator_traits`](https://www.en.cppreference.com/w/cpp/iterator/iterator_traits.html), so we would be free to stick with our usual naming convention and just specialize the std trait correctly.

iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
ComicSansMS: I don't mind that.

Although in general, I am somewhat reluctant to design unsafe APIs for brevity. 🥲

iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
ComicSansMS: Sure, that sounds like a useful facility.

iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
elBoberido: I also don't have strong feeling about this ... with my current duties, I might not write that much code that use these constructs anyway :sweat_smile:

iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
elBoberido: Oh, just for clarification. I did not suggest to introduce them with this PR. Just would like to know it there is consensus.

iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
elBoberido: No problem. I guess the compiler will do the optimization anyway.

iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
```diff
@@ -0,0 +1,319 @@
...
+
+namespace detail {
+template <typename T, uint64_t N>
+class StaticVectorStorage {
+    // NOLINTNEXTLINE(modernize-type-traits), _v requires C++17
+    static_assert(std::is_standard_layout<T>::value, "Storage is only valid for standard layout types.");
+
+  private:
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays,hicpp-avoid-c-arrays,modernize-avoid-c-arrays) raw storage, will not be used as array
+    alignas(T) char m_bytes[sizeof(T) * N];
```
elBoberido: Food for thought. Not sure if it would make sense since we probably do not implement enough container. In iceoryx classic we had to write this code a lot, since there were more container and when we started with certification, the tools complained quite a lot. We encapsulated everything into [UninitializedArray](https://github.com/eclipse-iceoryx/iceoryx/blob/main/iceoryx_hoofs/container/include/iox/uninitialized_array.hpp).

Certainly not for this PR but since we try to re-license some of iceoryx_hoofs as Apache 2 or MIT, the question is if we would see some value in reusing the `UninitializedArray` later on?

The advantage of the `UninitializedArray` is that it already went through a certification cycle and would therefore hide one of the more low level constructs that certification tools quite often complain about. For example, in the case of the `UninitializedArray` we initially also just had a `alignas(T) char m_bytes[sizeof(T) * N];`, but the tool complained about array to pointer decay as well as unaligned access, since we used a pointer to a char, which is only aligned to 1 byte. The tool was not able to verify that we ensured that no unaligned access can happen. We then used the `UninitilizedArray` to fix these issues and still get complains if we really break the alignment. The cost was some type punning but that was considered safe.

iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
```diff
@@ -0,0 +1,319 @@
...
+        if (m_storage.size() < N) {
+            new (m_storage.pointer_from_index(m_storage.size)) T(std::forward<Args>(args)...);
+            m_storage.increment_size();
+            return true;
+        } else {
+            return false;
+        }
+    }
+
+    constexpr auto try_push_back(T const& value) -> bool {
```
elBoberido, line 183: Food for thought. In iceoryx classic, the philosophy was that it should be easy to port from the STL to our data types but not necessary the other way around. This meant that we had a `push_back` that did not throw exceptions but return a bool. In hindsight, this might have been not the best idea, since there was a semantic change but the code still compiled. The question is if there is some middleground, e.g. having a `push_back(T& v) -> bool` but with a deprecation annotation?

ComicSansMS: I agree that giving the same name to an operation with different semantics is a recipe for trouble. If we aim for consistency here, I guess the closest would be C++26's `inplace_vector`, which provides this operation under the same name, but returns a `pointer` to the inserted element instead of a `bool`. `inplace_vector` still throws `bad_alloc` on `push_back`, precisely for the reasons you mentioned.

I don't understand how a deprecation annotation would solve this problem. You would then get a compiler warning for every call to `push_back`, how is that useful?

elBoberido: The idea was that the code at least compiles and the user can start replacing the API calls at all occurrences since one gets all warnings with one compiler run instead of having to compile repeatedly until there are no errors anymore due to a missing API ... but yeah, might not be a good idea. It all boils down to how a migration path should look like. An additional benefit with the deprecation would be that the user immediately sees what function to use instead but the same could be achieved with a static assert in `push_back`.

ComicSansMS:
Closing this PR as the design review is finished. A new PR will be opened for merging. All changes have been moved to [this branch](https://github.com/ekxide/iceoryx2/tree/iox2-951-base-library-static-vector).

Issue #925, opened 2025-08-05
Event Handling


elfenpiff:
## Introduction

Currently, events have a limited buffer which depends on the mechanism (unix domain socket, socketpair etc.) and the operating system. It is unclear for the user how large the buffer size is, how it can be configured or if an event could be delivered to a specific listener.

Furthermore, iceoryx2 has internal events like "notifier dead" which might be critical and should always be delivered.

Also, a common bug we encountered was that users correlated event notifications one-to-one with sending samples. Meaning, for every sample they sent they also sent a notification. When the sample buffer is larger than the notification buffer, it may end up in deadlocks, or even if there are no more notifications, there are still samples in the buffer.

## Solutions

* **Bitset** 
  * every event id correlates with a specific bit
  * events can be delivered at most once, multiple events would still set the same bit
  * event id value would be very small
  * is already used for semaphores
  * event ordering is destroyed (event arrive in different order then sent)
* **SPSC Shared Memory Queue**
  * event connections are like pubsub connections, every notifier-listener pair has one shm queue
  * would be configured like a pubsub service
  * is from the behavior identically to pubsub, so it should be less confusing for the user
* **MPSC Shared Memory Queue**
  * every listener has one queue, similar to the buffer of a unix domain socket
  * challenge would be the implementation of a lock-free mpsc queue
* **FixedSizeArray of u64**
  * every event id correlates with a specific bit
  * every event delivery increments a counter, receive decrements it
  * event ordering is destroyed (event arrive in different order then sent) 
* **FixedSizeMap<event_id, u64>**
  * identical to fixedsize array

elBoberido:
@elfenpiff there were already request for a history for Events. The SPSC approach would also nicely solve this request.

eboasson:
Hi @elfenpiff, @elBoberido !

I think I am running into the issues mentioned here with an attempt at writing an Iceoryx2 plugin for Cyclone DDS: https://github.com/eclipse-cyclonedds/cyclonedds/pull/2264/commits/46402748cc715a24888463c8e62462543d4b2bb5 . Perhaps you can help me in deciding whether this is indeed the case.

My set-up is that an Iceoryx2 service instance is created for each (topic, DDS partition) pair, with its own listener, notifier and waitset and a thread blocked on the waitset. (Probably a sub-optimal approach, but one has to start somewhere.) The writer is in blocking mode, Iceoryx2 history depths are at the default.

[...]

PR #1005, opened 2025-09-03
iox-1002: Async API design draft

### LINKED ISSUE #1002: Async API
## Brief feature description

Iceoryx2 messaging patterns should support `async` API so they can be used in non blocking, reactive way in runtimes like `tokio` or `async_runtime`

Related:
- https://github.com/eclipse-iceoryx/iceoryx2/issues/627
- https://github.com/eclipse-iceoryx/iceoryx2/issues/246
- https://github.com/eclipse-iceoryx/iceoryx2/issues/47

pawelrutkaq:

* [ ] Assign PR to reviewer

## References

Closes #1002


pawelrutkaq:
@elBoberido  @elfenpiff  here is PR for async api we dicussed. Matrix is dead for now so I ping You here

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,203 @@
+# Async API Design Document
```
elfenpiff, line 1: I already skimmed over the document and I think it makes sense to have a meeting to discuss the details of this design document, when the major questions are addressed.

Could you please address the following things in the design doc:
1. What additional dependencies would be added to iceoryx2 (if it would not add any additional dependency you can ignore the following sub-points)
  * Are the dependencies `no_std`, if not, how do we port iceoryx2 to `no_std` platforms?
  * In an ASIL-D deployment, who would certify the dependencies?
2. When only the listener port of the event messaging pattern is blocking, why do we need to implement also other async ports? Could you please introduce the async topic a bit in this design document since not everyone (including me) knows rust async API in detail - but only the things required to understand the design doc.

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,203 @@
...
+    * polling and repetition logic within functions, causing additional not needed
+    code on user side
+    * custom WaitSet usage in separate thread and a bridge to `async` world.
+    * or some other custom work
+* The `async` usage in Rust is already well established technique that is
+  adopted by many crates, including those with highest usage
+
+## Requirements
+
+* **R1: Async API look and feel** \* The new `async` API shall provide the same
```
elfenpiff, line 133: This design document must address if this API is usable in a safety-critical context (for example ASIL D or for medical devices SIL-3). If not it may make sense to provide this API not directly on the iceoryx2 API layer but in the iceoryx2-userland layer which would give us even more freedom but it would not be usable in a safety-critical context, except as QM software.

If this API shall be usable in a mission critical context, we must add requirements like:
1. The API must be deterministic, no blocking calls, no implicit scheduling in background threads etc.
2. The API shall not use the heap or allocate

elfenpiff (commented):
@pawelrutkaq  Thanks for writing up the async design document - it was already requested from the community right from the beginning and we are happy that you started to tackle the challenge - thank you! 

I already read over it once but now I have to start sadly my meeting marathon - I left some initial questions.

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,203 @@
...
+* **As a** developer
+* **I want** to wait on `new sample` in pub-sub
+* **So that** it does not block current thread and continues only once sample is
+  delivered
+
+## Usage
+
+### Example: Await on events
+
+```rust
```
elfenpiff, line 158: Just for some context, could you provide a truly async example in pseudo-rust-code (maybe at the end of the design doc). So I would like to see how it would be used if a user, for instance, has two subscribers and two listeners, and whenever one listener is triggered, it takes out the sample of the subscriber and prints it on the console.

I am asking for a more complex example to just get a feeling of what else it would require to use this API to its full extent. 
Could you also assume, for example, that it would be a mission-critical context - so the threads must be explicitly controllable by the user, for instance, when they shall be pinned to a specific CPU core or run with a predefined priority.

doc/design-documents/drafts/async_api/async_api.md
pawelrutkaq: 1. The goal is to not add deps to `iceoryx2` and keep Event implementation in adapter crates which depends on runtime. So for S-CORE it would have it build in into https://github.com/eclipse-score/inc_orchestrator and if one wants for other runtime he can just create ie `iceoryx2-runtimeXYZ` and implement there
2. Okey will do

doc/design-documents/drafts/async_api/async_api.md
pawelrutkaq: This may be some dicussion as this is `RUNTIME` dependent. So what in my opinion can be done on `iceoryx2` is to provide API, but it's safety-critical context can only be evaluated by user in given project once coupling it with runtime. I still  agree we shall keep other parts like `zero-alloc` etc.

doc/design-documents/drafts/async_api/async_api.md
pawelrutkaq: I will add example.
```
Could you also assume, for example, that it would be a mission-critical context - so the threads must be explicitly controllable by the user, for instance, when they shall be pinned to a specific CPU core or run with a predefined priority.
```
This is not part of our concern here. This is bounded to runtime itself. Ie in Tokio I think you dont have much control, however in S-CORE async_runtime You are providing all those details upfront creating runtime. But this API will (and can't) have no control on it. At the end iceoryx2 is (i guess) `Safety element Out of context` and here probably some `Assumption of use/Safety manual` would be needed for this

pawelrutkaq:
@elfenpiff requests are added

doc/design-documents/drafts/async_api/async_api.md
elfenpiff: So we should add also a use case/user story that the async API shall also work together with the s-core inc orchestrator.

I think this is valuable background knowledge since it provides us with a context. If for whatever reason we would break the API we would also know who would be affected and that we need to find a solution.

doc/design-documents/drafts/async_api/async_api.md
elfenpiff: This makes sense, still such an example would illustrate the assumptions of use and may become also part of the iceoryx2 documentation itself or we add this in our examples folder.

pawelrutkaq:
@elBoberido @orecham kind remainder ;)

doc/design-documents/drafts/async_api/assets/event_example.rs
```diff
@@ -0,0 +1,103 @@
+// Copyright (c) 2023 Contributors to the Eclipse Foundation
```
elBoberido: Nitpick: I guess this is most probably not copied but new code ... except the license header, of course. If this is the case, then please change the year to 2025

doc/design-documents/drafts/async_api/assets/event_example.rs
```diff
@@ -0,0 +1,103 @@
...
+    listener: iceoryx2::port::listener::Listener<Service>,
+    io: BridgedFd<RawFdBridge<AsyncSelector>>,
+}
+
+impl<Service> Listener<Service>
+where
+    Service: service::Service,
+    <Service::Event as iceoryx2_cal::event::Event>::Listener: FileDescriptorBased,
+{
+    pub(crate) fn from(listener: iceoryx2::port::listener::Listener<Service>) -> Result<Self, CommonErrors> {
```
elBoberido, line 26: Out of curiosity, would it make sense to implement the TryFrom trait here? If not, it might be a good idea to call this function `try_from`, since it has the same signature like the one from the trait.

doc/design-documents/drafts/async_api/assets/event_example.rs
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /// Returns the deadline of the corresponding [`Service`](crate::service::Service).
+    pub fn deadline(&self) -> Option<Duration> {
+        self.listener.deadline()
+    }
+
+    /// Async wait for a new [`EventId`]. On error it returns [`ListenerWaitError`] is returned which describes
+    /// the error in detail.
+    pub async fn wait_one(&self) -> Result<EventId, ListenerWaitError> {
```
elBoberido, line 51: Clarification: Do you expect that there is a one-to-one relation between sent samples and received events?

doc/design-documents/drafts/async_api/assets/event_example.rs
```diff
@@ -0,0 +1,103 @@
...
+                Ok(event) if event.is_some() => return Ok(Ok(event.unwrap())),
+                Ok(_) => {
+                    // This is None, so there was and error, probably EAGAIN or EWOULDBLOCK
+                    if std::io::Error::last_os_error().kind() == std::io::ErrorKind::WouldBlock {
+                        error!("Iceoryx listener would block, should do re-register!... {}", unsafe {
+                            self.listener.file_descriptor().native_handle()
+                        });
+                        return Err(std::io::ErrorKind::WouldBlock.into());
+                    } else {
+                        panic!("Something went wrong!");
```
elBoberido, line 79: I probably would make sense to print the actual error kind which caused this panic, else debugging this would be a nightmare.

doc/design-documents/drafts/async_api/assets/event_example.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// Async wait for a new [`EventId`]. On error it returns [`ListenerWaitError`] is returned which describes
+    /// the error in detail.
+    pub async fn wait_one(&self) -> Result<EventId, ListenerWaitError> {
+        self.io
+            .async_call(IoEventInterest::READABLE, |raw_fd| {
+                raw_fd.io_call(|fd| {
+                    info!("Checking for Iceoryx event on fd: {}", fd);
+                    self.wait_one_internal()
+                })
+            })
```
elBoberido, line 58: I'm quite curious about this call. Maybe it will be answered later on, but is there a background thread which will trigger the waker or are you able to "register" the FD directly at the executor level so that no additional thread is required?

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,384 @@
...
+The Rust `async` APIs are build with the help of usage of compiler. At the end
+`async` and `await` is just a syntactic sugar that is removed during `HIR`. In
+practice it means:
+
+* `async fn abc() -> bool` signature is turned into
+  `fn abc() -> impl core::future::Future<Output=bool>`. This basically means
+  each async fn is a anonymous type that implements `core::future::Future`
+* `await` is turned be compiler into generated state machine that simply
+  decompose to all states (continues and returns) in function using `Future`
+  trait API. To have small example, below code:
```
elBoberido: Either it's late or there are some typos in this sentence. Right now, this would make more sense for my brain
```suggestion
* `await` is turned by the compiler into a generated state machine that simply
  decomposes to all states (continues and returns) in a function using the `Future`
  trait API. For a small example, see the code below:
```

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,384 @@
...
+
+#### Events
+
+Nothing to add, blocking API needs async versions
+
+#### Publisher Subscriber
+
+Here in async world (and even in non async really) as a user expectation would
+be that I can react once a new `sample` is produced so that I don't have to poll
+for it, instead simply `await` it.
```
elBoberido, line 121: I'm not sure how this would look like with iceoryx2. The publish-subscribe messageing pattern completely circumvents to OS, therefore there is no other option than "polling" the subscriber with that `receive` function. In order to have a meaningful `await`, it would need to be combined with an `Event`. But my async knowledge is limited. Is it required or at least doe it simplify the design when `receive` also returns a future, even though it does not have a meaningful waking mechanism?

doc/design-documents/drafts/async_api/async_api.md
elBoberido: If it is runtime dependent, would it make sense to have the requirements stated by @elfenpiff in this document, but only for the parts of the implementation that are merged to this repo? Or maybe a requirement that the code in iceoryx2 should not be runtime dependent would be sufficient?
I know it's not always possible with the current async features but I think it is also your goal.

doc/design-documents/drafts/async_api/async_api.md
elBoberido: Our ideas for certification are a little bit different. We do not want to be too opinionated with our API in order to not artificially limit the design space of our users. Therefore the idea is to provide a safety manual which states guarantees instead assumptions of use, when a specific API is used.

But your points hold. It should be the runtime do decide where to run a task and with which priority. Of course the runtime could offer options to configure this. But it would not be part of the components in iceoryx2 itself.

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,384 @@
...
+API only on change of data. This means:
+
+* `PubSub` should be paired with `Event`
+* `RequestResponse` should be paired with `Event`
+* and so on
+
+Due to this, further document assumes direct usage of high level `Event`
+messaging pattern to facilitate the feature.
+
+> DISCLAIMER: IT may be that iceoryx2 authors will have better idea than that.
```
elBoberido:
```suggestion
> DISCLAIMER: It may be that iceoryx2 authors will have better idea than that.
```

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,384 @@
...
+
+### Achieving async API
+
+Since all iceoryx2 messaging patterns (except `Event`) are poll based, we need
+to pair them them with the `Event` to achieve possibility to react in `async`
+API only on change of data. This means:
+
+* `PubSub` should be paired with `Event`
+* `RequestResponse` should be paired with `Event`
+* and so on
```
elBoberido, line 189: Okay, I guess my comment from above is moot :sweat_smile:

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,384 @@
...
+### Split of code
+
+The main idea is to split source code into two parts:
+
+1. The Event implementation that is both `OS` dependent and `runtime` dependent
+    since it incurs some IO call
+2. All the other API that do need `Event` implementation, but the rest is
+    purely `runtime` independent and can be pure `async`
+
+To facilitate above, below class diagram is showing one of solution
```
elBoberido:
```suggestion
To facilitate above, below class diagram is showing one solution
```

or maybe `one of the solutions`?

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,384 @@
...
+
+##### Cons
+
+* limit a service to only single messaging pattern as using event will cause no
+  way to use it again
+
+Considering above, continuation is done based on option 1. Below shows only
+small snippet where extension for creating object can be placed.
+
+![Port factory](port_factory.svg)
```
elBoberido: I'm not quite sure if you mean that the linked diagram is supposed to belong to option 1

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,384 @@
...
+`non-blocking` api of `Listener`. Working example:
+[Code sample](assets/event_example.rs) During implementation it may come
+beneficial to either expose some properties (of current listener) or `add` new
+sync api with different signature.
+
+#### AsyncSubscriber - `messaging-patter == pubsub`
+
+Pure `async` implementation not using any specifics of `runtime` shall be
+doable. In case some unexpected dependency will be needed it has to be exposed
+over defined abstraction same as `AsyncListenerTrait`
```
elBoberido, line 270: I'm a bit confused. Further above, it is stated that pub-sub always needs to be combined with the event messaging pattern.

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,384 @@
...
+              .wait_all(&mut |event_id| {
+                  print!("Received Iceoryx event: {:?}\n", event_id);
+              })
+              .await // During this not being ready, worker can do any other work
+              .unwrap();
+       } 
+  });
+
+
+ spawn(async move {
```
elBoberido:
```suggestion
  spawn(async move {
```

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,384 @@
...
+      .service_builder(&"MyEventName".try_into().unwrap())
+      .event()
+      .open_or_create()
+      .unwrap();
+
+  let listener = event.listener_builder().create_async().unwrap();
+  let listener2 = event.listener_builder().create_async().unwrap();
+
+  let task1_handle = spawn(async move {
+    println!("Awaiting for Iceoryx event in batches while doing something else ...");
```
elBoberido: nitpicking
```suggestion
    println!("Awaiting for iceoryx event in batches while doing something else ...");
```

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,384 @@
...
+    loop {
+
+        listener
+              .wait_all(&mut |event_id| {
+                  print!("Received Iceoryx event: {:?}\n", event_id);
```
elBoberido: same here :)
```suggestion
                  print!("Received iceoryx event: {:?}\n", event_id);
```

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,384 @@
...
+async fn main() {
+  let node = NodeBuilder::new().create::<ipc_threadsafe::Service>().unwrap();
+
+  let event = node
+    .service_builder(&"MyEventName".try_into().unwrap())
+    .event()
+    .open_or_create()
+    .unwrap();
+
+  let notifier = event.notifier_builder().create_async().unwrap();
```
elBoberido, line 392: For me it is not clear why there should be an async notifier?

elBoberido (commented):
@pawelrutkaq thanks for bringing this forward and sorry for taking so long to review.

For me it's still not clear what changes in iceroyx2 would be required to implement the design. Partly also because I'm not that familiar with async and therefore some things that might be obvious for you, are unclear to me.

From my current feeling, I'm not even sure we need a port factory with a `create_async` method. This is something that we won't need for the other language bindings. It might be sufficient to have an `AsyncListener` in the userland crates, which just consumes a `Listener` and offers the async API.

But in order to give a better response, I need a better picture of the proposed changes and where they need to be done in iceoryx2.

doc/design-documents/drafts/async_api/assets/event_example.rs
pawelrutkaq: The IO driver in muti threaded runtime is traveling across threads. So here FD is directly registered in IO driver and driver will get polled on some worker of runtime once the time comes. But there is no separate IO driver thread doing IO only.

doc/design-documents/drafts/async_api/async_api.md
pawelrutkaq: The real async need to take the waker and be waked on some condition(when there is real chance that connected future can do some progress). So since API of `receive` is polling as You said, we need to add reactivity by Event as written below.

doc/design-documents/drafts/async_api/async_api.md
pawelrutkaq: This can be skipped at first line since blocking is unlikelly. But to be 101% accurate, notifier is a `write` syscall that can block ;)

doc/design-documents/drafts/async_api/assets/event_example.rs
pawelrutkaq: one of two you said makse sense. Since it's example I dont correct it for now

doc/design-documents/drafts/async_api/assets/event_example.rs
pawelrutkaq: This is `Event` part. so here we have one to one realtions. However second part of your questions is more into pub-sub. Here our idea is:
- first implemetation - yes, each new produced sample fires event
- later stage (optimization when rolled out) - one can limit number of events by implemeting special checks in pub-sub internals where we can fire event only if we know async subscriber as now awaiting new sample.

Since my approach is to first do not touch iceoryx2 implemetation details.

doc/design-documents/drafts/async_api/assets/event_example.rs
pawelrutkaq: Yup, as said, its only example code.

doc/design-documents/drafts/async_api/async_api.md
pawelrutkaq: done

doc/design-documents/drafts/async_api/async_api.md
pawelrutkaq: done

doc/design-documents/drafts/async_api/async_api.md
pawelrutkaq: done

doc/design-documents/drafts/async_api/async_api.md
pawelrutkaq: done

doc/design-documents/drafts/async_api/async_api.md
pawelrutkaq: done

doc/design-documents/drafts/async_api/async_api.md
pawelrutkaq: Yes, but once You have `iml AsyncListenerTrait` the rest is pure async (no runtime dependent). So assuming we have somehwere `type X:  AsyncListenerTrait`, X is becoming given and we dont depend on runtime in implementation (we only depend indirectly on concreet X type ensuring it has out trait implemented).

doc/design-documents/drafts/async_api/async_api.md
pawelrutkaq: Yes

doc/design-documents/drafts/async_api/async_api.md
pawelrutkaq: I think we can:
- state guarantees for pure async API implemented in iceroyx2 given the AsyncListener is fullfilling requirements we specify.

The scheduling part is always out of iceoryx2 scope (sync or async API).

doc/design-documents/drafts/async_api/async_api.md
pawelrutkaq: I think stating requirements how AsyncListenerTrait shall behave is doable as requirements are not runtime dependent. Only implementation is.

pawelrutkaq:
> @pawelrutkaq thanks for bringing this forward and sorry for taking so long to review.
> 
> For me it's still not clear what changes in iceroyx2 would be required to implement the design. Partly also because I'm not that familiar with async and therefore some things that might be obvious for you, are unclear to me.
> 
> From my current feeling, I'm not even sure we need a port factory with a `create_async` method. This is something that we won't need for the other language bindings. It might be sufficient to have an `AsyncListener` in the userland crates, which just consumes a `Listener` and offers the async API.
> 
> But in order to give a better response, I need a better picture of the proposed changes and where they need to be done in iceoryx2.

My assumption is that first API can be created without changing the code, only extending is needed.
I think in case of `async` it may be impossible to always have same API across languages since EACH language do this differently.
At the end, I would disagree with doing all that in userland since this will limit further performance improvements and will not keep it consistent imho - and no one wants iceoryx2-tokio API and iceoryx2-other_runtime API looking similar, but actually not.

doc/design-documents/drafts/async_api/assets/event_example.rs
elfenpiff: @pawelrutkaq I just skimmed over the comments - so this comment might be misplaced.

> first implemetation - yes, each new produced sample fires event

This is an implementation flaw of iceoryx2 that will be fixed. The issue is here: https://github.com/eclipse-iceoryx/iceoryx2/issues/925 and we are currently writing a design doc for it.

The event is in the and a switch that is triggered when a state changes. If the same state changes multiple times, the switch is only triggered once until the counterpart resets it by handling all events.

This means for the following sequence, that only one event is emitted.

1. Publish sample
2. Notify sample-published (emits event notification)
3. Publish sample
4. Notify sample-published (does nothing)
5. ....

When the listener is activated and acquires all events then the event is reseted and the event notified once again.

The reason for this behavior change is, that it is essential in a mission-critical context that no notification is lost. This can be best realized with some form of bitset - not a queue with a limited buffer size.

doc/design-documents/drafts/async_api/async_api.md
elfenpiff: Actually, it is not. The underlying socket is configured with `fcntl(fd, F_SETFL, O_NONBLOCK)` and will return an errno when it would need to block.

Let's assume the notifier has 2 listeners and of one listener, the buffer is full, then the listener is skipped and not notified, the other listener will receive a notification. If the user want to handle this error, a degradation-callback can be registered which is called whenever a delivery problem is encountered.

This is also, as contract, verified in the concept test suite so you can therefore safely assume that this is always non-blocking - if its blocking it is a bug.

doc/design-documents/drafts/async_api/assets/event_example.rs
elfenpiff: @pawelrutkaq We have iceoryx2 users that use semaphores instead of unix domain sockets for signaling mechanism - since unix domain sockets are non-certifiable and not allowed in mission-critical contexts.

As you see here, the Listener implements only `file_descriptor()` when the underlying concept is file descriptor based: https://github.com/eclipse-iceoryx/iceoryx2/blob/main/iceoryx2/src/port/listener.rs#L128

When a user now uses the `Semaphore` variant, `async` cannot be implemented since `Semaphore`s alone do not support event multiplexing.

Currently, we already have the issue, that not all iceoryx2 code compiles (waitset breaks) when the semaphore configuration is used. But we must avoid adding more code that has the same issue.

One way of addressing this would be to introduce another service variant called `ipc_async_threadsafe` and define the event: https://github.com/eclipse-iceoryx/iceoryx2/blob/main/iceoryx2/src/service/ipc.rs#L53 as socket based.

But this will wonderfully clash with the C, C++ and Python language bindings - so it would be a service variant that is not cross-language compatible.

@elBoberido what is your opinion here.

doc/design-documents/drafts/async_api/assets/event_example.rs
```diff
@@ -0,0 +1,103 @@
...
+                Ok(event) => Ok(event),
+                Err(e) => Err(e),
+            })
+    }
+
+    fn wait_one_internal(&self) -> IoResult<Result<EventId, ListenerWaitError>> {
+        loop {
+            match self.listener.try_wait_one() {
+                Ok(event) if event.is_some() => return Ok(Ok(event.unwrap())),
+                Ok(_) => {
```
elfenpiff, line 71: So the assumption is that when `try_wait_one()` returns nothing that there was an error?

This assumption would be wrong. One-to-one relations between sample and event notification will soon no longer be possible and events will have switch semantics - otherwise we have no way of dealing with limited queue buffer sizes like every mechanism like unix domain sockets has.

doc/design-documents/drafts/async_api/assets/event_example.rs
```diff
@@ -0,0 +1,103 @@
...
+                    // This is None, so there was and error, probably EAGAIN or EWOULDBLOCK
+                    if std::io::Error::last_os_error().kind() == std::io::ErrorKind::WouldBlock {
```
elfenpiff, line 73: Please remove the whole error handling based on `errno`. This is broken since iceoryx2 resets the errno variable always to zero when it is handled - like the POSIX standard recommends.

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,384 @@
...
+    code on user side
+    * custom WaitSet usage in separate thread and a bridge to `async` world.
+    * or some other custom work
+* The `async` usage in Rust is already well established technique that is
+  adopted by many crates, including those with highest usage
+
+### Introduction to async in context of iceoryx2
+
+The Rust `async` APIs are build with the help of usage of compiler. At the end
+`async` and `await` is just a syntactic sugar that is removed during `HIR`. In
```
elfenpiff, line 33: Please add `HIR` to the terminology section with this link: https://rustc-dev-guide.rust-lang.org/hir.html and a short one sentence description.

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,384 @@
...
+
+fn some_other_async_fn() -> impl core::future::Future<Output=()> {
+  println!("Test");
+  core::task::Poll::Ready(())
+}
+
+fn abc() -> impl core::future::Future<Output=bool> {
+
+  let future = some_other_async_fn();
+  let mut _hidden_satte = 0;
```
elfenpiff: I think you mean `_hidden_state`

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,384 @@
...
+
+  if _hidden_satte == 0 {
```
elfenpiff: `hidden_state`

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,384 @@
...
+    match future.poll(...) {
+      core::task::Poll::Pending => return Pending
+      core::task::Poll::Ready(_) => {
+        ///
+        _hidden_satte = 1;
```
elfenpiff: `_hidden_state`

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,384 @@
...
+The main idea is to split source code into two parts:
+
+1. The Event implementation that is both `OS` dependent and `runtime` dependent
+    since it incurs some IO call
+2. All the other API that do need `Event` implementation, but the rest is
+    purely `runtime` independent and can be pure `async`
+
+To facilitate above, below class diagram is showing one of the solution
+
+![Class Diagram](new_classes.svg)
```
elfenpiff, line 210: Can you please use mermaid. It is supported directly in GitHub, can be written directly in markdown code and we get rid of the puml to svg conversion.

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,384 @@
...
+
+### Building objects
+
+The next step is to provide a way to build objects that provide `async` API.
+
+There are two approaches that can be chosen for implementation:
+
+#### 1. Use custom event for each pair (messaging patern, data type) based on `ServiceName`
```
elfenpiff: Maybe I had more context last time, but I have currently a hard time understanding this. Let me translate this.

1. Create corresponding event-service for each data-flow service. So when we have a service `my_pub_sub_service` we create also a `my_pub_sub_service_events`.
2. Global event service, we use one event service like `global_events` and all data-flow services use that for notification.

If I am correct, could you please rewrite this a bit to make it more understandable.

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,384 @@
...
+
+#### AsyncListener - `messaging-patter == event`
+
+This is purely `runtime` specific implementation but is currently doable with
+`non-blocking` api of `Listener`. Working example:
+[Code sample](assets/event_example.rs) During implementation it may come
+beneficial to either expose some properties (of current listener) or `add` new
+sync api with different signature.
+
+#### AsyncSubscriber - `messaging-patter == pubsub`
```
elfenpiff, line 266: I would recommend removing `AsyncSubscriber` from the design document for now. I agree that we need some kind of `async` subscriber, and we have this on our radar.

This would be more like a "meta-port" instead of a subscriber located in `iceoryx2-userland` and it combines control-flow pattern (event) with data-flow pattern (pubsub, request-response, blackboard). It would have an `async` API and would enable the user to wait on a wide variety of events of the data-flow messaging pattern. But on the `iceoryx2` layer the async API makes, in my opinion, only sense for the listener.

Another point is that we have to always consider the full iceoryx2 API. Even when you only require an async subscriber, still an async publisher, async client, async server, async reader (blackboard), async writer (blackboard) might be missing. They might wait until there is a subscriber, or a corresponding other part.

doc/design-documents/drafts/async_api/async_api.md
pawelrutkaq: `EWOULDBLOCK` is normally translated to `.await` doing async waiting. Here, this can happen too. So this looks like it still shall be async API

elfenpiff (commented):
I think we should reduce the scope of this design document to the async listener and plan it. I added some more comments but I am optimistic that we can merge the document until the end of this week.

To add further async ports we need to discuss the iceoryx2-userland layer and how there data-flow and control-flow messaging patterns are merged together which would than naturally converge to an `async` API.

doc/design-documents/drafts/async_api/assets/event_example.rs
pawelrutkaq: This is esentially needed in FD case that we get EWouldBlock in any runtime. Otherwise  we dont know when to wait. And You are not clearing this specific error ;) Anyway this information shall be preserved or we need side API.

doc/design-documents/drafts/async_api/assets/event_example.rs
pawelrutkaq: The Semaphores are not events. So in case user have system where we dont have proper notification backend (hope we dont dicuss here QNX ;)) then the only thing user can do is polling or ? In that case, using impl for specific trait (Event - FileDescriptor or Event - Semaphore)  we can implement different stategy. For semaphore it would be simply timeout based polling.

elBoberido: For Semaphores we could implement a WaitSet that works with them and for async, a separate thread could be spawned to wait on this WaitSet and the use e.g. EventFDs to wake the async runtime.

Regarding the additional service variant. I think we need a solution that scales better than what we have today. But that is not a simple task and needs some investment.

doc/design-documents/drafts/async_api/async_api.md
elBoberido: I do not fully understand this. The Pub-Sub messaging pattern will always be polling and reactivity can only be achieved by Events. So one would never do an `await` on a `receive` but only on a `*_wait_*` call of the Listener.

On a higher level, there might be a subscriber which is combined with an listener, but that is not the case on the low level API.

doc/design-documents/drafts/async_api/async_api.md
elBoberido: Maybe it's because I'm not a native speaker but could you rephrase the sentence a little bit, e.g. `... the continuation of the proposal is based on option 1. The small snippet below shows where the extension for creating the object can be placed.`

doc/design-documents/drafts/async_api/async_api.md
elBoberido: Maybe I'm not getting the point. Do you propose to add an separate `AsyncSubscriber` which has an integrated `Listener`?

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,384 @@
...
+The Rust `async` APIs are build with the help of usage of compiler. At the end
+`async` and `await` is just a syntactic sugar that is removed during `HIR`. In
+practice it means:
+
+* `async fn abc() -> bool` signature is turned into
+  `fn abc() -> impl core::future::Future<Output=bool>`. This basically means
+  each async fn is a anonymous type that implements `core::future::Future`
+* `await` is turned be the compiler into a generated state machine that simply
+  decomposes to all states (continues and returns) in a function using `Future`
+  trait API. For a small example, see the code below:
```
elBoberido:
```suggestion
* `await` is turned by the compiler into a generated state machine that simply
  decomposes to all states (continues and returns) in a function using the `Future`
  trait API. For a small example, see the code below:
```

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,384 @@
...
+### Split of code
+
+The main idea is to split source code into two parts:
+
+1. The Event implementation that is both `OS` dependent and `runtime` dependent
+    since it incurs some IO call
+2. All the other API that do need `Event` implementation, but the rest is
+    purely `runtime` independent and can be pure `async`
+
+To facilitate above, below class diagram is showing one of the solution
```
elBoberido:
```suggestion
To facilitate above, below class diagram is showing one of the solutions
```

elBoberido (commented):
Just some small typos

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,384 @@
...
+
+Here in async world (and even in non async really) as a user expectation would
+be that there is one entity that is a "Server" (the one that hosts method and
+produces responses) and there are clients that do requests. Due to this,
+`request-response` would need `async` API to only act once there is request and
+once there is reply for it.
+
+## Requirements
+
+* **R1: Async API look and feel** \* The new `async` API shall provide the same
```
elfenpiff, line 133: We must add here one essential requirement. The implementation must be `no_std`. We are currently in the middle of porting iceoryx2 to `no_std`. If this feature cannot be implemented with `no_std` we need to introduce a feature flag that explicitly enables this functionality.

pawelrutkaq: For now I would say shall be doable as Waker, Context,Future, Pin are in core.

pawelrutkaq: added

doc/design-documents/drafts/async_api/async_api.md
pawelrutkaq: Tried, was looking horribly bad ;)

doc/design-documents/drafts/async_api/async_api.md
pawelrutkaq: added

doc/design-documents/drafts/async_api/assets/event_example.rs
pawelrutkaq: Not sure what that mean, we need to dicuss.

doc/design-documents/drafts/async_api/assets/event_example.rs
pawelrutkaq: Makse no sense really. Waitset is doing sleep (or what have you meant?) and will add new thread. The runtimes can do the same with one liner.

Also the runtimes has own IO multiplexers, so adding new one in seperate thread witll normally cause multiple thread hops to process a change instead processing in place by runtime. So I would not do it as default solution.

doc/design-documents/drafts/async_api/async_api.md
pawelrutkaq: As I said, I have different opinion. Imho iceoryx2 shall provide working, out of box solution for this. So user does not need to combine N things to make this works. Code would anyway be composed in such a way that user **might** do it manually if he wants (ie manually combining AsyncListener + Subscriber instead using AsyncSubscriber).

doc/design-documents/drafts/async_api/async_api.md
pawelrutkaq: At the end yes, Subscriber needs event source so it would have Listener.,

doc/design-documents/drafts/async_api/async_api.md
pawelrutkaq: 1. Yes
2. No, but it's interesting what have you meant ;) Second is that we would use event in same service but this would mean that in that service You can only have single messaging pattern as async. 
Added examples.

doc/design-documents/drafts/async_api/async_api.md
pawelrutkaq: I think without `AsyncSubscriber`, the Listener is anyway implemented outside iceoryx2, so we could only do some minor things here. `AsyncSubscriber` would pave us a way to others. Once we are happy with implementation, we can add missing part of this messaging pattern and make it officially supported for Pub-Sub. Then, we can continue with others incrementally.

doc/design-documents/drafts/async_api/async_api.md
pawelrutkaq: its rather me being bad at gramma (even in polish) ;)

doc/design-documents/drafts/async_api/assets/event_example.rs
pawelrutkaq: This does not matter I think. So we are doing `notify` for each sample produced. Whether this send event, or already have send event does not matter, and is even better since we would send less events. At the end, some event was send to listener which will wake up and process all until it has no more events, correct? ;)
Speaking in code it means `AsyncSubscriber` doing `receive().await`:
- check if there is sample, if there is, its done, next cal does same.
- if no samples, wait for notification
- repeat

Still i wonder what You will do, counter for each event value  ? At the end, it will mean less OS calls -> better preformance in both async and non async I think.

doc/design-documents/drafts/async_api/assets/event_example.rs
pawelrutkaq: We agreed to stick to a fact that `try_receive` will return `Ok(None)` only whne there would be `EWOULDBLOCK` error from os call in FD case.

doc/design-documents/drafts/async_api/assets/event_example.rs
pawelrutkaq: As we dicussed today: We can write specialized handling of API when events do use Semaphore or so. And the other option is what @elBoberido mentioned, once custom waitset will be there. This would be up to `Runtime` to choose an integration way for this.

doc/design-documents/drafts/async_api/assets/event_example.rs
pawelrutkaq: As dicussed: No issue

doc/design-documents/drafts/async_api/async_api.md
pawelrutkaq: As discussed: We will do async PubSub implementation in [userland](https://github.com/eclipse-iceoryx/iceoryx2/tree/main/iceoryx2-userland) . This crate hold the same quality and processes as whole code base and will be maintained.

doc/design-documents/drafts/async_api/async_api.md
pawelrutkaq: As dicussed: We **may** not need to do anything as per mentioned guarantess it will not block and it will do correct action (after reimplementation). I would leave still open for later Whether `AsyncPublisher` shall be created as proxy wrapper (or type alias) to `Publisher` to keep async API symetric only.

doc/design-documents/drafts/async_api/assets/event_example.rs
pawelrutkaq: As dicussed: Not a problem, some notification will happen underneath and this is enought to get waked in async. The implemented need to take care before waiting that it consumed all events/samples etc.

doc/design-documents/drafts/async_api/assets/event_example.rs
pawelrutkaq: As discussed: We may not need to do anything as per mentioned guarantee it will not block and it will do correct action (after reimplementation). I would leave still open for later Whether AsyncPublisher shall be created as proxy wrapper (or type alias) to Publisher to keep async API symetric only.

doc/design-documents/drafts/async_api/async_api.md
pawelrutkaq: We agreed to do `AsyncSubscriber` in userland.

doc/design-documents/drafts/async_api/async_api.md
elBoberido: Yeah, we forgot to mention that iceoryx2 is a toolbox with different layers and the iceoryx2 crate offers the low level building blocks with an un-opinionated API. I the `userland` crate we offer a more high-level but also more opinionated API. But it will be developed according to the same quality level than the more lower level crates.

doc/design-documents/drafts/async_api/async_api.md
elBoberido: As discussed. This would be part of the `userland` crates

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,400 @@
...
+##### Pros
+
+* no need to create dynamic event name
+
+##### Cons
+
+* limit a service to only single messaging pattern as using event will cause no
+  way to use it again
+
+Considering above, the continuation of the proposal is based on option 1. The
```
elBoberido, line 250: Considering option 1, one could also extend the Service Discovery with the service attributes. Please have a loot at https://github.com/eclipse-iceoryx/iceoryx2/tree/main/examples/rust/service_attributes

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,400 @@
...
+* Applicable standards (e.g., ASIL-D, ISO 26262)
+* Support for **zero-trust deployments**: can rogue processes break it?
+* Evidence for claims (e.g., subscribers cannot corrupt publisher-owned
+  read-only memory)
+* Real-time suitability: any blocking calls, background threads, or
+  indeterminism?
+* If unsuitable for zero-trust or real-time use, how do we prevent accidental
+  misuse?
+
+PR: To me idea, we will build async API on top of existing non-blocking, non
```
elBoberido, line 289: nitpicking: Just to have this a little bit more formal :)
```suggestion
PR: The author proposes to build the `async` API on top of existing non-blocking, non
```

doc/design-documents/drafts/async_api/async_api.md
```diff
@@ -0,0 +1,400 @@
...
+
+* TBD later
+
+**Results:**
+
+* User will get first support for async API for specific runtime. This will also
+  open a way for other to implement a bridge to other runtime like `tokio` as
+  basic idea will be shown
+
+### Milestone 2 – Implement PubSub
```
elBoberido, line 314:
```suggestion
### Milestone 3 – Implement PubSub
```

elBoberido (commented):
Just a little bit of nitpicking. Looks good. Waiting for @elfenpiff for his final review.

doc/design-documents/drafts/async_api/async_api.md
elfenpiff: Then lets stick with puml and svg

pawelrutkaq:
> Just a little bit of nitpicking. Looks good. Waiting for @elfenpiff for his final review.

Kk, will fix it once @elfenpiff review

pawelrutkaq:
@elfenpiff  ?

pathscale:
This would be great - I'm just catching up on this and curious what happened?

elfenpiff:
@pathscale We didn't find the time to pursue this on our side, but if you'd like to continue the work, we would be happy to support you.

I need to freshen up my knowledge on this PR. 

But the important need-to-know things for you are:

1. iceoryx2 is made for a safety-critical system, so we do not have background threads or blocking calls in general (with a few exceptions). Furthermore, we have an implementation of an even-multiplexer, called the [waitset](https://github.com/eclipse-iceoryx/iceoryx2/blob/main/iceoryx2/src/waitset.rs), which is comparable to [tokio's mio](https://github.com/tokio-rs/mio)

2. The only communication that can actually be blocked is the [listener port](https://github.com/eclipse-iceoryx/iceoryx2/blob/main/iceoryx2/src/port/listener.rs) of the event messaging pattern. So on the current API level, only this construct needs to be `async`ed.
We strictly separated data from control-flow, and this means that the other communication patterns like publish-subscribe, request-response, or blackboard are purely data-flow patterns, without any syscalls, and they never block.
If you want to achieve a behavior like my subscriber port will wake up whenever I receive a message, then you have to combine the subscriber port with the event messaging pattern. So on one side, you would have publisher and notifier port. First, you send a sample and then trigger a notification. The notification (control-flow) wakes up the other side, and then you know that you have received one or many samples. See the [event based communication example](https://github.com/eclipse-iceoryx/iceoryx2/tree/main/examples/rust/event_based_communication) to get the idea.

3. Due to this separation (data and control flow) and that iceoryx2 does not have background threads (is made for safety-critical and embedded systems), it seems to make sense to provide an async API just for the blocking listener calls in the iceoryx2 crate. But I agree that we also need a clean solution for less strict use cases. Here, the idea was to introduce high-level publisher/subscriber ports in the [iceoryx2 userland layer](https://github.com/eclipse-iceoryx/iceoryx2/tree/main/iceoryx2-userland) that again combine data and control flow. This could be a new crate called `iceoryx2-userland-async`. The combination of data- and control-flow would lead to a performance hit, but it would enable the user to use iceoryx2 completely with a pure async API.
So we would offer both API flavors, and the user can choose the API that fits their use case best.

dkroenke:
I'm closing this PR now due to inactivity, this does not mean that the proposal is rejected. When there is a demand or someone wants to work on it, we can reopen and continue.

PR #1015, opened 2025-09-06
[#1014] fix undefined behavior in user header

### LINKED ISSUE #1014: Undefined behavior in user header when loaning payload
When a payload is loaned in the publisher, client or server and a custom user header is set, the user header is never constructed. This causes undefined behavior. The contract should be that the `UserHeader` must implement `Default` and that iceoryx2 always writes the default values into the sample.

Later, one could also think of an extended API where the user can provide the user header value when loaning data, but this is out of scope for this bug fix.

elfenpiff:

* [x] Assign PR to reviewer

## References

Closes #1014

### DIFF
--- a/iceoryx2-cxx/include/iox2/active_request.hpp
+++ b/iceoryx2-cxx/include/iox2/active_request.hpp
@@ -13,6 +13,7 @@
 #ifndef IOX2_ACTIVE_REQUEST_HPP
 #define IOX2_ACTIVE_REQUEST_HPP
 
+#include "internal/helper.hpp"
 #include "iox2/internal/helper.hpp"
 #include "iox2/payload_info.hpp"
 #include "iox2/response_mut_uninit.hpp"
@@ -176,7 +177,7 @@ ActiveRequest<Service, RequestPayload, RequestUserHeader, ResponsePayload, Respo
     ResponseMutUninit<Service, ResponsePayload, ResponseUserHeader> response;
     auto result = iox2_active_request_loan_slice_uninit(
         &m_handle, &response.m_response.m_response, &response.m_response.m_handle, NUMBER_OF_ELEMENTS);
-    internal::EmplaceNew<ResponseUserHeader>::emplace(response);
+    internal::PlacementDefault<ResponseUserHeader>::placement_default(response);
     if (result == IOX2_OK) {
         return iox::ok(std::move(response));
     }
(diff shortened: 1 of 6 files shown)


iceoryx2-cxx/include/iox2/active_request.hpp
```diff
@@ -175,6 +176,7 @@ ActiveRequest<Service, RequestPayload, RequestUserHeader, ResponsePayload, Respo
     ResponseMutUninit<Service, ResponsePayload, ResponseUserHeader> response;
     auto result = iox2_active_request_loan_slice_uninit(
         &m_handle, &response.m_response.m_response, &response.m_response.m_handle, NUMBER_OF_ELEMENTS);
+    internal::EmplaceNew<ResponseUserHeader>::emplace(response);
```
elBoberido: This is not easy to understand what is going on. I had to look at the implementation. The naming is not ideal, since all the other `emplace` APIs take the actual data as parameter but here it's the reference to the underlying memory. Doesn't this more or less resemble `PlacementDefault`?

iceoryx2-cxx/include/iox2/internal/helper.hpp
```diff
@@ -0,0 +1,35 @@
...
+#define IOX2_SERVICE_INTERNAL_HELPER_HPP
+
+namespace iox2::internal {
+template <typename T>
+struct EmplaceNew {
+    template <typename S>
+    static void emplace(S& payload) {
+        new (&payload.user_header_mut()) T();
+    }
+};
```
elBoberido: Food for thought. I would suggest to not use the word `emplace` here, since no data is provided, but the reference to the memory. I would also propose to make this more generic by just expecting a `void*` as parameter and having only one generic parameter, e.g.

```cpp
template <typename T>
inline void placement_default(T* ptr) {
    new (ptr) T();
}

template <>
inline void placement_default(void*) {
}
```

This also shows via the signature that the dtor of the argument will not be called. The current implementation allows to call `emplace` multiple times and contrary to e.g. the `Optional` the dtor of the payload object is not called.

... oh, just realized that one would also need to implement `user_header_mut_ptr` in order to use the proposed API. I'll leave the comment here anyway.

elBoberido (commented):
Looks good. Just some food for thought regarding the naming and potential invalid usage of the `EmplaceNew` API

elBoberido (dismissed):

elBoberido (approved):

PR #1020, opened 2025-09-07
[#936] Create cmake install targets for iceoryx2-bb-cxx

elBoberido:
## Notes for Reviewer

This PR adds `cmake install` support for iceoryx2-bb-cxx

* [x] Assign PR to reviewer

## References

Relates to #936

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -499,7 +499,7 @@ jobs:
       - name: Print native libs of FFI target
         if: false # This step takes 1 to 2 minutes; only enable if there are linker issues with the FFI target
         run: |
-          cd iceoryx2-ffi/ffi
+          cd iceoryx2-ffi/c
           cargo rustc -q -- --print=native-static-libs
 
       - name: Build language bindings


iceoryx2-bb/cxx/CMakeLists.txt
```diff
@@ -29,9 +29,13 @@ include(Iceoryx2PlatformSettings)
...
 
+# TODO is this required? We install the full include directory anyway and
+# the natvis file has the build path in the iceoryx2-bb-cxxTargets.cmake file,
+# which probably will cause issues on Windows since I was not able to restrict
+# it to the BUILD_INTERFACE
 target_sources(iceoryx2-bb-containers-cxx INTERFACE
-    ${PROJECT_SOURCE_DIR}/include/iox2/container/config.hpp
-    ${PROJECT_SOURCE_DIR}/include/iox2/container/optional.hpp
+    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include/iox2/container/config.hpp>
+    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include/iox2/container/optional.hpp>
```
elBoberido: @ComicSansMS  we didn't use `target_sources` for header up until now. I'm not quire sure what's the intention behind this?

ComicSansMS: This is useful for IDEs that maintain project files independent of the physical file layout on disk, most prominently Visual Studio. If you don't list them here, they don't show up in those IDEs at all and it's a pain to edit them. The need to manually restrict to the build interface is a pain here, but one that fortunately goes away once we are able to utilize CMake 3.23's file sets.

elBoberido: How would this look like with file sets?

ComicSansMS: > How would this look like with file sets?

Something like this:
```cmake
target_sources(my_target
    INTERFACE
    FILE_SET HEADERS
    BASE_DIRS <what used to be the target_include_directories path>
    FILES
    <header files>
)

install(TARGETS my_target
    [...]
    FILE_SET HEADERS
)
```
This takes care of setting up IDEs correctly, setting the include directories for the build and packaging headers correctly for installation. It also propagates correctly to any exported package config scripts.

iceoryx2-bb/cxx/CMakeLists.txt
```diff
@@ -29,9 +29,13 @@ include(Iceoryx2PlatformSettings)
 add_library(iceoryx2-bb-containers-cxx INTERFACE)
 add_library(iceoryx2-bb-cxx::iceoryx2-bb-containers-cxx ALIAS iceoryx2-bb-containers-cxx)
 
+# TODO is this required? We install the full include directory anyway and
+# the natvis file has the build path in the iceoryx2-bb-cxxTargets.cmake file,
+# which probably will cause issues on Windows since I was not able to restrict
+# it to the BUILD_INTERFACE
```
ComicSansMS: Btw, what was the problem with the natvis file you are referring to here? If I wrap the natvis source in a `$<BUILD_INTERFACE:>` generator expression, it seems to work as expected for me.

iceoryx2-bb/cxx/CMakeLists.txt
elBoberido: Okay, I think we can assume that developer are using a newer version than cmake 3.22 and we need it only for deployment reasons. Therefore, a solution could be
```cmake
if(${CMAKE_VERSION} VERSION_GREATER_EQUAL 3.23)
  # solution with target_sources
else()
  # solution with target_include_directories
end()

Comment (ComicSansMS) [reply iceoryx2-bb/cxx/CMakeLists.txt]: Sure, that would work (although I would probably write that as `if(CMAKE_VERSION VERSION_GREATER_EQUAL ...)` because of CMake's weird variable expansion rules for `if`-commands), but what exactly is wrong with the solution as written?

File: iceoryx2-bb/cxx/CMakeLists.txt (hunk 2 above)
Comment (elBoberido) [reply iceoryx2-bb/cxx/CMakeLists.txt]: Okay, I need to give it another try. Was already late when I tried it. But Windows also did not complain about paths, so this should be fine anyway.

File: iceoryx2-bb/cxx/CMakeLists.txt (hunk 1 above)
Comment (elBoberido) [reply iceoryx2-bb/cxx/CMakeLists.txt]: I was just thinking if we can use a more ergonomic approach with CMake 3.23 while still have the option to build with CMake 3.22 for Yocto 4.0. If the ergonomics are not different, then it doesn't matter. I'm just unfamiliar with the target_sources approach for header and cannot tell if one or the other approach is less annoying in the long term.

Comment (elBoberido) [reply iceoryx2-bb/cxx/CMakeLists.txt]: Oh, just one question. Do you know if it would still be required to list all headers like
```cmake
target_sources(my_target
    INTERFACE
    FILE_SET HEADERS
    BASE_DIRS include/
    FILES
    iox2/my_public_header.hpp
)
```

ComicSansMS: > Oh, just one question. Do you know if it would still be required to list all headers like

Yes, absolutely. You want to list the headers individually and then also only the listed headers will become part of the install package. The generated `*Targets.cmake` then replicates the same `target_sources` command for the install tree. 

CMake philosophically wants to know about all files that participate in the build. I'm aware that this is not the most popular aspect of their design and you can get away somewhat well with ignoring that preference, but it is not how CMake wants to be used.

According to Vito Gamberini, Kitware currently plans to expand the file set mechanism in post-4.0 CMake to become the universal way of managing source files. Right now it only works for headers and C++20 module sources, but there are concepts how to extend it to ordinary cpp source files. Irrespective of whether that comes to be, I find the approach how it handles header files superior to the legacy approach, as it eliminates a lot of unintuitive quirks from the workflow.

elBoberido: I'm usually also in favor of not using globs but with the headers I'm a bit worried. I need to check whether we get some visible error when a header is missing from `target_sources`. If not, we might end up releasing something and do not notice a missing header since not all of them are used in our out of tree builds but only in the tests, which uses the `BUILD_INTERFACE`. In such a case I would be rather pragmatic and just do what we did in the past :sweat_smile:

elBoberido: Aaaargh. Either I overlooked something or they implemented the FILE_SET in the error prone way :/

Due to `BASE_DIRS`, the tests will find the headers but if a header is forgotten under `FILES`, it is not installed. If `BASE_DIR` is not set, the header are not found in the tests.

Maybe we are doing it also wrong with the tests and the tests should use an out-of-tree configuration but I've never seen such a setup anywhere. But at least it's an option to detect missing headers.

Additionally, we install the iceoryx header under a prefix and I also did not find a way to specify the prefix.

Anyway, not using `FILE_SET` does not cause the aforementioned issues at the moment. This means it is a problem of our future selves, once we update to CMake >= 3.23. They would have hated us anyway, so no need to be gentle.

iceoryx2-bb/cxx/CMakeLists.txt
elBoberido: You were right, it works. Does the `natvis` file need to be installed?

elBoberido:
> Looks generally good and a quick test by having an out-oft-tree build with `iceoryx2-bb/cxx/tests` shows no problem. It would be maybe good to have an automatic check against the installed files like we do with iceoryx2-ffi. Could be also something that can be done in a nightly job because these files are not touched often. I can add a CI for this in a separate PR.

Once we use the `iceoryx2-bb-cxx` package, it would be part of the out-of-tree build anyway ... or do you suggest to build the tests in an out of tree configuration? If yes, this should probably also be done for the C++ bindings.

iceoryx2-cmake-modules/CMakeLists.txt
```diff
@@ -41,7 +41,7 @@ set(${PROJECT_NAME}_DIR ${PROJECT_SOURCE_DIR}/cmake
 ########## set variables for export ##########
 #
 set(CMAKE_INSTALL_LIBDIR "lib" CACHE PATH "Library install directory")
-include(GNUInstallDirs)
+set(CMAKE_INSTALL_DATAROOTDIR "share" CACHE PATH "Data install directory")
```
elBoberido: @dkroenke @ComicSansMS I'm not quite sure if this is a good idea but when I did it, there was an issue with not specifying a language in the cmake file. This cmake project is also just for cmake modules and there are not actual targets. One could of course add a dummy target but I thought I could get away with it. I just noticed some cmake warnings when building the iceoryx2-cmake-modules package standalone.

Do you know it this can cause some trouble in the future or on some platforms?

iceoryx2-bb/cxx/CMakeLists.txt
ComicSansMS: No, it is not essential. I would leave that for a follow-up task, as there is a number of possible ways how to include it, if desired.

iceoryx2-cmake-modules/CMakeLists.txt
ComicSansMS: As a rule of thumb, if you find yourself writing to a `CMAKE_*` variable, you are probably doing something wrong. 😅

I don't quite understand what it is you are trying to do here.

dkroenke: @elBoberido I think it is ok here to add languages (`C` and `CXX` by default) when specifying the CMake project. With that we can easily add the `GNUInstallDirs` module without any warnings. If we may add later some other language we can extend the list. We should avoid to define too many of these `CMAKE_*` variables on our own and use the standards.
Additionally we could omit the `CMAKE_INSTALL_LIBDIR` on line 43 above because it gets overloaded from the GNUInstallDirs module.

elBoberido: @ComicSansMS true :sweat_smile: 

When I initially wrote that package I got some weird errors which I initially fixed by commenting out GNUInstallDirs but since it is now present, it seems I fixed/overshadowed the warning later on. Anyway, removing `LANGUAGES NONE` also fixes the issue.

dkroenke (approved):

Issue #989, opened 2025-08-29
Request Response Advanced Connection Handling


elfenpiff:
## Overview

1. When a client or a server dies the cleanup does not set the zero copy connection state to disconnected.
2. The user shall be able to communicate the intend to terminate the connection from the server and the client side and also check if the other side has signaled the intend to disconnect so that the last package can be sent/received.

We suggest the following usage:
1. Side A communicates the intend to disconnect
2. Side B sends the last data and then let the corresponding sender go out of scope
3. Side A checks if side B is disconnected and then receives the last data and goes out of scope as well.

## Implementation

1. [ ] On cleanup, set the disconnect flag in the zero copy connection when a port is forcefully removed.
2. [ ] The `Server` and `Client` get 2 new methods to signal the intend to disconnect and to check for the state of the connection (maybe extend/rename the is_connected method).
3. [ ] Add C, C++, Python language bindings for the feature

PR #1028, opened 2025-09-09
[#989] advanced connection handling

elfenpiff:

* [x] Assign PR to reviewer

## References

Relates #989

### DIFF
--- a/iceoryx2/src/port/details/channel_management.rs
+++ b/iceoryx2/src/port/details/channel_management.rs
@@ -57,7 +57,7 @@ pub(crate) trait ChannelManagement: ZeroCopyPortDetails {
             Ordering::Relaxed,
             Ordering::Relaxed,
         ) {
-            Ok(_) => return,
+            Ok(_) => (),
             Err(v) => {
                 let graceful_disconnect_state = expected_state | (1u64 << 63);
                 if v == graceful_disconnect_state {
--- a/iceoryx2/src/port/details/channel_management.rs
+++ b/iceoryx2/src/port/details/channel_management.rs
@@ -15,6 +15,7 @@ use core::sync::atomic::Ordering;
 use iceoryx2_cal::zero_copy_connection::{ChannelId, ZeroCopyPortDetails};
 
 pub(crate) const INVALID_CHANNEL_STATE: u64 = u64::MAX;
+const GRACEFUL_DISCONNECT_BIT: u64 = 1u64 << 63;
 
 pub(crate) trait ChannelManagement: ZeroCopyPortDetails {
     fn set_channel_state(&self, channel_id: ChannelId, state: u64) -> bool {
@@ -29,7 +30,7 @@ pub(crate) trait ChannelManagement: ZeroCopyPortDetails {
     }
 
     fn set_channel_to_graceful_disconnect(&self, channel_id: ChannelId, expected_state: u64) {
-        let graceful_disconnect_state = expected_state | (1u64 << 63);
+        let graceful_disconnect_state = expected_state | GRACEFUL_DISCONNECT_BIT;
 
         let _ = self.channel_state(channel_id).compare_exchange(
             expected_state,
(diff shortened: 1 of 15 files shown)


iceoryx2/src/port/details/channel_management.rs
```diff
@@ -28,17 +29,48 @@ pub(crate) trait ChannelManagement: ZeroCopyPortDetails {
...
+            graceful_disconnect_state,
+            Ordering::Relaxed,
+            Ordering::Relaxed,
+        );
+    }
+
+    fn in_graceful_disconnect_state(&self, channel_id: ChannelId, expected_state: u64) -> bool {
+        let graceful_disconnect_state = expected_state | GRACEFUL_DISCONNECT_BIT;
+        graceful_disconnect_state == self.channel_state(channel_id).load(Ordering::Relaxed)
+    }
```
elBoberido: This pattern is a bit unfamiliar and can be simplified. No need to cary the `expected_state`
```suggestion
    fn in_graceful_disconnect_state(&self, channel_id: ChannelId) -> bool {
        let state = self.channel_state(channel_id).load(Ordering::Relaxed);
        GRACEFUL_DISCONNECT_BIT == state & !GRACEFUL_DISCONNECT_BIT
    }
```

Also, the current implementation is racy. In case someone changes the state concurrently, the function will return false, even if the bit is set.

elBoberido (commented):
Looks good. There is a potential race in `in_graceful_disconnect_state` that can easily be fixed.

I would also suggest to remove the `graceful` part of function names. For me it implies that the other method of disconnecting is bad. It can be kept in the documentation or be replaced with `explicit`/`explicitly`. Not a must, though.

iceoryx2/src/port/details/channel_management.rs
elfenpiff: @elBoberido Yes you are right that this works without the expected_state. But the reason why this is in there is, that the user must be aware what state they are changing to avoid any kind of race.

For instance, the `expected_state` contains the request id of the current request that uses the channel. If the channel is reused for another request id and the user sets it to invalid state but actually wanted to do this for the previous request id, we would have a race.

This is a race that can never happen in Rust since the lifetime and ownerships prevent this and we could go with your approach. But sadly, this is a realistic race in C++ when the user misuses the API and violates the lifetime constraints. Therefore, I decided for the more robust approach.

You are right, this is again undocument and implicit knowledge. We have to write more documentation.

elfenpiff: @elBoberido 

> Also, the current implementation is racy. In case someone changes the state concurrently, the function will return false, even if the bit is set.

In my opinion is it as racy as asking a concurrent container how many elements it has. As soon as you get the information it is out-of-date and means nothing.

In this case, it is the same. Let's assume someone changes the state between line 44 and 45 to graceful disconnect. Then the function would seemingly "wrongly" return false despite that this is true. But the user cannot observe the difference to the case where it was set to graceful disconnect right after the function returned. So somehow following the "as-if" rule, either race or valid behavior from outside the user cannot tell the difference and making the function race free would not solve the observed behavior of the user where they:
1. Acquire the state
2. Now concurrently the state changes
3. Check if the state has a specific value (already outdated)

elBoberido: From my point of view, this is quite error prone. Please correct me if I'm wrong, but
* there are concurrent writer and reader on the state
* the graceful disconnect bit is set on the client side and read/reset on the server side

What now can happen is that the client set's the bit and depending on other writers to the state, there is a chance the server will never notice that the bit is set, since the function does not only check for the bit but also for other invariants (and potentially also non-invariants) and just fails silently.

We can keep it for the v0.7 release but I strongly suggest to change the behavior and mask out all the bits that are irrelevant for the check, i.e. keeping only the graceful bit and the request id. The question then is what to do when the request id changed? Ignore it, since it doesn't matter or use a Result as return value?

elBoberido (dismissed):

elBoberido (approved):

Issue #347, opened 2024-08-06
Fully enable FreeBSD CI


elBoberido:
## Brief feature description

Due to disk space limitations on the github CI, the FreeBSD jobs do not build the examples, neither in-tree nor out-of-tree.

The root cause needs to be found and the build of the examples need to be re-enabled.

elBoberido:
@dkroenke fyi: I guess once #1302 is properly fixed, we should be able to enable the FreeBSD builds again

Issue #950, opened 2025-08-14
Bump minimal cmake version to 3.24


elBoberido:
## Brief

From https://github.com/eclipse-iceoryx/iceoryx2/pull/946#issue-3316914376

>  There is further potential for cleaning up these scripts if we decide to bump the minimum required version in the future:
>
>   * 3.23 introduces file sets for `target_sources`
>   * 3.24 introduces the `COMPILE_WARNING_AS_ERROR` target property

elBoberido:
@ComicSansMS there were some requests to support Yocto 4.0 Kirkstone, which has cmake 3.22. Yocto 4.1 has support for cmake 3.24. Let me check if 4.1 is also an options for the guys wo asked for 4.0.

PR #1052, opened 2025-09-16
[#969] Add code coverage for C++ code

### LINKED ISSUE #969: Measure code coverage for C and C++ sources
## Brief feature description

Currently, no code coverage is obtained for iceoryx2-bb-cxx and the C and C++ language bindings. Add an option to the build system that automatically generates code coverage data from executed unit tests and adjust the CI configuration so that this data is uploaded to Codecov along with the existing coverage report.

## Detailed information

--

dkroenke:
## Notes for Reviewer

- Extend script for coverage generation
- Add coverage compile flags to C++ targets
- Exclude iceoryx_hoofs from coverage generation

There is no change in usage. The coverage will be added to the existing one for Rust.
For local coverage generation execute `./internal/scripts/generate-cov-report.sh --generate`

* [x] ~Changelog updated [in the unreleased section][changelog] including API breaking changes~
* [x] Assign PR to reviewer

## References

Closes #969
Realtes #347 
Relates #3


iceoryx2-cxx/cmake/fetch-iceoryx-hoofs.cmake
```diff
@@ -46,10 +46,21 @@ if(ICEORYX_WITH_FETCH_CONTENT)
     set(EXAMPLES OFF)
     set(BUILD_TEST OFF)
 
+    # We need to disable temporary the ${COVERAGE} variable here to avoid interference
+    # because it is a external dependency
+    if(COVERAGE)
+        set(IOX2_COVERAGE_PREVIOUSLY_DISABLED ON)     
+        set(COVERAGE OFF CACHE INTERNAL "")
+    endif()
+
```
elBoberido, line 55: Did you try to add `target/ff/iceoryx/` to the ignore list instead of temporally deactivating the coverage?

internal/scripts/generate-cov-report.sh
```diff
@@ -21,17 +21,21 @@ COLOR_YELLOW='\033[1;33m'
...
+        echo -e "${COLOR_GREEN}rust nightly compiler found, activating MC/DC coverage check${COLOR_OFF}"
+        export RUSTFLAGS="-C instrument-coverage -Z coverage-options=branch"
+    else
+        echo -e "${COLOR_YELLOW}no rust nightly compiler found, MC/DC coverage is not available only line coverage${COLOR_OFF}"
+        export RUSTFLAGS="-C instrument-coverage"
+    fi
+}
 
 COVERAGE_DIR="target/debug/coverage"
+CMAKE_BUILD_DIR=build
```
elBoberido: Please create a folder in the target directory, e.g. `target/ff/cc/coverage`. We also build the C and C++ libraries into this directory in order to have to delete only one directory to cleanup everything.

Unfortunately, the Rust coverage results are stored in target directories alongside the crates.

internal/scripts/generate-cov-report.sh
```diff
@@ -21,17 +21,21 @@ COLOR_YELLOW='\033[1;33m'
...
-    echo -e "${COLOR_YELLOW}no rust nightly compiler found, MC/DC coverage is not available only line coverage${COLOR_OFF}"
-    export RUSTFLAGS="-C instrument-coverage"
-fi
+export DEBUGINFOD_URLS=/dev/null
+
+set_rustc_flags() {
+    if [[ "$(rustc --version | grep nightly | wc -l)" == "1" ]]
+    then
+        echo -e "${COLOR_GREEN}rust nightly compiler found, activating MC/DC coverage check${COLOR_OFF}"
+        export RUSTFLAGS="-C instrument-coverage -Z coverage-options=branch"
```
elBoberido: Can you crate an issue to use the nightly compiler to get MC/DC coverage?

internal/scripts/generate-cov-report.sh
```diff
@@ -136,6 +152,7 @@ generate_html_report() {
           --ignore "**/iceoryx2-ffi/*" \
           --ignore "**/build.rs" \
           --ignore "**/tests/*" \
+          --ignore "**/_deps/*" \
```
elBoberido: Out of curiosity, shouldn't this already be handled by `**/target/*`? ... oh, I have a suspicion. Is the `RUST_BUILD_ARTIFACT_PATH` used when building the tests for the coverage?

internal/scripts/generate-cov-report.sh
```diff
@@ -136,6 +152,7 @@ generate_html_report() {
           --ignore "**/iceoryx2-ffi/*" \
           --ignore "**/build.rs" \
           --ignore "**/tests/*" \
```
elBoberido: For some reasons, the codecove upload also shows test files, e.g. `iceoryx2-cxx/tests/src/slice_tests.cpp` or `iceoryx2-bb/cxx/tests/src/testing/test_utils.hpp`. At lest the tests in `iceoryx2-cxx` should be filtered out. Can you check why it does not happen? Do the ignores for C++ need to be relative to the build folder?

iceoryx2-cmake-modules/platform/generic/modules/Iceoryx2PlatformSettings.cmake
```diff
@@ -39,3 +39,13 @@ set(ICEORYX2_SANITZER_FLAGS CACHE INTERNAL "")
...
 endif()
+
+set(ICEORYX2_COVERAGE_FLAGS CACHE INTERNAL "")
+if(COVERAGE)
+    set(CMAKE_CXX_OUTPUT_EXTENSION_REPLACE 1)
+    if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+        set(ICEORYX2_COVERAGE_FLAGS -g -O0 --coverage CACHE INTERNAL "")
+    else(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+        message( FATAL_ERROR "You need to run gcov with gcc compiler." )
+    endif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
```
elBoberido:
```suggestion
    else()
        message( FATAL_ERROR "You need to run gcov with gcc compiler." )
    endif()
```

To be consistent with the style in the file :)

iceoryx2-cmake-modules/modules/Iceoryx2CommonOptionsAndParams.cmake
```diff
@@ -17,12 +17,22 @@ if(NOT ICEORYX2_COMMON_OPTIONS_AND_PARAMS_LISTED)
 
     message(STATUS "[i] iceoryx2 common options and params:")
 
+if(COVERAGE)
+    set(BUILD_TESTING ON CACHE INTERNAL "")
+endif()
+
```
elBoberido: I know did it this way in iceoryx classic but up until now, we tried to avoid setting options that the user did not explicitly set. What do you think about printing an error if `BUILD_TESTING` is not set instead of setting it?

internal/scripts/generate-cov-report.sh
dkroenke: Done: https://github.com/eclipse-iceoryx/iceoryx2/issues/1054

internal/scripts/generate-cov-report.sh
dkroenke: At least it's not explicitly set in the script and the CMake lot output shows its empty

internal/scripts/generate-cov-report.sh
```diff
@@ -21,17 +21,21 @@ COLOR_YELLOW='\033[1;33m'
...
+    then
+        echo -e "${COLOR_GREEN}rust nightly compiler found, activating MC/DC coverage check${COLOR_OFF}"
+        export RUSTFLAGS="-C instrument-coverage -Z coverage-options=condition" # Todo (iox2-#1052) use mcdc here
+    else
+        echo -e "${COLOR_YELLOW}no rust nightly compiler found, MC/DC coverage is not available only line coverage${COLOR_OFF}"
+        export RUSTFLAGS="-C instrument-coverage"
+    fi
+}
 
 COVERAGE_DIR="target/debug/coverage"
```
elBoberido:
```suggestion
RUST_COV_DIR="target/debug/coverage"
```

Only if there are other commits.

internal/scripts/generate-cov-report.sh
```diff
@@ -153,7 +171,7 @@ generate_lcov_report() {
     grcov \
           **/${LLVM_PROFILE_PATH} \
           **/**/${LLVM_PROFILE_PATH} \
-          --binary-path ./target/debug \
+          --binary-path ./target/debug ./build \
```
elBoberido: To match `generate_html_report`

```suggestion
          --binary-path ./target/debug ${CMAKE_COV_DIR} \
```

iceoryx2-cxx/cmake/fetch-iceoryx-hoofs.cmake
```diff
@@ -46,10 +46,21 @@ if(ICEORYX_WITH_FETCH_CONTENT)
     set(EXAMPLES OFF)
     set(BUILD_TEST OFF)
 
+    # We need to disable temporary the ${COVERAGE} variable here to avoid interference
+    # because it is a external dependency
+    if(COVERAGE)
+        set(IOX2_COVERAGE_PREVIOUSLY_DISABLED ON)     
+        set(COVERAGE OFF CACHE INTERNAL "")
+    endif()
```
ComicSansMS, line 54: Unless the subproject explicitly reads from the cache, it should be sufficient here to just set a new local variable `COVERAGE` which shadows the cache option. Writing directly to the cache is almost always a bad idea, as it may lose information put there by the user.

With CMake 3.25 you can combine this with the `block()` command for the rather pleasant looking
```cmake
block()
  set(COVERAGE OFF)
  add_subdirectory(...)
endblock()
```
[Example on godbolt](https://godbolt.org/z/rTbM6b5sz).

iceoryx2-cmake-modules/platform/generic/modules/Iceoryx2PlatformSettings.cmake
```diff
@@ -39,3 +39,13 @@ set(ICEORYX2_SANITZER_FLAGS CACHE INTERNAL "")
 if(SANITIZERS)
     include(Iceoryx2Sanitizer)
 endif()
+
+set(ICEORYX2_COVERAGE_FLAGS CACHE INTERNAL "")
+if(COVERAGE)
+    set(CMAKE_CXX_OUTPUT_EXTENSION_REPLACE 1)
+    if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+        set(ICEORYX2_COVERAGE_FLAGS -g -O0 --coverage -fprofile-abs-path CACHE INTERNAL "")
```
ComicSansMS: Food for thought: Instead of turning every configuration into a debug build when switching on coverage, how about only allowing the coverage option for the debug config?

iceoryx2-cxx/cmake/fetch-iceoryx-hoofs.cmake
dkroenke: It is now the `block()` syntax used as in https://github.com/eclipse-iceoryx/iceoryx2/pull/1052#discussion_r2362008441 proposed.

ComicSansMS: Wait, `block` requires CMake 3.25, we are still [stuck on 3.22 for the time being](https://github.com/eclipse-iceoryx/iceoryx2/issues/950). Sorry, I should have made this more clear in the comment.

internal/scripts/generate-cov-report.sh
dkroenke: It seems that Codecov ignore the exclusions set in `gcovr`. The [html report](https://github.com/eclipse-iceoryx/iceoryx2/actions/runs/17862796661/artifacts/4069727877) does not show tests folders here. I needed to add them to the ignore list in `codecov.yaml`.

.github/workflows/build-test.yml
```diff
@@ -730,15 +737,26 @@ jobs:
       uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 # ratchet:actions/upload-artifact@v4
       with:
         name: coverage-html
-        path: target/debug/coverage/html/*
+        path: target/coverage/html/*
```
elBoberido: We try to put everything not cargo related to `target/ff` (ff = foreign functions) in order to reduce the chance to interfere with future cargo features, which might use the same folders. Would this also work with `target/ff/coverage/html` also work?

internal/scripts/generate-cov-report.sh
```diff
@@ -245,9 +244,19 @@ if [[ $REPORT == "1" ]]; then
 fi
 
 if [[ $LCOV == "1" ]]; then
-    generate_lcov_report
+    mkdir -p ${COVERAGE_OUT_DIR}/lcov
+    OUTPUT_TYPE=lcov
+    RUST_COVERAGE_OUT=${COVERAGE_OUT_DIR}/lcov/iceoryx2_lcov_rust.info
+    CPP_COVERAGE_OUT="--lcov ${COVERAGE_OUT_DIR}/lcov/iceoryx2_lcov_cpp.info"
+    generate_report
```
elBoberido: Food for thought. Would it make sense to add the three parameter `OUTPUT_TYPE`, `RUST_COVERAGE_OUT` and `CPP_COVERAGE_OUT` as parameter to the function instead of setting the global variables?

Just a question. I've the feeling though, that a larger refactoring would be needed to make this possible.

iceoryx2-cxx/cmake/fetch-iceoryx-hoofs.cmake
```diff
@@ -46,9 +46,14 @@ if(ICEORYX_WITH_FETCH_CONTENT)
     set(EXAMPLES OFF)
     set(BUILD_TEST OFF)
 
-    # use iceoryx platform and hoofs in source code version
-    add_subdirectory(${iceoryx_SOURCE_DIR}/iceoryx_platform  ${iceoryx_BINARY_DIR}/iceoryx_platform EXCLUDE_FROM_ALL)
-    add_subdirectory(${iceoryx_SOURCE_DIR}/iceoryx_hoofs ${iceoryx_BINARY_DIR}/iceoryx_hoofs EXCLUDE_FROM_ALL)
+    # We need to disable temporary the ${COVERAGE} variable here to avoid interference
+    # because it is a external dependency
+    block()
```
elBoberido: Please remove the block if it is only available with CMake 3.25. Unfortunately, we need to support CMake 3.22.

internal/scripts/generate-cov-report.sh
```diff
@@ -245,9 +244,19 @@ if [[ $REPORT == "1" ]]; then
...
 fi
 
 if [[ $HTML == "1" ]]; then
-    generate_html_report
+    mkdir -p $COVERAGE_OUT_DIR/html/cpp
+    OUTPUT_TYPE=html
+    RUST_COVERAGE_OUT=${COVERAGE_OUT_DIR}/html/rust
+    CPP_COVERAGE_OUT="--html-details ${COVERAGE_OUT_DIR}/html/cpp/index.html"
+    generate_report
```
elBoberido: These two lines are missing
```
    sed -i 's/coverage/grcov/' ${COVERAGE_DIR}/html/coverage.json
    sed -i 's/coverage/grcov/' ${COVERAGE_DIR}/html/badges/*.svg
```

Did you check whether this was fixed upstream in grcov? IIRC, the generated html was subtly broken without that change.

internal/scripts/generate-cov-report.sh
```diff
@@ -120,57 +136,40 @@ show_report() {
...
-          --ignore "**/.cargo/*" \
-          --llvm-path ${LLVM_PATH} \
-          --output-path ./${COVERAGE_DIR}/lcov.info
-}
-
+        **/${LLVM_PROFILE_PATH} \
+        **/**/${LLVM_PROFILE_PATH} \
+        --binary-path target/debug \
+        --source-dir . \
+        --output-type $1 \
```
elBoberido: For maintainability, can you add something like this at the top of the function

```
LOCAL_FOO=$1
```

Same for $2 and $3.

elBoberido (commented):
Can you also try to fully deactivate the FreeBSD build for Rust 1.81?
Maybe also the 'stable' FreeBSD on 'main' for the time being?

internal/scripts/generate-cov-report.sh
```diff
@@ -120,57 +136,43 @@ show_report() {
     eval $CMD
 }
 
-generate_html_report() {
+# Signature: generate_report(OUTPUT_TYPE, RUST_COVERAGE_OUT, CPP_COVERAGE_OUT)
+generate_report() {
+    local OUTPUT_TYPE=$1
+    local GRCOV_OUTPUT_PATH=$2
+    local GCOVR_OUTPUT_PATH=$3
```
elBoberido, line 143: Which evil genius named these two tools like that 🤣

.github/workflows/build-test.yml
```diff
@@ -613,28 +613,29 @@ jobs:
...
-          copyback: false
-          prepare: pkg install -y cmake curl gcc git llvm
-          run: |
-            git config --global --add safe.directory /home/runner/work/iceoryx2/iceoryx2
-            ./internal/scripts/ci_prepare_freebsd.sh
-            ./internal/scripts/ci_build_and_test_freebsd.sh --toolchain ${{ matrix.toolchain }} --mode ${{ matrix.mode }}
+  # todo(#347): Reenable FreeBSD once it got stabilized and update the config
+  # freebsd:
+  #   needs: [preflight-check, static-code-analysis]
+  #   if: ${{ needs.changes.outputs.source-code == 'true' }}
```
elBoberido, line 619: Friendly hint for the next time :)

Using just `if: false &&` preserves a meaningful `git blame` for the other lines and reduces the risk of merge conflicts with other PRs which might have also added changes to this sections ... I know, we have set bad precedence with the section right above

elBoberido (dismissed):

.github/workflows/build-test.yml
```diff
@@ -730,15 +738,34 @@ jobs:
...
         token: ${{ secrets.CODECOV_TOKEN }}
+
+
+  # Have a sync point for all CI jobs that need to pass, will be enforced by Otterdog
+  mandatory-checks:
+    needs: [
+            static-code-analysis,
+            x86_64, x86_64_bazel,
+            x86_32,
+            coverage,
```
elBoberido: What do you think about removing coverage from the list. Somethimes there were issues with the upload which were resolved later by codecove. 

The tunnel end-to-end tests might also be a candidate for a non-mandatory test. 

The lint-markdown should go in, though. 

I can do this in my open PRif you agree.

elBoberido (dismissed):

.github/workflows/build-test.yml
dkroenke: I agree removing coverage and the tunnel end-to-end tests. I'm not sure how it reacts when a check got skipped like `lint-markdown` but we can test it out here.

I will make a commit here.

elBoberido (approved):

elBoberido (approved):

PR #1058, opened 2025-09-18
[#709] Add initial set of end to end tests for CLI

### LINKED ISSUE #709: Add end-to-end tests for `iox2` CLI
## Brief Description

End-to-end tests for that CLI should be added to ensure changes to not break functionality.

orecham:
## Notes for Reviewer

Adds initial set of end-to-end tests for `iceoryx2-cli`.

Not all CLI functionality is covered, just the most straight-forward. Provides the foundation for adding more complex test cases in follow-ups.

Additionally:

1. A number of bugs that were discovered have been fixed:
    1. Print newline character after console output to prevent odd artifacts from being inserted by the terminal
    2. Print help for positional arguments
    3. Properly organize provided CLI options in `iox2-tunnel`
2. Some new features were added to facilitate testing and and some more QoL:
    1. Option to print summarized or detailed information of discovered services in `iox2 service discovery`
    2. Option to force overwrite in `iox2 config generate`
3. Each command was organized into their own modules for cleanliness
    1. Some commands are complex and utilize multiple helper functions
    2. Same convention applied to all commands (even the most simple ones) for consistency 

* [ ] Assign PR to reviewer

## References

Closes #709


iceoryx2-testing/end-to-end-testing/common.exp
orecham, line 1: Not sure if this is the best place to put this common script now that it is used outside of the examples directory ... accepting ideas.

elBoberido: We also did not like it when we created it but we had no better idea where to place it. Suggestions welcome.

iceoryx2/src/config.rs
```diff
@@ -126,7 +126,6 @@ impl core::error::Error for ConfigCreationError {}
 /// All configurable settings of a [`Service`](crate::service::Service).
 #[non_exhaustive]
 #[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq)]
-#[serde(rename_all = "kebab-case")]
```
orecham, line 132: @elBoberido FYI I would like to remove this from our structs because it causes the RON output to prepend `#r` before every field (since it is not valid Rust syntax):

```
        global: (
            r#root-path: "/tmp/iceoryx2/",
            prefix: "iox2_",
            service: (
                directory: "services",
                r#data-segment-suffix: ".data",
                r#static-config-storage-suffix: ".service",
                r#dynamic-config-storage-suffix: ".dynamic",
                r#creation-timeout: (
                    secs: 0,
                    nanos: 500000000,
                ),
                r#connection-suffix: ".connection",
                r#event-connection-suffix: ".event",
                r#blackboard-mgmt-suffix: ".blackboard_mgmt",
                r#blackboard-data-suffix: ".blackboard_data",
            ),
            node: (
                directory: "nodes",
                r#monitor-suffix: ".node_monitor",
                r#static-config-suffix: ".details",
                r#service-tag-suffix: ".service_tag",
                r#cleanup-dead-nodes-on-creation: true,
                r#cleanup-dead-nodes-on-destruction: true,
            ),
        )
```

iceoryx2-testing/end-to-end-testing/common.exp
orecham: @elfenpiff Thoughts on creating this top-level `iceoryx2-testing` directory for utility scripts like this? I was also thinking to move the types used in examples to here too so they can be re-used in tests (e.g. the `ComplexData` type, which I am currently duplicating in the `iceoryx2-tunnel` test.

`iceoryx2-bb-testing` did not seem like the right place for these to me. The nesting in `iceoryx2-bb` does not seem fitting.

iceoryx2/src/config.rs
elBoberido: I wouldn't do that. It would completely break every config setup out there.

iceoryx2-testing/end-to-end-testing/common.exp
elBoberido: @orecham in theory, this could as well go into `internal/something`. It is currently only meant to be used in the repo

iceoryx2/src/config.rs
orecham: @elBoberido Hmm, if we ever want to change it, now (before `v1.0.0`) would be the time. Otherwise we need to live with this not-so-nice output from the default formater, which is currently `ron`.

We could of course change the default format to something else... although `ron` seems the most appropriate, given we the core is Rust.

elBoberido: Isn't this just for testing? Shouldn't the default formatter be TOML, which is also used for the config files

orecham: @elBoberido Not everything output by the CLI is a config. The format can be selected by the user using the `--format` option - the motivation here is that users can choose what they prefer and if they are parsing the output programatically specify the output (e.g. if they decide they prefer to parse it as JSON). For instance they might want to parse the output of `iox2 service subscribe` from a script or other application.

The default format was chosen kind of arbitrarily, but the train of thought was to use `ron` since it is a Rust project and besides the `r#` it seemed to be the most human-readable (other output formats deal with`ValueEnums` kind of strangely), but only Ron has this not-so-nice-to-read `r#` before strings.

Moving forward, we either:
1. Leave it as is and just live with the `r#` being  prepended for `ron`
2. Change the default formatting to something else (although users will still be able to choose something else)
3. Remove the `kebab-case` modification to the output

iceoryx2-testing/end-to-end-testing/common.exp
orecham: @elBoberido I moved the script to `internal/end-to-end-testing/`.

iceoryx2/src/config.rs
elBoberido: @orecham I'm not sure if I have the full picture :sweat_smile: 

The changes you are proposing are only for the config, which is currently TOML and editable by the user. Sure, for listing of the services, etc. something like RON makes more sense.

Where exactly comes the user into contact with the config and the output in RON?

From my subjective impression, it would not be worth to break the existing user config for this.

orecham: @elBoberido 

Maybe it is easier to illustrate when describing the user perspectives I am considering. Since my last comment, I have noticed that the CLI parsing perspective has an additional problem.

1. Consider a user reading the CLI output
    1. The prepended `'r#` is not so pretty
    1. The output is inconsistent, some outputs have fields with `kebab-case` and others with `camel_case` 
    1. In this case of just reading the output it is not too much of a big deal, it doesn't looks so nice but maybe we can live with it
        1. The not-so-pretty output is the original perspective I was thinking, but it is not so strong, I have even already reverted it on this branch, but unfortunately thought of something else we should consider
1. Consider a user writing a program (or script) to parse CLI output
    1. Here, the RON output is likely not to be used, so the `r#` is less of an issue
    2. More commonly we can expect JSON format being parsed
    3. Now we have some fields using `snake_case` and some fields using `kebab-case`
    4. If the user knows this ahead of time, they can work with this, but I would assume it would trip some users up initially

<details>
<summary>Current CLI Outputs</summary>

```
$ cargo run --bin iox2-node -- details 0000000018fdd2ed000043be0003f503 --format JSON
{
  "state": "Alive",
  "id": "0000000018fdd2ed000043be0003f503",
  "pid": 259331,
  "executable": "publish_subscribe_subscriber",
  "name": "",
  "config": {
    "global": {
      "root-path": "/tmp/iceoryx2/",
      "prefix": "iox2_",
      "service": {
        "directory": "services",
        "data-segment-suffix": ".data",
        "static-config-storage-suffix": ".service",
        "dynamic-config-storage-suffix": ".dynamic",
        "creation-timeout": {
          "secs": 0,
          "nanos": 500000000
        },
        "connection-suffix": ".connection",
        "event-connection-suffix": ".event",
        "blackboard-mgmt-suffix": ".blackboard_mgmt",
        "blackboard-data-suffix": ".blackboard_data"
      },
      "node": {
        "directory": "nodes",
        "monitor-suffix": ".node_monitor",
        "static-config-suffix": ".details",
        "service-tag-suffix": ".service_tag",
        "cleanup-dead-nodes-on-creation": true,
        "cleanup-dead-nodes-on-destruction": true
      }
    },
    "defaults": {
      "publish-subscribe": {
        "max-subscribers": 8,
        "max-publishers": 2,
        "max-nodes": 20,
        "subscriber-max-buffer-size": 2,
        "subscriber-max-borrowed-samples": 2,
        "publisher-max-loaned-samples": 2,
        "publisher-history-size": 0,
        "enable-safe-overflow": true,
        "unable-to-deliver-strategy": "Block",
        "subscriber-expired-connection-buffer": 128
      },
      "event": {
        "max-listeners": 16,
        "max-notifiers": 16,
        "max-nodes": 36,
        "event-id-max-value": 255,
        "deadline": null,
        "notifier-created-event": null,
        "notifier-dropped-event": null,
        "notifier-dead-event": null
      },
      "request-response": {
        "enable-safe-overflow-for-requests": true,
        "enable-safe-overflow-for-responses": true,
        "max-active-requests-per-client": 4,
        "max-response-buffer-size": 2,
        "max-servers": 2,
        "max-clients": 8,
        "max-nodes": 20,
        "max-borrowed-responses-per-pending-response": 2,
        "max-loaned-requests": 2,
        "server-max-loaned-responses-per-request": 2,
        "client-unable-to-deliver-strategy": "Block",
        "server-unable-to-deliver-strategy": "Block",
        "client-expired-connection-buffer": 128,
        "enable-fire-and-forget-requests": true,
        "server-expired-connection-buffer": 128
      },
      "blackboard": {
        "max-readers": 8,
        "max-nodes": 20
      }
    }
  }
}

$ cargo run --bin iox2-service -- details "My/Funk/ServiceName" --format JSON
{
  "service_id": "4eacadf2695a3f4b2eb95485759246ce1a2aa906",
  "service_name": "My/Funk/ServiceName",
  "attributes": [],
  "pattern": {
    "PublishSubscribe": {
      "max_subscribers": 8,
      "max_publishers": 2,
      "max_nodes": 20,
      "history_size": 0,
      "subscriber_max_buffer_size": 2,
      "subscriber_max_borrowed_samples": 2,
      "enable_safe_overflow": true,
      "message_type_details": {
        "header": {
          "variant": "FixedSize",
          "type_name": "iceoryx2::service::header::publish_subscribe::Header",
          "size": 40,
          "alignment": 8
        },
        "user_header": {
          "variant": "FixedSize",
          "type_name": "()",
          "size": 0,
          "alignment": 1
        },
        "payload": {
          "variant": "FixedSize",
          "type_name": "TransmissionData",
          "size": 16,
          "alignment": 8
        }
      }
    }
  },
  "nodes": {
    "num": 1,
    "details": [
      {
        "state": "Alive",
        "id": "0000000018fdd2ed000043be0003f503",
        "pid": 259331,
        "executable": "publish_subscribe_subscriber",
        "name": ""
      }
    ]
  }
}
```

</details>

I favor having uniform `camel_case` fields for the:
1. More human-readable output in the `ron` case (in my opinion)
2. Consistency in the CLI output, which is effectively an API

To mitigate the pain of updating configurations, I could implement something like `iox2 config check --fix` or something.

That being said, I would be fine to leave it as-is if the issues above are not deemed significant. 

Another option I guess would to be to create wrapper structs for the CLI output to keep the config use-cases separate.

elBoberido: Unfortunately, we did not version the config files. If this would be the case, we would be able to read the version and the print a warning to use `iox2 config check --fix`. We cold of course start with the versioning now, and if no version info is present, print the warning.

We will most probably also break the C++ API with the new `Optional`, so we might as well declare the v0.8.0 as `broken records` release :grimacing: 

@elfenpiff what is your opinion?

orecham: @elfenpiff Ping

elfenpiff: I side with @orecham on this one but I have no strong feelings. But this would have to be prepared a bit more in the release notes. We have to state search `-` and replace it with `_`

orecham: What about the idea of implementing a cli command to check and fix the config to make it easier for users to update their configs.

Something like `iox2 config check` and `iox2 config fix`.

It might take some time to actually get to this though. Not so sure it will be trivial to do it well.

elBoberido: Also no strong feeling about this. We will introduce breakages with the new C++ `Optional` so it might be a good idea to use the next release for other breaking changes as well.

orecham: @elBoberido @elfenpiff What is the final verdict for this release? Keep `kebab-case` or use `camel_case` to be consistent with the other config params? If we change it, we won't have an `iox2 config check` command for this release though.

dkroenke (approved):

iceoryx2/src/config.rs
orecham: I reverted to `kebab-case`

PR #1071, opened 2025-09-26
C++ Static Vector Implementation

ComicSansMS:
## Notes for Reviewer
This contains the full implementation and initial tests for the C++ Base library StaticVector. Additional unit tests will be added in a follow-up PR.

* [x] Assign PR to reviewer

## References

Relates to #951


iceoryx2-bb/cxx/include/iox2/container/detail/raw_byte_storage.hpp
```diff
@@ -0,0 +1,180 @@
...
+        for (uint64_t index = 0; index < m_size; ++index) {
+            new (pointer_from_index(index)) T(*rhs.pointer_from_index(index));
+        }
+    }
+
+    constexpr RawByteStorage(RawByteStorage&& rhs) noexcept
+        : m_bytes {}
+        , m_size(rhs.m_size) {
+        for (uint64_t index = 0; index < m_size; ++index) {
+            new (pointer_from_index(index)) T(std::move_if_noexcept(*rhs.pointer_from_index(index)));
```
elfenpiff, line 56: Didn't know until now that this function existed. Neat little helper

iceoryx2-bb/cxx/include/iox2/container/detail/raw_byte_storage.hpp
```diff
@@ -0,0 +1,180 @@
...
+        }
+    }
+
+    template <uint64_t M, std::enable_if_t<(N > M), bool> = true>
+    // NOLINTNEXTLINE(hicpp-explicit-conversions), conceptually a move constructor
+    constexpr RawByteStorage(RawByteStorage<T, M>&& rhs)
+        : m_bytes {}
+        , m_size(rhs.size()) {
+        for (uint64_t index = 0; index < m_size; ++index) {
+            new (pointer_from_index(index)) T(std::move(*rhs.pointer_from_index(index)));
```
elfenpiff: Shouldn't we use here the same `std::move_if_noexcept` to have consistent behavior with the normal move ctor. Also shouldn't this function be declared `noexcept`?

iceoryx2-bb/cxx/include/iox2/container/detail/raw_byte_storage.hpp
```diff
@@ -0,0 +1,180 @@
...
+    constexpr
+#endif
+        ~RawByteStorage() {
+        for (uint64_t i = m_size; i != 0; --i) {
+            uint64_t const index = i - 1;
+            pointer_from_index(index)->~T();
+        }
+    }
+
+    constexpr auto operator=(RawByteStorage const&) -> RawByteStorage& = delete;
```
elfenpiff: Why did you allow only move and copy construction but not assignment?

iceoryx2-bb/cxx/include/iox2/container/detail/raw_byte_storage.hpp
```diff
@@ -0,0 +1,180 @@
...
+
+    // @pre (size() < (N - 1)) && (index <= size())
+    template <typename... Args>
+    constexpr void emplace_at(uint64_t index, Args&&... args) {
+        make_room_at(index, 1);
+        T& target = *pointer_from_index(index);
+        target = T(std::forward<Args>(args)...);
+    }
+
+    // @pre (index <= size()) && (size() + count < N)
```
elfenpiff: When `index == size` wouldn't that mean that the element is not yet constructed at the index and therefore we need to use placement new instead of assignment?

We could solve this by either weaken the precondition to `index < size` or we need handle this case in the implementation.

iceoryx2-bb/cxx/include/iox2/container/detail/raw_byte_storage.hpp
```diff
@@ -0,0 +1,180 @@
...
+#include <new>
+#include <type_traits>
+
+namespace iox2 {
+namespace container {
+namespace detail {
+
+/// A class for storing at most N objects of type T in a contiguous storage.
+/// All operations on this class are unchecked.
+template <typename T, uint64_t N>
```
elfenpiff: Could you rename `N` into `Capacity`. I started wondering what `N` was and had to scroll up to see its definition but with `Capacity` it would have been totally clear.

iceoryx2-bb/cxx/include/iox2/container/detail/raw_byte_storage.hpp
```diff
@@ -0,0 +1,180 @@
...
+class RawByteStorage {
+    // NOLINTNEXTLINE(modernize-type-traits), _v requires C++17
+    static_assert(std::is_standard_layout<T>::value, "Storage is only valid for standard layout types.");
+
+  private:
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays,hicpp-avoid-c-arrays,modernize-avoid-c-arrays) raw storage, will not be used as array
+    alignas(T) char m_bytes[sizeof(T) * N];
```
elfenpiff: Could we use `uint8_t` here instead of `char`, or maybe even declare a type alias like `using Byte = uint8_t;` or `using Byte = char` just to make it clear it is a byte.

It always feels a bit iky because in Rust it is 4 bytes - even though this code is correct.

The reason behind this is that maybe a less knowledgeable C++ developer will encounter this code later, and then weird stuff happens when it is mixed with Rust.

iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
```diff
@@ -0,0 +1,396 @@
...
+        }
+
+      public:
+        ~UncheckedConstAccessor() = default;
+        UncheckedConstAccessor(UncheckedConstAccessor const&) = delete;
+        UncheckedConstAccessor(UncheckedConstAccessor&&) = delete;
+        auto operator=(UncheckedConstAccessor const&) -> UncheckedConstAccessor& = delete;
+        auto operator=(UncheckedConstAccessor&&) -> UncheckedConstAccessor& = delete;
+
+        constexpr auto operator[](SizeType index) const -> ConstReference {
```
elfenpiff: I am wondering if we should convert all methods in the `Unchecked{Const}Accessor` to rvalue methods like:
```cxx
 constexpr auto operator[](SizeType index) const -> ConstReference && {
```
to avoid the user storing the accessor in a variable and using it then all over the place.

When the destruction order or lifetime is violated to the corresponding vector we can introduce quickly undefined behavior.

iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
```diff
@@ -0,0 +1,396 @@
...
+#include <functional>
+#include <initializer_list>
+#include <memory>
+#include <type_traits>
+
+namespace iox2 {
+namespace container {
+
+/// A resizable container with compile-time fixed static capacity and contiguous inplace storage.
+template <typename T, uint64_t N>
```
elfenpiff: Could you please add a doxygen documentation for all public methods.

iceoryx2-bb/cxx/include/iox2/container/detail/raw_byte_storage.hpp
ComicSansMS: Yes and yes.

iceoryx2-bb/cxx/include/iox2/container/detail/raw_byte_storage.hpp
ComicSansMS: `uint8_t` would not be a good choice here, as it is an arithmetic type that loses all of the special properties of `char` as a representation of binary storage. But we can roll our own version of C++17's  `std::byte`, which is basically an `enum class byte : unsigned char {};`.

iceoryx2-bb/cxx/include/iox2/container/detail/raw_byte_storage.hpp
ComicSansMS: You are right, there is a bug in how new elements get inserted here.

iceoryx2-bb/cxx/include/iox2/container/detail/raw_byte_storage.hpp
elBoberido: Just skimmed through the comments and this one took my attention. In iceoryx_hoofs we once also tried the `enum class byte` approach, but AFAIR we had some issues with some compilers, so we kept the alias to a `char`.

Another issue we had with some tools was that they complained about unaligned access. It was ensured that there wasn't an unaligned access, but wince we only had `char*`, the tools could not prove it. The solution for this was to define an `Element` like
```cpp
template <typename ElementType>
struct alignas(ElementType) Element {
    char data[sizeof(ElementType)];
};

Element<Foo> buffer[13];
```

None of the tools complained but I'm not quite sure anymore if we overdid it with the type punning ... leaving this PR now :)

iceoryx2-bb/cxx/include/iox2/container/detail/raw_byte_storage.hpp
ComicSansMS: You are right, this should be `=default`.

iceoryx2-bb/cxx/include/iox2/container/detail/raw_byte_storage.hpp
ComicSansMS: The entire approach of `make_room_at` was misguided. You cannot make room and then insert, instead you have to insert at the back and then rotate into place. I adjusted the implementation.

iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
ComicSansMS: I was thinking about this and I don't have a strong opinion.

I guess it can be useful to discourage keeping the accessors around longer than necessary, but is not exactly what the rvalue refs where designed for. Maybe we just add it and wait if we get complaints?

iceoryx2-bb/cxx/include/iox2/container/detail/raw_byte_storage.hpp
ComicSansMS: I don't see a clear best way forward here for now. I would suggest to leave this with `char` for now and file a follow-up issue?

iceoryx2-bb/cxx/include/iox2/container/static_vector.hpp
ComicSansMS: Added ref qualifiers on all accessor functions. Let's see how it goes.

iceoryx2-bb/cxx/include/iox2/container/detail/raw_byte_storage.hpp
ComicSansMS: Tracked as [#1079](https://github.com/eclipse-iceoryx/iceoryx2/issues/1079).

elfenpiff (approved):

PR #1110, opened 2025-10-14
[#1109] unchecked system type creator functions

### LINKED ISSUE #1109: Add unchecked `from_path_and_file` for FilePath
## Brief feature description

When it is used in combination with compile time constants this would be helpful.

elfenpiff:

* [x] Assign PR to reviewer

## References

Closes #1109

### DIFF
--- a/iceoryx2-bb/container/src/string/static_string.rs
+++ b/iceoryx2-bb/container/src/string/static_string.rs
@@ -299,6 +299,8 @@ impl<const CAPACITY: usize> StaticString<CAPACITY> {
     ///  * all unicode code points must be smaller 128 and not 0.
     ///
     pub const unsafe fn from_bytes_unchecked(bytes: &[u8]) -> Self {
+        debug_assert!(bytes.len() <= CAPACITY);
+
         Self::from_bytes_unchecked_restricted(bytes, bytes.len())
     }
 
--- a/iceoryx2-bb/system-types/src/file_name.rs
+++ b/iceoryx2-bb/system-types/src/file_name.rs
@@ -174,12 +174,15 @@ impl<const CAPACITY: usize> iceoryx2_bb_container::semantic_string::SemanticStri
     }
 
     unsafe fn new_unchecked(bytes: &[u8]) -> Self {
+        debug_assert!(bytes.len() <= CAPACITY);
+
         Self {
             value: iceoryx2_bb_container::string::StaticString::from_bytes_unchecked(bytes),
         }
     }
 
     unsafe fn insert_bytes_unchecked(&mut self, idx: usize, bytes: &[u8]) {
+        debug_assert!(bytes.len() + self.len() <= CAPACITY);
         use iceoryx2_bb_container::string::String;
         self.value.insert_bytes_unchecked(idx, bytes);
     }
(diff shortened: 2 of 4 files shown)


iceoryx2-bb/system-types/src/file_path.rs
```diff
@@ -103,27 +103,56 @@ semantic_string! {
...
+        let mut buffer = [0u8; PATH_LENGTH];
+        let mut buffer_len = path.as_bytes_const().len();
+
+        core::ptr::copy_nonoverlapping(
+            path.as_bytes_const().as_ptr(),
+            buffer.as_mut_ptr(),
+            buffer_len,
+        );
+
+        if 0 < buffer_len && path.as_bytes_const()[buffer_len - 1] != PATH_SEPARATOR {
```
elBoberido, line 140: I don't know how other see this, but the notation with the constant of the left and the variable on the right ist quite confusing and I haven't seen it elsewhere, except in some old automotive code when it comes to the equal operator.

For example, in this case the requirement is not that 0 is less than `buffer_length` but that `buffer_length` is larger than 0 because that means that 1 can be subtracted from `buffer_length`. When it written from left to right is, to think around the corner one has ;)

iceoryx2-bb/container/src/string/static_string.rs
```diff
@@ -279,13 +279,29 @@ impl<const CAPACITY: usize> StaticString<CAPACITY> {
     ///  * `bytes` len must be smaller or equal than [`StaticString::capacity()`]
     ///  * all unicode code points must be smaller 128 and not 0.
     ///
-    pub unsafe fn from_bytes_unchecked(bytes: &[u8]) -> Self {
+    pub const unsafe fn from_bytes_unchecked_restricted(bytes: &[u8], len: usize) -> Self {
         debug_assert!(bytes.len() <= CAPACITY);
+        debug_assert!(len <= bytes.len());
```
elBoberido, line 284: Is this sufficient for release builds? I know that the function is unsafe and called unchecked. But from the usage further below, the invariant might not have been checked before this function is called.

iceoryx2-bb/system-types/src/file_name.rs
```diff
@@ -186,6 +186,19 @@ impl<const CAPACITY: usize> iceoryx2_bb_container::semantic_string::SemanticStri
...
+    /// # Safety
+    ///
+    /// * The provided slice must have a length smaller or equal to the capacity. `value.len() < Self::max_len()`
+    /// * The contents of the slice must follow the content contract
+    ///
+    pub const unsafe fn new_unchecked_const(value: &[u8]) -> RestrictedFileName<CAPACITY> {
+        Self {
+            value: iceoryx2_bb_container::string::StaticString::from_bytes_unchecked(value),
+        }
+    }
```
elBoberido, line 204: For example, would it make sense to add a check here? Assuming a const check is possible at all.

iceoryx2-bb/system-types/src/file_path.rs
elfenpiff: I read an artical that stated that less than operator is easier to grasp then the greater than operator - and I agreed with it.

Ever since I try to write `a < b` instead of `b > a`

elBoberido: Well, I disagree with it since I'm used to read from left to right and when the semantic meaning is that x needs to be larger than y, then having to read that y is smaller than x requires extra mental capacity to understand what is actually means.

PR #1113, opened 2025-10-15
[#1099] Add component test for containers

### LINKED ISSUE #1099: Cross-Language Verification Tests
## Goal

Ensure that the container `StaticVec` & `StaticString` can be used in an inter-process communication context (C++ <-> Rust) without serialization.

Therefore, we require tests that ensure
1. an identical memory layout of that classes
  * the order/offsets of the members must be identical
  * the size and alignment of the members
  * the semantic type of the members (see `int64_t` vs `u64` in the `size` member)
2. the have the same content contracts
  * see `StaticString` that supports only characters < 128, if one programming language is able to add illegal characters it would break the contract on the other side

### 1. Identifcal memory layout

This requires whitebox testing on C++ and Rust side. We could introduce a `testing.hpp` file in the C++ library and a new `iceoryx2-bb-container-conformance-tests` crate on the C++ side.

* Introduce C function that can and return the size, alignment, offset and type of the members and return it
  * for the type we actually need only to know if the `m_size` variable is an unsigned integer, the rest can be deduced from the size
* Use bindgen and build.rs to generate a rust file to access the `testing.hpp` functions within a rust test suite
* write integration tests in `iceoryx2-bb-container-conformance-tests` that uses the `testing.hpp` functions and compares the layout with the rust counterpart

### 2. Same content contract

Introduce a new root folder called `component-tests` and structure it identical to the `examples` folder.
* `component-tests/cxx`
* `component-tests/rust`
* `component-tests/e2e-tests`

Write a request response process, where the client requests a change in a given container and the server responds with the performed change. See: 
* rust: https://github.com/eclipse-iceoryx/iceoryx2/tree/main/examples/rust/request_response
* c++: https://github.com/eclipse-iceoryx/iceoryx2/tree/main/examples/cxx/request_response
For instance: client add character `x` to given string, server responds with string and clients prints the result to the console.

The output can be verified with `expect` like it is done in the cross-language-end-to-end-tests in the examples folder. The character to add request can be parameterized via cli arguments.

ComicSansMS:
## Notes for Reviewer
The component tests consist of a C++ and a Rust executable.
* When starting up, both executable wait for the other side to appear
* Once connected, the Rust side publishes the name of a test it wants to execute
* The C++ side then opens a Req/Resp server for that test
* The Rust side proceeds to send requests for individual test cases
* Both sides validate their assumptions for the individual test case and can fail if they detect a fault
* If either side fails, the other side will terminate with a failure as well

The test cases are naturally quite tightly coupled, so it is unlikely that you can make changes to one side without breaking the other. If people have ideas how to make this more robust, I'm all ears.

* [x] Assign PR to reviewer

## References

Closes #1099

### DIFF
--- a/component-tests/cxx/src/test_containers.cpp
+++ b/component-tests/cxx/src/test_containers.cpp
@@ -60,7 +60,7 @@ auto as_string_literal(const VectorTypeSequence value) noexcept -> const char* {
     }
 }
 
-std::ostream& operator<<(std::ostream& stream, VectorTypeSequence value) noexcept {
+auto operator<<(std::ostream& stream, VectorTypeSequence value) noexcept -> std::ostream& {
     stream << as_string_literal(value);
     return stream;
 }


ComicSansMS:
Switching the PR to ready for review to trigger CI. Bazel build and CI scripts are still missing at this point and should be added before merging.

component-tests/rust/src/test_container_mutation.rs
```diff
@@ -0,0 +1,157 @@
...
+// This program and the accompanying materials are made available under the
+// terms of the Apache Software License 2.0 which is available at
+// https://www.apache.org/licenses/LICENSE-2.0, or the MIT license
+// which is available at https://opensource.org/licenses/MIT.
+//
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use super::common::*;
+use iceoryx2::prelude::*;
+use std::fmt::Debug;
```
orecham:
```suggestion
use core::fmt::Debug;
```

component-tests/rust/src/test_containers.rs
```diff
@@ -0,0 +1,156 @@
...
+// terms of the Apache Software License 2.0 which is available at
+// https://www.apache.org/licenses/LICENSE-2.0, or the MIT license
+// which is available at https://opensource.org/licenses/MIT.
+//
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use super::common::*;
+use iceoryx2::prelude::*;
+use iceoryx2_bb_container::vector::StaticVec;
+use std::fmt::Debug;
```
orecham:
```suggestion
use core::fmt::Debug;
```

component-tests/rust/src/test_container_mutation.rs
ComicSansMS: Thanks. Out of curiosity, would the `std::` version actually break under a `no_std` build? It was my understanding that there is actually only one definition of the type which then gets aliased under both `core` and `std` crate names. Is that correct?

orecham: @ComicSansMS The compilation does fail even though it is the same type under the hood.

I have seen the compilation failures myself, but it is also mentioned here: https://rust-lang.github.io/rust-clippy/master/index.html#std_instead_of_core

> Crates which have no_std compatibility may wish to ensure types are imported from core to ensure disabling std does not cause the crate to fail to compile.

EDIT: Looking up the reasoning, it seems that since the `std` module does not exist in `no_std` builds, the compiler cannot find it to verify it is the same type.

Cargo.toml
```diff
@@ -47,7 +47,9 @@ members = [
     "benchmarks/request-response",
     "benchmarks/publish-subscribe",
     "benchmarks/event",
-    "benchmarks/queue"
+    "benchmarks/queue",
+
+    "component-tests/rust"
```
elBoberido: nitpicking: To prevent having this line shown as changed with the next crate to be added.
```suggestion
    "component-tests/rust",
```

Only if there are other findings. Else, this will most probably be merged before my conformance tests suite PR, which needs to be rebased due to the bazel changes anyway, and I can fix it there.

internal/scripts/ci_build_and_run_end_to_end_tests.sh
```diff
@@ -83,7 +84,7 @@ if [[ ${BUILD_END_TO_END_TESTS} == true ]]; then
         -DCMAKE_BUILD_TYPE=Debug \
         -DBUILD_CXX=ON \
         -DBUILD_EXAMPLES=ON \
-        -DBUILD_TESTING=OFF
+        -DBUILD_TESTING=ON
```
elBoberido, line 87: Okay, now I remember again why I added the separate cmake flag. I wanted to prevent to have to build the C++ unit tests just to run the end-to-end-tests. There are other CI jobs which take much longer to run and the C++ unit tests should build quite fast. Semantically, this is the better approach, though. Let's keep it.

component-tests/cxx/src/test_containers.cpp
```diff
@@ -0,0 +1,197 @@
...
+    bool all_fields_match;
+};
+
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-magic-numbers,readability-magic-numbers)
+struct alignas(64) ContainerTestOverAligned {
+    int32_t i;
+};
+
+template <typename TestType, uint64_t TestCapacity>
+auto check_metrics_for_vector(ContainerTestRequest const& req) -> bool {
```
elBoberido, line 94: Shouldn't there be a `check_metrics_for_string`?

component-tests/rust/src/main.rs
```diff
@@ -0,0 +1,63 @@
...
+// information regarding copyright ownership.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Apache Software License 2.0 which is available at
+// https://www.apache.org/licenses/LICENSE-2.0, or the MIT license
+// which is available at https://opensource.org/licenses/MIT.
+//
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+mod common;
```
elBoberido, line 17:
```suggestion
// SPDX-License-Identifier: Apache-2.0 OR MIT

#![warn(clippy::alloc_instead_of_core)]
#![warn(clippy::std_instead_of_alloc)]
#![warn(clippy::std_instead_of_core)]

mod common;
```

We also need to run these tests on the target platform, which might not support `std`. While these tests do not need to be `no_std` right away, we should try to use the `core` and `alloc` crates if the type is re-exported from there.

You just need to replace `std::fmt` with `core::fmt`.

component-tests/cxx/src/test_containers.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+// NOLINTBEGIN(readability-identifier-naming)
+// NOLINTNEXTLINE(performance-enum-size)
+enum class VectorTypeSequence : int32_t {
+    VecI32_10 = 1,
+    VecI64_20 = 2,
+    VecOverAligned_5 = 3,
+    VecVec8_10 = 4,
+    EndOfTest = -1,
+};
```
elBoberido, line 43: In iceoryx classic we had the convention to use a `asStringLiteral` to function to easily convert this to a string for std::cout and our logger implementation. With our naming convention, this would be
```cpp
auto as_string_literal(const VectorTypeSequence value) noexcept -> const char* {
    switch (value)
    {
    case VectorTypeSequence::VecI32_10:
        return "VectorTypeSequence::VecI32_10";
    case VectorTypeSequence::VecI64_20:
        return "VectorTypeSequence::VecI64_20";
    case VectorTypeSequence::VecOverAligned_5:
        return "VectorTypeSequence::VecOverAligned_5";
    case VectorTypeSequence::VecVec8_10:
        return "VectorTypeSequence::VecVec8_10";
    case VectorTypeSequence::EndOfTest:
        return "VectorTypeSequence::EndOfTest";
    default:
        return "[Undefined VectorTypeSequence]";  
    }
}

std::ostream& operator<<(std::ostream& stream, VectorTypeSequence value) noexcept
{
    stream << as_string_literal(value);
    return stream;
}
```

This would make the output of these values more human friendly.

component-tests/cxx/src/test_containers.cpp
```diff
@@ -0,0 +1,197 @@
...
+// NOLINTEND(readability-identifier-naming)
+
+struct ContainerTestResponse {
+    // IOX2_TYPE_NAME is equivalent to the payload type name used on the Rust side
+    static constexpr const char* IOX2_TYPE_NAME = "ContainerTestResponse";
+    int32_t vector_type_sequence;
```
elBoberido: This could then become
```suggestion
    VectorTypeSequence vector_type_sequence;
```

component-tests/cxx/src/test_containers.cpp
```diff
@@ -0,0 +1,197 @@
...
+    auto test_name() const -> char const* override {
+        return "containers";
+    }
+    auto run_test(iox2::Node<iox2::ServiceType::Ipc> const& node) -> bool override;
+};
+
+struct ContainerTestRequest {
+    // IOX2_TYPE_NAME is equivalent to the payload type name used on the Rust side
+    static constexpr const char* IOX2_TYPE_NAME = "ContainerTestRequest";
+    int32_t vector_type_sequence;
```
elBoberido: This could then become
```suggestion
    VectorTypeSequence vector_type_sequence;
```

Which would then also make the casts to `int` further down obsolete.

elBoberido (commented):
Mostly minor things, except maybe a missing `check_metrics_for_string`. Not sure what was discussed with @elfenpiff but should be easy to add.

elBoberido (approved):
Looks good. Let's merge this and have the `check_metrics_for_string` in a follow up PR. There are currently multiple PRs open which touch `Cargo.Bazel.lock` and when this one is merged, we can prepare the next one with  `Cargo.Bazel.lock` changes

component-tests/cxx/src/test_containers.cpp
ComicSansMS: Covered by [#1143](https://github.com/eclipse-iceoryx/iceoryx2/pull/1143)

PR #1118, opened 2025-10-16
[#865] Use `core` and `alloc` in Rust code

orecham:
## Notes for Reviewer

Updates all crates to explicitly use components in `core` and `alloc`.

All crates were updated except for the following (to be done in a follow-up):
1. `iceoryx2-cli`

Additionally, the following changes were added to fix some quirks that were uncovered:
1. Logging
    1. Remove indentation in log output when `origin` is not provided
    2. Print log message on first line when `origin` is not provided
    3. Remove `|` character from log output when `origin` is not provided
    4. Add a `null` logger default which does nothing

* [x] Assign PR to reviewer

## References

Relates #865

### DIFF
--- a/examples/rust/blackboard/creator.rs
+++ b/examples/rust/blackboard/creator.rs
@@ -14,10 +14,10 @@ use core::time::Duration;
 
 extern crate alloc;
 use alloc::boxed::Box;
-use alloc::format;
 
 use examples_common::BlackboardKey;
 use iceoryx2::prelude::*;
+use iceoryx2_bb_log::println;
 
 const CYCLE_TIME: Duration = Duration::from_secs(1);
 
--- a/examples/rust/blackboard/creator.rs
+++ b/examples/rust/blackboard/creator.rs
@@ -11,8 +11,14 @@
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
 use core::time::Duration;
+
+extern crate alloc;
+use alloc::boxed::Box;
+use alloc::format;
+
 use examples_common::BlackboardKey;
 use iceoryx2::prelude::*;
+use iceoryx2_bb_log::info;
 
 const CYCLE_TIME: Duration = Duration::from_secs(1);
 
@@ -30,7 +36,7 @@ fn main() -> Result<(), Box<dyn core::error::Error>> {
         .add::<f64>(key_1, INITIAL_VALUE_1)
         .create()?;
 
-    println!("Blackboard created.\n");
+    info!("Blackboard created.\n");
 
     let writer = service.writer_builder().create()?;
 
@@ -43,16 +49,16 @@ fn main() -> Result<(), Box<dyn core::error::Error>> {
         counter += 1;
 
         entry_handle_mut_0.update_with_copy(counter);
-        println!("Write new value for key 0: {counter}");
+        info!("Write new value for key 0: {counter}");
 
         let entry_value_uninit = entry_handle_mut_1.loan_uninit();
         let value = INITIAL_VALUE_1 * counter as f64;
         let entry_value = entry_value_uninit.write(value);
         entry_handle_mut_1 = entry_value.update();
-        println!("Write new value for key 1: {}\n", value);
+        info!("Write new value for key 1: {}\n", value);
     }
 
-    println!("exit");
+    info!("exit");
 
     Ok(())
 }
--- a/examples/rust/blackboard/creator.rs
+++ b/examples/rust/blackboard/creator.rs
@@ -18,7 +18,6 @@ use alloc::format;
 
 use examples_common::BlackboardKey;
 use iceoryx2::prelude::*;
-use iceoryx2_bb_log::info;
 
 const CYCLE_TIME: Duration = Duration::from_secs(1);
 
@@ -36,7 +35,7 @@ fn main() -> Result<(), Box<dyn core::error::Error>> {
         .add::<f64>(key_1, INITIAL_VALUE_1)
         .create()?;
 
-    info!("Blackboard created.\n");
+    println!("Blackboard created.\n");
 
     let writer = service.writer_builder().create()?;
 
@@ -49,16 +48,16 @@ fn main() -> Result<(), Box<dyn core::error::Error>> {
         counter += 1;
 
         entry_handle_mut_0.update_with_copy(counter);
-        info!("Write new value for key 0: {counter}");
+        println!("Write new value for key 0: {counter}");
 
         let entry_value_uninit = entry_handle_mut_1.loan_uninit();
         let value = INITIAL_VALUE_1 * counter as f64;
         let entry_value = entry_value_uninit.write(value);
         entry_handle_mut_1 = entry_value.update();
-        info!("Write new value for key 1: {}\n", value);
+        println!("Write new value for key 1: {}\n", value);
     }
 
-    info!("exit");
+    println!("exit");
 
     Ok(())
 }
--- a/examples/rust/blackboard/creator.rs
+++ b/examples/rust/blackboard/creator.rs
@@ -17,7 +17,7 @@ use alloc::boxed::Box;
 
 use examples_common::BlackboardKey;
 use iceoryx2::prelude::*;
-use iceoryx2_bb_log::println;
+use iceoryx2_bb_log::cout;
 
 const CYCLE_TIME: Duration = Duration::from_secs(1);
 
@@ -35,7 +35,7 @@ fn main() -> Result<(), Box<dyn core::error::Error>> {
         .add::<f64>(key_1, INITIAL_VALUE_1)
         .create()?;
 
-    println!("Blackboard created.\n");
+    cout!("Blackboard created.\n");
 
     let writer = service.writer_builder().create()?;
 
@@ -48,16 +48,16 @@ fn main() -> Result<(), Box<dyn core::error::Error>> {
         counter += 1;
 
         entry_handle_mut_0.update_with_copy(counter);
-        println!("Write new value for key 0: {counter}");
+        cout!("Write new value for key 0: {counter}");
 
         let entry_value_uninit = entry_handle_mut_1.loan_uninit();
         let value = INITIAL_VALUE_1 * counter as f64;
         let entry_value = entry_value_uninit.write(value);
         entry_handle_mut_1 = entry_value.update();
-        println!("Write new value for key 1: {}\n", value);
+        cout!("Write new value for key 1: {}\n", value);
     }
 
-    println!("exit");
+    cout!("exit");
 
     Ok(())
 }
--- a/examples/rust/discovery_service/discovery_service.rs
+++ b/examples/rust/discovery_service/discovery_service.rs
@@ -10,7 +10,11 @@
 //
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
+extern crate alloc;
+use alloc::boxed::Box;
+
 use iceoryx2::prelude::*;
+use iceoryx2_bb_log::{error, info};
 use iceoryx2_services_discovery::*;
 use service_discovery::Discovery;
 
@@ -24,7 +28,7 @@ fn main() -> Result<(), Box<dyn core::error::Error>> {
         .publish_subscribe::<service_discovery::Payload>()
         .open()
         .inspect_err(|_| {
-            eprintln!("Unable to open service discovery service. Was it started?");
+            error!("Unable to open service discovery service. Was it started?");
         })?;
 
     let subscriber = publish_subscribe.subscriber_builder().create()?;
@@ -34,7 +38,7 @@ fn main() -> Result<(), Box<dyn core::error::Error>> {
         .event()
         .open()
         .inspect_err(|_| {
-            eprintln!("unable to open service discovery service. Was it started?");
+            error!("unable to open service discovery service. Was it started?");
         })?;
     let listener = event.listener_builder().create()?;
 
@@ -42,7 +46,7 @@ fn main() -> Result<(), Box<dyn core::error::Error>> {
     let guard = waitset.attach_notification(&listener)?;
     let attachment = WaitSetAttachmentId::from_guard(&guard);
 
-    println!("Discovery service ready!");
+    info!("Discovery service ready!");
 
     let on_event = |attachment_id: WaitSetAttachmentId<ipc::Service>| {
         if attachment_id == attachment {
@@ -53,10 +57,10 @@ fn main() -> Result<(), Box<dyn core::error::Error>> {
             while let Ok(Some(sample)) = subscriber.receive() {
                 match sample.payload() {
                     Discovery::Added(details) => {
-                        println!("Added: {:?}", details.name());
+                        info!("Added: {:?}", details.name());
                     }
                     Discovery::Removed(details) => {
-                        println!("Removed: {:?}", details.name());
+                        info!("Removed: {:?}", details.name());
                     }
                 }
             }
--- a/examples/rust/discovery_service/discovery_service.rs
+++ b/examples/rust/discovery_service/discovery_service.rs
@@ -14,7 +14,8 @@ extern crate alloc;
 use alloc::boxed::Box;
 
 use iceoryx2::prelude::*;
-use iceoryx2_bb_log::error;
+use iceoryx2_bb_log::eprintln;
+use iceoryx2_bb_log::println;
 use iceoryx2_services_discovery::*;
 use service_discovery::Discovery;
 
--- a/examples/rust/discovery_service/discovery_service.rs
+++ b/examples/rust/discovery_service/discovery_service.rs
@@ -14,7 +14,7 @@ extern crate alloc;
 use alloc::boxed::Box;
 
 use iceoryx2::prelude::*;
-use iceoryx2_bb_log::{error, info};
+use iceoryx2_bb_log::error;
 use iceoryx2_services_discovery::*;
 use service_discovery::Discovery;
 
@@ -46,7 +46,7 @@ fn main() -> Result<(), Box<dyn core::error::Error>> {
     let guard = waitset.attach_notification(&listener)?;
     let attachment = WaitSetAttachmentId::from_guard(&guard);
 
-    info!("Discovery service ready!");
+    println!("Discovery service ready!");
 
     let on_event = |attachment_id: WaitSetAttachmentId<ipc::Service>| {
         if attachment_id == attachment {
@@ -57,10 +57,10 @@ fn main() -> Result<(), Box<dyn core::error::Error>> {
             while let Ok(Some(sample)) = subscriber.receive() {
                 match sample.payload() {
                     Discovery::Added(details) => {
-                        info!("Added: {:?}", details.name());
+                        println!("Added: {:?}", details.name());
                     }
                     Discovery::Removed(details) => {
-                        info!("Removed: {:?}", details.name());
+                        println!("Removed: {:?}", details.name());
                     }
                 }
             }
--- a/examples/rust/discovery_service/discovery_service.rs
+++ b/examples/rust/discovery_service/discovery_service.rs
@@ -14,8 +14,8 @@ extern crate alloc;
 use alloc::boxed::Box;
 
 use iceoryx2::prelude::*;
-use iceoryx2_bb_log::eprintln;
-use iceoryx2_bb_log::println;
+use iceoryx2_bb_log::cerr;
+use iceoryx2_bb_log::cout;
 use iceoryx2_services_discovery::*;
 use service_discovery::Discovery;
 
@@ -29,7 +29,7 @@ fn main() -> Result<(), Box<dyn core::error::Error>> {
         .publish_subscribe::<service_discovery::Payload>()
         .open()
         .inspect_err(|_| {
-            eprintln!("Unable to open service discovery service. Was it started?");
+            cerr!("Unable to open service discovery service. Was it started?");
         })?;
 
     let subscriber = publish_subscribe.subscriber_builder().create()?;
@@ -39,7 +39,7 @@ fn main() -> Result<(), Box<dyn core::error::Error>> {
         .event()
         .open()
         .inspect_err(|_| {
-            eprintln!("unable to open service discovery service. Was it started?");
+            cerr!("unable to open service discovery service. Was it started?");
         })?;
     let listener = event.listener_builder().create()?;
 
@@ -47,7 +47,7 @@ fn main() -> Result<(), Box<dyn core::error::Error>> {
     let guard = waitset.attach_notification(&listener)?;
     let attachment = WaitSetAttachmentId::from_guard(&guard);
 
-    println!("Discovery service ready!");
+    cout!("Discovery service ready!");
 
     let on_event = |attachment_id: WaitSetAttachmentId<ipc::Service>| {
         if attachment_id == attachment {
@@ -58,10 +58,10 @@ fn main() -> Result<(), Box<dyn core::error::Error>> {
             while let Ok(Some(sample)) = subscriber.receive() {
                 match sample.payload() {
                     Discovery::Added(details) => {
-                        println!("Added: {:?}", details.name());
+                        cout!("Added: {:?}", details.name());
                     }
                     Discovery::Removed(details) => {
-                        println!("Removed: {:?}", details.name());
+                        cout!("Removed: {:?}", details.name());
                     }
                 }
             }
--- a/examples/rust/discovery_service/discovery_service.rs
+++ b/examples/rust/discovery_service/discovery_service.rs
@@ -28,7 +28,7 @@ fn main() -> Result<(), Box<dyn core::error::Error>> {
         .publish_subscribe::<service_discovery::Payload>()
         .open()
         .inspect_err(|_| {
-            error!("Unable to open service discovery service. Was it started?");
+            eprintln!("Unable to open service discovery service. Was it started?");
         })?;
 
     let subscriber = publish_subscribe.subscriber_builder().create()?;
@@ -38,7 +38,7 @@ fn main() -> Result<(), Box<dyn core::error::Error>> {
         .event()
         .open()
         .inspect_err(|_| {
-            error!("unable to open service discovery service. Was it started?");
+            eprintln!("unable to open service discovery service. Was it started?");
         })?;
     let listener = event.listener_builder().create()?;
 
--- a/examples/rust/domains/discovery.rs
+++ b/examples/rust/domains/discovery.rs
@@ -16,7 +16,7 @@ use alloc::string::String;
 
 use clap::Parser;
 use iceoryx2::prelude::*;
-use iceoryx2_bb_log::println;
+use iceoryx2_bb_log::cout;
 
 fn main() -> Result<(), Box<dyn core::error::Error>> {
     set_log_level_from_env_or(LogLevel::Info);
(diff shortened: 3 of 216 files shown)


iceoryx2-pal/concurrency-sync/src/once.rs
```diff
@@ -0,0 +1,120 @@
...
+/// static INIT: Once = Once::new();
+///
+/// fn initialize() {
+///     INIT.call_once(|| {
+///         // This will only run once, even if called from multiple threads
+///         println!("Initializing...");
+///     });
+/// }
+/// ```
+pub struct Once {
```
orecham, line 54: Where should this go? I presume we don't want to have crates depend directly on `iceoryx2-pal-concurrency-sync`

iceoryx2-bb/log/src/logger/console.rs
```diff
@@ -170,7 +171,7 @@ impl crate::Log for Logger {
                     Self::print_origin(log_level, &origin_str);
                 }
                 true => std::eprint!(
-                    "{}{:9} {} ",
+                    "{}{} {} ",
```
orecham, line 174: Removed this indentation so that logs with and without `origin` are aligned.

iceoryx2-bb/log/src/logger/console.rs
```diff
@@ -118,12 +118,13 @@ impl Logger {
...
 
     fn print_message(log_level: crate::LogLevel, formatted_message: &str) {
-        Self::print("| ", Self::message_color(log_level), formatted_message);
+        Self::print("", Self::message_color(log_level), formatted_message);
     }
 
     fn print_origin(log_level: crate::LogLevel, origin: &str) {
         eprint!("{} ", Logger::log_level_string(log_level));
         Self::print("", Logger::origin_color(log_level), origin);
+        eprint!("| ");
```
orecham, line 127: Changes this so the `|` character is only printed when there is an `origin` provided.

iceoryx2-bb/log/Cargo.toml
```diff
@@ -11,10 +11,19 @@ rust-version = { workspace = true }
 version = { workspace = true }
 
 [features]
+default = []
+std = []
+
+# Enables logging to buffer
```
orecham, line 17: Added a feature for each provided logger type. If none are compiled in, the `null` logger is used.

iceoryx2-bb/container/src/lib.rs
```diff
@@ -102,6 +102,16 @@ pub mod semantic_string;
 /// A container to store key-value pairs.
 pub mod flatmap;
 
+// TODO: Consider replacing with own implementations to reduce certification complexity
```
orecham: Added these re-exports for HashMap and HashSet.

We might want to eventually have our own implementations for these.

iceoryx2/src/node/mod.rs
```diff
@@ -733,21 +743,29 @@ fn remove_node<Service: service::Service>(
 
 #[derive(Debug)]
 pub(crate) struct RegisteredServices {
-    data: Mutex<HashMap<ServiceId, (ContainerHandle, u64)>>,
+    handle: MutexHandle<HashMap<ServiceId, (ContainerHandle, u64)>>,
```
orecham: `iceoryx2-bb-posix::Mutex` is now used here

iceoryx2-pal/concurrency-sync/src/once.rs
elBoberido: Good question. Maybe a re-export in one of the bb crates or a new bb crate? @elfenpiff what do you think?

iceoryx2-bb/log/src/logger/console.rs
elBoberido: Hmm, what do you think about moving the origin at the second position, after the log message? Maybe not in this PR but just in general.

iceoryx2-bb/container/src/lib.rs
elBoberido: Maybe the BTreeMap can be used instead of the HashMap. The BTreeMap is in alloc.

orecham: @elBoberido Ah yes, you did mention this. I have removed these re-exports and instead substituted `HashMap` and `HashSet` with `BTreeMap` and `BTreeSet`.

elBoberido: @orecham and it worked out of the box? Awesome. It should even be faster than the HashMap and HashSet :)

iceoryx2-bb/log/Cargo.toml
elfenpiff: Can you please add all those feature flags, with documentation to the `iceoryx2/src/libs.rs` file. This is the main start page that all users see on docs.rs. There is also a separate section for all feature flags at the end.

iceoryx2-ffi/c/Cargo.toml
```diff
@@ -18,6 +18,12 @@ build = "build.rs"
 path = "src/lib.rs"
 crate-type = ["rlib", "cdylib", "staticlib"] # without "rlib" the doc examples are not run
 
+[features]
+default = ["std"]
```
elfenpiff: Could you please also add the `std` feature flag to `iceoryx2/src/lib.rs` into the `Feature Flag` section of the documentation.

iceoryx2-pal/concurrency-sync/src/once.rs
elfenpiff: For now we could keep it here. The `IoxAtomics` are also here and used all over the place. But to be more compliant with the architecture layers of iceoryx2 we would need to move `Once` and the `IoxAtomics` into some crate of iceoryx2-bb, maybe iceoryx2-bb-elementary - or at least re-export it there.

iceoryx2-bb/container/src/lib.rs
orecham: @elBoberido Yep!

iceoryx2-bb/log/src/logger/console.rs
orecham: @elBoberido @elfenpiff I will leave it to you guys to decide. Could you create an issue if it's decided that we want to make the change.

examples/rust/discovery_service/discovery_service.rs
```diff
@@ -24,7 +28,7 @@ fn main() -> Result<(), Box<dyn core::error::Error>> {
         .publish_subscribe::<service_discovery::Payload>()
         .open()
         .inspect_err(|_| {
-            eprintln!("Unable to open service discovery service. Was it started?");
+            error!("Unable to open service discovery service. Was it started?");
```
elBoberido: This is not a good idea. The user might have redirected the log to a file or completely switched off and will never see this. I guess it was initially meant to always be printed.

I'm fine with having this as temporary solution but please use a different alias so that we can easily find all the locations later on and change them according to #1127 

Don't get mad if I annotate all this cases with `nostd_println` :)

examples/rust/discovery_service/discovery_service.rs
```diff
@@ -34,7 +38,7 @@ fn main() -> Result<(), Box<dyn core::error::Error>> {
         .event()
         .open()
         .inspect_err(|_| {
-            eprintln!("unable to open service discovery service. Was it started?");
+            error!("unable to open service discovery service. Was it started?");
```
elBoberido: nostd_println

examples/rust/domains/discovery.rs
```diff
@@ -10,12 +10,14 @@
...
+extern crate alloc;
+use alloc::boxed::Box;
+use alloc::string::String;
+
 use clap::Parser;
 use iceoryx2::prelude::*;
-use iceoryx2_bb_log::{set_log_level, LogLevel};
 
 fn main() -> Result<(), Box<dyn core::error::Error>> {
-    set_log_level_from_env_or(LogLevel::Info);
```
elBoberido, line 18: Please revert this

examples/rust/domains/discovery.rs
```diff
@@ -56,7 +58,7 @@ fn define_log_level(args: &Args) {
     if args.debug {
         set_log_level(LogLevel::Trace);
     } else {
-        set_log_level(LogLevel::Warn);
+        set_log_level(LogLevel::Info);
     }
```
elBoberido: Please remove this. It predates the `set_log_level_from_env_or` approach.

Please do the same in the `publisher.rs` and `subscriber.rs` files.

iceoryx2-bb/log/src/lib.rs
```diff
@@ -150,22 +165,34 @@ pub mod log;
 pub mod fail;
 pub mod logger;
 
-use iceoryx2_pal_concurrency_sync::iox_atomic::IoxAtomicU8;
+#[cfg(feature = "std")]
+pub use from_env::{set_log_level_from_env_or, set_log_level_from_env_or_default};
```
elBoberido, line 169: Fine for now. But can you create a task or a sub-issue to use `getenv` from `stdlib` in our platform abstraction. See also `iceoryx_platform`.

iceoryx2-bb/log/src/lib.rs
```diff
@@ -142,6 +142,21 @@
...
 //! ```
 
+#[cfg(all(feature = "logger_buffer", not(feature = "std")))]
+compile_error!("The 'logger_buffer' feature requires the 'std' feature to be enabled");
+
+#[cfg(all(feature = "logger_file", not(feature = "std")))]
+compile_error!("The 'logger_file' feature requires the 'std' feature to be enabled");
+
+#[cfg(all(feature = "logger_console", not(feature = "std")))]
+compile_error!("The 'logger_console' feature requires the 'std' feature to be enabled");
```
elBoberido, line 152: Fine for now, but can you create a task or sub-issue to use `write` from `unistd` in our platform abstraction. See also `iceoryx_platform`.

iceoryx2-bb/posix/src/config.rs
```diff
@@ -109,9 +112,9 @@ pub fn does_system_satisfy_posix_requirements(mode: ComplianceCheckMode) -> bool
     let mut is_compliant: bool = true;
 
     if mode == ComplianceCheckMode::Verbose {
-        println!("{HEADER_COLOR}system requirements check{COLOR_RESET}");
-        println!();
-        println!(" {HEADER_COLOR}minimum system{COLOR_RESET}");
+        info!("{HEADER_COLOR}system requirements check{COLOR_RESET}");
+        info!("");
+        info!(" {HEADER_COLOR}minimum system{COLOR_RESET}");
```
elBoberido: nostd_println and all the other changes in this file

iceoryx2/src/node/mod.rs
```diff
@@ -733,21 +743,29 @@ fn remove_node<Service: service::Service>(
 
 #[derive(Debug)]
 pub(crate) struct RegisteredServices {
-    data: Mutex<HashMap<ServiceId, (ContainerHandle, u64)>>,
+    handle: MutexHandle<BTreeMap<ServiceId, (ContainerHandle, u64)>>,
```
elBoberido, line 746: I need to review this change with a fresh mind.

elBoberido (commented):
Reviewed everything but the `Once` implementation and the changes with the `MutexHandle`.

Not sure about the spin-based implementation of `Once`. I need to check where it is used and how it affects the code. AFAIK, the `Once` from `std` is `futex` based. Should we potentially have a mutex based implementation with a spin-based fallback in the platform abstraction?

examples/rust/discovery_service/discovery_service.rs
orecham: @elBoberido This a leftover I forgot to revert in https://github.com/eclipse-iceoryx/iceoryx2/pull/1118/commits/ce92eda5e1f75f7e73b3f960ee5ce4beb8e6b294, thanks for pointing it out.

examples/rust/domains/discovery.rs
orecham: Also a leftover that should have been reverted in: https://github.com/eclipse-iceoryx/iceoryx2/pull/1118/commits/ce92eda5e1f75f7e73b3f960ee5ce4beb8e6b294

My bad.

iceoryx2-bb/posix/src/config.rs
orecham: This would need to use https://github.com/eclipse-iceoryx/iceoryx2/issues/1127.

We can revert it or leave it as is until https://github.com/eclipse-iceoryx/iceoryx2/issues/1127 is implemented. If we  revert, we cannot build for no_std in the meantime.

This code is seldom called so I am not sure it will cause a problem to leave as is in the interim.

iceoryx2-bb/log/src/lib.rs
orecham: Use `write` for the `compile_error` macro?

Also, do we expect to be able to implement`unistd` for all platforms?

orecham:
@elBoberido @elfenpiff 

In this PR, the following changes were made to put the code in a state where it could compile for `no_std`:

1. Use logger instead of std print functionality
2. Use spin-based `Once` implementation

As per previous discussions and the review comments, these solutions could be improved with:

1. An abstraction for printing functionality
2. Implementations for `Once` that use platform-specific waiting mechanisms (would need a solution (or different ones) for all platforms)

How shall we proceed?

1. Remove the use of logger and the spin-based once and merge proper solutions in separate PRs
    1. `no_std` compilation would not work but there would be no sub-optimal intermediate solutions merged into
1. Merge the use of logger and spin-based `Once`
    1. `no_std` compilation would work but sub-optimal solutions would be merged in

elBoberido:
@orecham I'm fine with option 2. I would adjust it a little bit, though.

* config.rs
  * can you create an alias for `info!` at the top of the file, e.g. `foo_println!` and use the alias in the file
  * also, add a todo with the `#1127`

* once.rs
  * can you move the implementation in a module and build it only for `no_std`
  * re-export the spin-based `Once` in the parent scope for `no_std`
  * re-export `std::sync::Once` for non `no_std` builds

With this, our current builds with `std` still behave the same and we can brainstorm whether we want to use the spin-based `Once` for all builds or if we want to have a dedicated implementation for each platform with platform specific OS calls

Btw. I still need to review the remaining files.

iceoryx2/src/node/mod.rs
elBoberido: Okay. It seems that `MutexHandle` is probably not the best name, since it contains the storage for `pthrad_mutex_t`, clock type and the value it protects.

Something like `MutexCell`, `MutexState` or `MutexResources` would probably be better. What do you think? Does not need to be changed in this PR but maybe you could create an issue if you have a better name and @elfenpiff agrees to a name change

Also, this should not be called just `handle` but probably `data`, like it was before. A variable that contains data wrapped by a std Mutex is also not called `mutex` but according to the data it wraps.

iceoryx2/src/node/mod.rs
```diff
@@ -733,21 +743,29 @@ fn remove_node<Service: service::Service>(
...
 
 impl RegisteredServices {
+    pub(crate) fn new() -> Self {
+        let handle = MutexHandle::new();
+
+        MutexBuilder::new()
+            .is_interprocess_capable(false)
+            .mutex_type(MutexType::Normal)
+            .create(BTreeMap::new(), &handle)
+            .expect("Failed to create mutex");
```
elBoberido: I guess elfenpiff would rather see a `fatal_panic!` here

iceoryx2/src/node/mod.rs
```diff
@@ -777,18 +796,27 @@ impl RegisteredServices {
...
+
+        if let Some(entry) = guard.get_mut(service_id) {
             entry.1 -= 1;
             if entry.1 == 0 {
-                cleanup_call(entry.0);
-                data.remove(service_id);
+                let handle = entry.0;
+                guard.remove(service_id);
+                drop(guard);
+                cleanup_call(handle);
```
elBoberido, line 817: We need to be extremely careful here, not to introduce cleanup races. @elfenpiff can you please check if the changed order is problematic?

iceoryx2-pal/concurrency-sync/src/once.rs
```diff
@@ -0,0 +1,120 @@
...
+    ///
+    /// This method can be used to check whether the closure passed to
+    /// `call_once` has already been executed without re-triggering initialization.
+    pub fn is_completed(&self) -> bool {
+        self.load_state() == State::Complete
+    }
+
+    #[inline]
+    fn load_state(&self) -> State {
+        self.state.load(Ordering::Acquire).into()
```
elBoberido, line 106: I think this must be `Ordering::SeqCst` since there is no corresponding `Ordering::Release`. `Ordering::Relaxed` will not work, since it actually needs to sync memory.

cc @elfenpiff @ComicSansMS @FerdinandSpitzschnueffler

iceoryx2-pal/concurrency-sync/tests/once_tests.rs
```diff
@@ -0,0 +1,68 @@
...
+        for _ in 0..NUMBER_OF_THREADS {
+            s.spawn(|| {
+                barrier.wait(|_, _| {}, |_| {});
+                once.call_once(|| {
+                    counter.fetch_add(1, Ordering::Relaxed);
+                });
+            });
+        }
+
+        barrier.wait(|_, _| {}, |_| {});
```
elBoberido, line 57: This can be removed. AFAIK, the `main` thread blocks until the scoped threads are joined.

iceoryx2-pal/concurrency-sync/src/once.rs
ComicSansMS: The only relevant store for the synchronization here seems to be the one on line 88 and that one has Release ordering? Or am I missing another race?

examples/rust/domains/discovery.rs
orecham: Oh you mean remove this arg completely? Sure.

iceoryx2-bb/log/src/lib.rs
orecham: @elBoberido Could you point me to this function? I might be blind but I cannot find it with `rg`

iceoryx2/src/node/mod.rs
orecham: @elBoberido I named it that way due to the method `Mutex::from_handle`. From how I understood it, it's not really "data wrapped in a mutex" but a handle to a mutex somewhere that wraps the data, which then you access using the `from_handle` function?

iceoryx2/src/node/mod.rs
orecham: @elfenpiff

iceoryx2/src/node/mod.rs
orecham: @elBoberido I changed it to preserve order of operations.

iceoryx2-pal/concurrency-sync/tests/once_tests.rs
orecham: @elBoberido This barrier is to wait until all threads finish spawning and are ready to race to initialize.

orecham:
> @orecham I'm fine with option 2. I would adjust it a little bit, though.
> 
> * config.rs
>   * can you create an alias for `info!` at the top of the file, e.g. `foo_println!` and use the alias in the file
>   * also, add a todo with the `#1127`
> 
> * once.rs
>   * can you move the implementation in a module and build it only for `no_std`
>   * re-export the spin-based `Once` in the parent scope for `no_std`
>   * re-export `std::sync::Once` for non `no_std` builds
> 
> With this, our current builds with `std` still behave the same and we can brainstorm whether we want to use the spin-based `Once` for all builds or if we want to have a dedicated implementation for each platform with platform specific OS calls

@elBoberido 

I added a wrapper for `println` in `iceoryx2-bb-log` (will be moved later) which does nothing when `std` is not enabled. I also added a switch to use `std::sync::Once` when `std` is available in crates that use it.

iceoryx2-bb/posix/src/config.rs
orecham: @elBoberido This has been changed to use a wrapper around `println` which does nothing in no_std builds. In #1127 a proper abstraction will be implemented.

examples/rust/domains/discovery.rs
elBoberido: Yes, please remove completely :)

iceoryx2-bb/log/src/lib.rs
elBoberido: It's here -> https://github.com/eclipse-iceoryx/iceoryx/blob/main/iceoryx_platform/generic/source/stdlib.cpp#L36

iceoryx2-bb/log/src/lib.rs
elBoberido: No, I meant to use `write` from `unistd` for the `logger_console`, then this `cfg` could be removed.

Not sure it `unistd` would be implemented for each platform but something with the signature of `write`. On platforms with no meaningful output, this could just discard the string.

iceoryx2-bb/posix/src/config.rs
elBoberido: For a porting effort it would make sense to have the output and if it is temporally the logger on `no_std`, this is totally fine. But it can be changed when necessary. So no action required.

iceoryx2/src/node/mod.rs
elBoberido: Just to prevent that we talk from different things
```rs
pub struct MutexHandle<T: Sized + Debug> {
    pub(crate) handle: HandleStorage<posix::pthread_mutex_t>,
    clock_type: UnsafeCell<ClockType>,
    value: UnsafeCell<Option<T>>,
}
```

From my understanding, this holds the storage for the actual mutex, the clock type and the storage for the actual data.

Not suggesting to change it in this PR, at least not `MutexHandle`. But it might make sence to revert the change from `data` to `handle`

iceoryx2-pal/concurrency-sync/tests/once_tests.rs
elBoberido: Isn't that the barrier on line 50?

examples/rust/blackboard/creator.rs
```diff
@@ -11,8 +11,14 @@
...
 
 use core::time::Duration;
+
+extern crate alloc;
+use alloc::boxed::Box;
+use alloc::format;
+
 use iceoryx2::prelude::*;
 use iceoryx2_bb_container::string::*;
+use iceoryx2_bb_log::println;
```
elBoberido: Food for thought. I'm also not sure what would be the best approach.

On one hand, calling this just `println`, leads to not having to change other code, on the other hand, one could easily add an example without this import an use `println` from std. The CI would then complain and one would wonder why it fails since the other examples also use `println`.

Unfortunately, I don't know what's the Rust idiomatic way.

elBoberido:
> I also added a switch to use std::sync::Once when std is available in crates that use it.

That's fine for now.

The reason why I suggested to have the re-export in `once.rs` was to have a single location for the `std`/`no_std` switch.

I also noticed that it is only used for the logger. In this case, I think it might make sense to move the implementation there and keep it private. I'm not sure why it is also used in the C FFI. That one should actually just delegate the calls to the Rust implementation. I need to investigate this.

examples/rust/blackboard/creator.rs
elfenpiff: Lets call it `iox2_println` to clearly distinguish it from normal `println`.

iceoryx2-bb/log/src/lib.rs
elfenpiff: I think `unistd` should be available on every platform where it would make sense to run iceoryx2

iceoryx2-bb/log/src/lib.rs
elfenpiff: Here we could put in a bit more effort and also implement a slight wrapper around `getenv` so that the user does not have to deal with raw c pointers. It could be a minimalistic `no_std` version (based on our platform) of the rust env: https://doc.rust-lang.org/std/env/index.html

iceoryx2-bb/log/src/lib.rs
```diff
@@ -271,9 +266,80 @@ pub fn get_logger() -> &'static dyn Log {
...
 pub fn __internal_print_log_msg(log_level: LogLevel, origin: Arguments, args: Arguments) {
     if get_log_level() <= log_level as u8 {
         get_logger().log(log_level, origin, args)
     }
 }
+
+// TODO(#1127): Add proper printing abstraction that can be implemented for no_std platforms
+#[cfg(feature = "std")]
+#[macro_export]
+macro_rules! println {
```
elfenpiff: Please rename it to `cout`

iceoryx2-bb/log/src/lib.rs
```diff
@@ -271,9 +266,80 @@ pub fn get_logger() -> &'static dyn Log {
...
+    ($($arg:tt)*) => {
+        std::println!($($arg)*)
+    };
+}
+#[cfg(feature = "std")]
+#[macro_export]
+macro_rules! eprintln {
```
elfenpiff: Please rename it to `cerr`

iceoryx2-ffi/c/src/api/log.rs
```diff
@@ -15,13 +15,17 @@
...
+
 use iceoryx2_bb_log::{
-    get_log_level, set_log_level, set_log_level_from_env_or, set_log_level_from_env_or_default,
-    set_logger, Log, LogLevel, __internal_print_log_msg,
-    logger::{use_console_logger, use_file_logger},
+    get_log_level, set_log_level, set_logger, Log, LogLevel, __internal_print_log_msg,
 };
 
-use core::ffi::{c_char, CStr};
+#[cfg(not(feature = "std"))]
```
elfenpiff: We should avoid having `cfg` directives in the code at much as possible. In this sense, could we move this selection to `iceoryx2_pal_concurrency_sync`. The `Once` implementation is either our own orn the `std` one. But the user just strictly uses `iceoryx2_pal_concurrency_sync::once::Once` - always!

To ensure this, please add `std::sync::Once` to the disallowed types list in the root `clippy.toml`.

iceoryx2/src/node/mod.rs
elfenpiff: You guessed correctly ;)

iceoryx2/src/node/mod.rs
```diff
@@ -777,17 +796,27 @@ impl RegisteredServices {
         service_id: &ServiceId,
         mut cleanup_call: F,
     ) {
-        let mut data = self.data.lock().unwrap();
-        if let Some(entry) = data.get_mut(service_id) {
+        let mut guard = self.mutex().lock().expect("Failed to lock mutex");
```
elfenpiff, line 811: Please use `fatal_panic!`

iceoryx2-bb/log/src/lib.rs
elBoberido: While I like the idea on one hand, it might look out of place in a Rust codebase, which uses `*print*` all over the place. How about `iprintln` :grin:

iceoryx2/src/node/mod.rs
elBoberido: @orecham what's your opinion on this?

iceoryx2-bb/log/src/lib.rs
elBoberido: @orecham can you create an issue for this?

iceoryx2-pal/concurrency-sync/src/once.rs
elBoberido: It has been quite some time since I reviewed lock-free code 😅 

@orecham could you create an issue for a more thorough review of this code? Maybe when all of us meet in person?

elBoberido (approved):

iceoryx2-bb/log/src/lib.rs
orecham: https://github.com/eclipse-iceoryx/iceoryx2/issues/1152

iceoryx2-bb/log/src/lib.rs
orecham: https://github.com/eclipse-iceoryx/iceoryx2/issues/1153

iceoryx2-pal/concurrency-sync/src/once.rs
orecham: https://github.com/eclipse-iceoryx/iceoryx2/issues/1154

PR #1168, opened 2025-11-02
[#1167] Support C++ 14

### LINKED ISSUE #1167: Support C++14 for the C++ Bindings
## Brief feature description

There are still a lot of users which require C++14 support, especially in automotive. The C++ bindings of iceoryx2 should build with the `CMAKE_CXX_STANDARD=14` cmake flag.

elBoberido:
## Notes for Reviewer

This PR adds C++14 support for the C++ bindings. For this, the already ported iceoryx_hoofs subset needs to be used.

* [x] Assign PR to reviewer

## References

Closes #1167

### DIFF
--- a/iceoryx2-cxx/tests/src/service_blackboard_tests.cpp
+++ b/iceoryx2-cxx/tests/src/service_blackboard_tests.cpp
@@ -566,14 +566,10 @@ TYPED_TEST(ServiceBlackboardTest, add_with_default_stores_default_value) {
 
     const auto service_name = iox2_testing::generate_service_name();
 
-    constexpr uint16_t VALUE = 27;
+    static constexpr uint16_t VALUE = 27;
     struct TestDefault {
-        TestDefault()
-            : t { VALUE } {
-        }
-
         // NOLINTNEXTLINE(misc-non-private-member-variables-in-classes), come on, its a test
-        uint16_t t;
+        uint16_t t { VALUE };
     };
 
     auto node = NodeBuilder().create<SERVICE_TYPE>().expect("");
--- a/iceoryx2-cxx/tests/src/service_publish_subscribe_tests.cpp
+++ b/iceoryx2-cxx/tests/src/service_publish_subscribe_tests.cpp
@@ -735,7 +735,7 @@ TYPED_TEST(ServicePublishSubscribeTest, send_receive_with_user_header_works) {
     const uint64_t payload = 781891729871;
     sample.payload_mut() = payload;
     for (uint64_t idx = 0; idx < TestHeader::CAPACITY; ++idx) {
-        sample.user_header_mut().value.at(idx) = 4 * idx + 3;
+        sample.user_header_mut().value.at(idx) = (4 * idx) + 3;
     }
     send(std::move(sample)).expect("");
     auto recv_sample = sut_subscriber.receive().expect("");


iceoryx2-bb/cxx/include/iox2/container/optional.hpp
```diff
@@ -332,22 +332,14 @@ Optional(T) -> Optional<T>;
...
-inline constexpr NulloptT nullopt { detail::NulloptTConstructorTag {} };
-#else
 namespace detail {
 template <typename = void>
 struct NulloptHelper {
     static NulloptT nullopt;
 };
-template <>
-// NOLINTNEXTLINE(misc-definitions-in-headers), not an ODR violation because of template
-NulloptT NulloptHelper<>::nullopt = NulloptT { detail::NulloptTConstructorTag {} };
```
elBoberido: @ComicSansMS I had to move this to a `optional.cpp` since there were `multiple definition` errors in the tests. Feel free to use your magic to revert my workaround :)

iceoryx2-bb/cxx/include/iox2/container/static_string.hpp
```diff
@@ -72,7 +72,11 @@ class StaticString {
       public:
         ~UncheckedConstAccessor() = default;
         UncheckedConstAccessor(UncheckedConstAccessor const&) = delete;
+#if __cplusplus >= 201703L
         UncheckedConstAccessor(UncheckedConstAccessor&&) = delete;
+#else
+        UncheckedConstAccessor(UncheckedConstAccessor&&) = default;
+#endif
```
elBoberido: @ComicSansMS I had to implement the move c'tor since RVO is not mandatory in C++14. I think it should not cause any issues. If we want to prevent that the move c'tor is used, I could add an `assert(false)` in the body of the implementation.

I'm not sure about the compiler switch, though. Although the string is header only and this is a temporary object, we might end up in having the C++ bindings compiled with C++14 and user code using C++17 or newer. This might end up in some hard to track issues. Shall we introduce a `deployment.hpp.in` and "hardcode" the value that is used to build this lib?

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
```diff
@@ -332,23 +328,6 @@ Optional(T) -> Optional<T>;
...
-struct NulloptHelper {
-    static NulloptT nullopt;
-};
-template <>
-// NOLINTNEXTLINE(misc-definitions-in-headers), not an ODR violation because of template
-NulloptT NulloptHelper<>::nullopt = NulloptT { detail::NulloptTConstructorTag {} };
-} // namespace detail
-// NOLINTNEXTLINE(readability-identifier-naming), for consistency with C++17 code using std::optional
-static constexpr NulloptT const& nullopt = detail::NulloptHelper<>::nullopt;
-#endif
```
elBoberido, line 350: @ComicSansMS I had to remove this since there were multiple definition errors in the tests and clang-tidy also complained. As workaround, I took the same approach as we had in iceoryx_hoofs and it seems to work.

Feel free to use your magic to revert my workaround :)

elfenpiff (approved):

iceoryx2-bb/cxx/include/iox2/container/static_string.hpp
ComicSansMS: Being able to move this is not a problem per-se, we mainly deleted it as a strong hint to the users to discourage doing anything with this other than call it as part of a bigger enclosing expression. So I have no issue with having it `=default` for all builds, and I think that this is probably the preferred solution to avoid ODR headaches in mixed builds.

I would not make things much more complicated, as this is a fairly minor point of the overall design.

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
ComicSansMS: Yep, you are right, this was actually broken.

The hoofs implementation does have observably different behavior, albeit in a subtle way. It boils down to [this example](https://godbolt.org/z/G6djMqfcd). This should not be a problem for the common uses of `nullopt`. I am considering to reintroduce an `#ifdef` to get the correct behavior in 17, as it should not make the situation worse, even on mixed builds. But I will ponder on that some more...

iceoryx2-bb/cxx/include/iox2/container/static_string.hpp
elBoberido: Okay, I will then remove the compiler switch and add a comment to change the `default` to `delete` when we move to C++17 as min version.

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
elBoberido: @ComicSansMS sure. Shall I create an issue?

ComicSansMS: I will take care of it, thanks.

iceoryx2-bb/cxx/include/iox2/container/static_string.hpp
elBoberido: see #1173

PR #1197, opened 2025-11-26
[#1176] Custom iceoryx2-pal-posix Abstraction

### LINKED ISSUE #1176: Customize Platform Abstraction Layer
## Customize Platform Abstraction Layer

For an easier integration of iceoryx2 on different platforms, users need sometimes to customize settings in `iceoryx2-pal`. An example is that users may want to change the `ICEORYX2_ROOT_PATH` as defined in `iceoryx2-pal/configuration/src/lib.rs` on a different embedded platform. Currently, users need to change the values within the lib.rs directly and recompile. 

## Improvement Suggestion

We should give the user the ability to pass a custom platform and platform setting to iceoryx2
so that they can run iceoryx on their own platforms.
To achieve this we could go in two stages:
- [x] Give ability to `iceoryx2-pal-configuration` to pass a custom platform config file
- [x] Give ability to `iceoryx2-pal-posix` to pass a custom POSIX platform abstraction

dkroenke:
## Notes for Reviewer

See config/README.md for instructions on how to test with a custom platform

* [x] Assign PR to reviewer

## References

Closes #1176

### DIFF
--- a/config/README.md
+++ b/config/README.md
@@ -283,3 +283,78 @@ For Bazel the equivalent build command is:
 ```cli
 bazel build //... --action_env=IOX2_CUSTOM_PAL_CONFIG_PATH=/my/funky/platform --//:custom_pal_config="on"
 ```
+
+## Custom Platform Abstraction Layer POSIX
+
+Similar to the PAL configuration users may want to use an own
+POSIX abstraction (`iceoryx2/iceoryx2-pal/posix/src`) for a
+specific operating system.
+
+To achieve this we can use the environment variable
+`IOX2_CUSTOM_PAL_POSIX_PATH` together with the
+feature flag `custom_pal_posix`.
+
+The first step is to create a POSIX platform folder
+(can be outside of iceoryx2).
+For this example we use `/my/funky/platform/posix`
+The easiest is to copy an existing POSIX abstraction
+and adapt it to the specific needs.
+
+Inside of the `posix` folder is the
+`os.rs` located that is loading
+`posix` mod.
+To make the custom POSIX abstraction discoverable
+we need to adapt the path to the mod.rs:
+
+```rust
+#[cfg(target_os = "windows")]
+#[path = "/my/funky/platform/posix/mod.rs"]
+pub mod posix;
+```
+
+The `path` attribute needs to contain the absolute path to the mod.rs
+of the custom POSIX abstraction.
+Optionally a `#[cfg()]` attribute can be added for selecting
+the target operating system.
+
+Once this is done the build steps follow:
+
+1. Set environment variable with absolute path to POSIX abstraction
+(no trailing slash)
+
+```cli
+export IOX2_CUSTOM_PAL_POSIX_PATH=/my/funky/platform/posix
+```
+
+To make it persistent, the `./cargo/config.toml` file can be used,
+either the global one or the local one from the project:
+
+```toml
+[env]
+IOX2_CUSTOM_PAL_POSIX_PATH = "/my/funky/platform/posix"
+```
+
+1. Build iceoryx2 with feature `custom_pal_posix`
+for custom POSIX platform abstraction.
+
+```cli
+cargo build --features "custom_pal_posix"
+```
+
+For CMake the feature can be enabled with:
+
+```cli
+cmake . -Bbuild -DIOX2_CUSTOM_PAL_POSIX=ON -DBUILD_CXX=on
+```
+
+In Python:
+
+```cli
+poetry --project iceoryx2-ffi/python run maturin develop --manifest-path iceoryx2-ffi/python/Cargo.toml --target-dir target/ff/python --features custom_pal_posix
+```
+
+For Bazel the equivalent build command is:
+
+```cli
+bazel build //... --action_env=IOX2_CUSTOM_PAL_POSIX_PATH=/home/dietrich/devel/tmp/posix/src/linux --//:custom_pal_posix="on"
+```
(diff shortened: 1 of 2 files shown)


iceoryx2-pal/posix/src/lib.rs
```diff
@@ -19,28 +19,55 @@ extern crate alloc;
...
-mod platform;
+#[path = "qnx/iox2_pal_posix.rs"]
+mod iox2_pal_posix;
+
+#[cfg(not(feature = "custom_pal_posix"))]
 #[cfg(target_os = "windows")]
-#[path = "windows/mod.rs"]
-mod platform;
+#[path = "windows/iox2_pal_posix.rs"]
+mod iox2_pal_posix;
```
orecham: @dkroenke 

Why the rename?

BTW `iox2` is the name of the CLI, so I would avoid using it here. Something shorter like just `posix` makes more sense to me if there is a reason for this rename, although I think `platform` was fine, personally.

iceoryx2-pal/posix/src/lib.rs
```diff
@@ -77,5 +104,5 @@ pub mod posix {
     #[allow(unused_imports)]
     pub(crate) use common::string_operations::*;
 
-    pub use crate::platform::*;
+    pub use crate::iox2_pal_posix::platform::*;
```
orecham: Ah maybe the reason is that `pub use crate::platform::platform::*;` reads weirdly?  
Again I vote for the alternative `pub use crate::posix::platform::*;`

iceoryx2-pal/posix/Cargo.toml
```diff
@@ -18,6 +18,11 @@ version = { workspace = true }
 # feature flag.
 libc_platform = ["dep:libc"]
 
+# Uses a custom POSIX abstraction for the iceoryx2 platform abstraction layer
+# The location of posix abstraction can be set via environment variable
+# `IOX2_CUSTOM_PAL_POSIX_PATH`.
+custom_pal_posix = []
```
orecham, line 24: @dkroenke 

This could maybe be done without the feature flag by checking if the environment variable is empty, and if it is specifying the path to the appropriate platform-specific pal implementation. Something like:

```rust
#[cfg(target_os = "linux")]
#[path = concat!(env!("IOX2_CUSTOM_PAL_POSIX_PATH", "src/linux"), "/iox2_pal_posix.rs")]
mod iox2_pal_posix;
```

I haven't tested it completely though to check for quirks.

orecham: @dkroenke I think my suggestion as proposed won't work exactly, but something in this direction might be possible... what do you think?

dkroenke: > This could maybe be done without the feature flag by checking if the environment variable is empty, and if it is specifying the path to the appropriate platform-specific pal implementation. Something like:
> 
> ```rust
> #[cfg(target_os = "linux")]
> #[path = concat!(env!("IOX2_CUSTOM_PAL_POSIX_PATH", "src/linux"), "/iox2_pal_posix.rs")]
> mod iox2_pal_posix;
> ```
> 
> I haven't tested it completely though to check for quirks.

I'm not sure if this will work because of two issues:
- the `env!()` macro only accept environment variables that are set, if not a compiler error is emitted, alternative would be maybe `option_env!()` as you suggested: https://doc.rust-lang.org/std/macro.env.html
- the `path` attribute can only be in the form `#[path = "file"]`, paths generated from macros are currently not allowed: https://github.com/rust-lang/rust/issues/118312

That's why the `iox2_pal_posix.rs` (name is subject to change) is introduced as abstraction so that the user can set a custom path in his own posix abstraction.

iceoryx2-pal/posix/src/lib.rs
elfenpiff: +1

elfenpiff: Yes please, the more I review the more I really prefer just `crate::posix::platform::*`

But shouldnt it be actually the other way around? `crate::platform::posix` maybe there is later also a completely different platform called foo and then we use `crate::platform::foo` ... the other way around wouldn't make sense for me.

dkroenke: @elfenpiff @orecham How about `crate::os::posix::*` ? 
That could make more sense since we switch between the posix abstractions with `#[cfg(target_os ...`.
A platform is in my understanding more a combination of operating system and hardware.

orecham: @dkroenke Good for me :)

elfenpiff: I am fine with `crate::os::posix::*`

orecham (approved):
Adding another feature flag for this is not ideal, but I don't know any alternatives off the top of my head ... ideally we keep the number of features to a minimum. Perhaps something to investigate another time though.

config/README.md
```diff
@@ -283,3 +283,78 @@ For Bazel the equivalent build command is:
 ```cli
 bazel build //... --action_env=IOX2_CUSTOM_PAL_CONFIG_PATH=/my/funky/platform --//:custom_pal_config="on"
 ```
+
+## Custom Platform Abstraction Layer POSIX
```
orecham, line 287: @dkroenke One question, is it expected that these custom platforms work with the existing `posix.h` file in `iceoryx2-bb/posix/src/c` or should it be possible for users to provide their own header to only include the APIs that their custom platform uses?

PR #1204, opened 2025-11-29
[#1202] Add concurrency primitives crate to building blocks

### LINKED ISSUE #1202: Export synchronization primitives from `iceoryx2-bb`
## Brief Description

There are a number of synchronization primitives in `iceoryx2-pal` used throughout all layers of `iceoryx2`.

1. Atomics
2. Lazy
3. Spin

To adhere to the architecture, crates should not be depending directly on `iceoryx2-pal`, so these primitives should be exposed via a crate in `iceoryx2-bb`.

Once changed, all crates should access these primitives from the new location.

1. Usages of atomics throughout `iceoryx2`
2. Usages of `spin` in the bare metal example

orecham:
## Notes for Reviewer

1. Adds `iceoryx2-bb-concurrency` crate
1. Re-exports atomic types from `iceoryx2-pal`
1. Renames atomic types to drop the `Iox` prefix
    1. This change is more of a proposal - can be reverted if not agreed upon
1. Change upper-layer crates to depend on `iceoryx2-bb-concurrency` instead of `iceoryx2-pal-concurrency-sync`
1. Adds `SpinLock` to `iceoryx2-pal-concurrency-sync`, re-exports it in `iceoryx2-bb-concurrency` and uses it instead of the `spin` crate

A number of crates in upper layers of the architecture were depending directly on `iceoryx2-pal` for concurrency primitives, which is not congruent with the defined architecture.

A new crate is added to `iceoryx2-bb` to re-export required concurrency primitives which are implemented in `iceoryx2-pal`. Only the primitives required by upper layers are currently re-exported, more can be added as needed.

* [x] Assign PR to reviewer

## References

Closes #1202

### DIFF
--- a/iceoryx2-bb/concurrency/src/atomic.rs
+++ b/iceoryx2-bb/concurrency/src/atomic.rs
@@ -62,7 +62,7 @@ pub struct AtomicI64(internal::AtomicI64);
 #[repr(transparent)]
 pub struct AtomicIsize(internal::AtomicIsize);
 
-macro_rules! zero_copy_send_atomic {
+macro_rules! Impl {
     ($type_name:ident, $base_type:ident) => {
         impl $type_name {
             #[inline]
@@ -94,17 +94,17 @@ macro_rules! zero_copy_send_atomic {
     };
 }
 
-zero_copy_send_atomic!(AtomicBool, bool);
-zero_copy_send_atomic!(AtomicU8, u8);
-zero_copy_send_atomic!(AtomicU16, u16);
-zero_copy_send_atomic!(AtomicU32, u32);
-zero_copy_send_atomic!(AtomicU64, u64);
-zero_copy_send_atomic!(AtomicUsize, usize);
-zero_copy_send_atomic!(AtomicI8, i8);
-zero_copy_send_atomic!(AtomicI16, i16);
-zero_copy_send_atomic!(AtomicI32, i32);
-zero_copy_send_atomic!(AtomicI64, i64);
-zero_copy_send_atomic!(AtomicIsize, isize);
+Impl!(AtomicBool, bool);
+Impl!(AtomicU8, u8);
+Impl!(AtomicU16, u16);
+Impl!(AtomicU32, u32);
+Impl!(AtomicU64, u64);
+Impl!(AtomicUsize, usize);
+Impl!(AtomicI8, i8);
+Impl!(AtomicI16, i16);
+Impl!(AtomicI32, i32);
+Impl!(AtomicI64, i64);
+Impl!(AtomicIsize, isize);
 
 mod internal {
     pub use iceoryx2_pal_concurrency_sync::atomic::AtomicBool;


iceoryx2-bb/concurrency/src/atomic.rs
```diff
@@ -0,0 +1,68 @@
...
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use core::ops::{Deref, DerefMut};
+
+use iceoryx2_bb_elementary_traits::placement_default::PlacementDefault;
+use iceoryx2_bb_elementary_traits::zero_copy_send::ZeroCopySend;
+
+macro_rules! zero_copy_send_atomic {
+    ($type_name:ident, $base_type:ident) => {
+        paste::paste! {
```
orecham: This approach adds a new (compile-time) dependency `paste` to simplify the code for wrapping atomics in `iceoryx2-pal-concurrency-sync` in a new type so that `iceoryx2-bb` traits may be defined on it. If this new dependency is an issue, could potentially manually import all types as aliases for use in the macro instead...

iceoryx2-bb/concurrency/src/atomic.rs
```diff
@@ -0,0 +1,68 @@
...
+// which is available at https://opensource.org/licenses/MIT.
+//
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use core::ops::{Deref, DerefMut};
+
+use iceoryx2_bb_elementary_traits::placement_default::PlacementDefault;
+use iceoryx2_bb_elementary_traits::zero_copy_send::ZeroCopySend;
+
+macro_rules! zero_copy_send_atomic {
```
elfenpiff: I think this approach is a bit cumbersome and introduces a dependency which we actually do not need. What do you think about the following approach:

1. `iceoryx2-bb-elementary` depends directly on `iceoryx2-pal-concurrency-sync` as the only crate. In there we implement `PlacementDefault` and `ZeroCopySend` as it was before.
2. `iceoryx2-bb-concurrency` re-exports the `iceoryx2-pal-concurrency-sync` `Atomic**`'s with the new namespace `iceoryx2-bb-concurrency`

With this we get rid of the `paste` dependency and do it like the Rust lib does it with `core` and the `std` re-exports. Also I do not see the benefit of removing the `iceoryx2-pal-concurrency-sync` dependency in `iceoryx2-bb-elementary` .

elfenpiff (dismissed):
When possible, please use rather this approach: https://github.com/eclipse-iceoryx/iceoryx2/pull/1204/files#r2573007733

iceoryx2-bb/concurrency/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+std = []
+
+[dependencies]
+iceoryx2-bb-elementary-traits = { workspace = true }
+iceoryx2-pal-concurrency-sync = { workspace = true }
+
+paste = { workspace = true }
+
+[dev-dependencies]
+iceoryx2-bb-posix = { workspace = true }
```
elfenpiff: This is a cyclic dependency and cannot be published to crates.io - even when it is a dev-dependency!

Since iceoryx2-bb-posix requires atomics, you must remove this dependency or iceoryx2-bb-posix dependends again on the iceoryx2-pal-concurrency-sync crate.

elfenpiff: As far as I can see only the `spin_tests.rs` depends on posix, see `use iceoryx2_bb_posix::system_configuration::SystemInfo;` It is used to acquire the number of cpu cores to parallelize the tests as much as the hardware allows.

But I think for something simple like a spinlock tests it is maybe a bit overkill and we can hardcode this to always `2` so that the tests use exactly two threads.

With this the dev-dependency to `iceoryx2-bb-posix` would be removed.

iceoryx2-bb/concurrency/src/atomic.rs
orecham: @elfenpiff 

Do you mean `iceoryx2-bb-elementary-traits` instead of `iceoryx2-bb-elementary` throughout this comment?

> I think this approach is a bit cumbersome

If you mean the use of the "new type" wrapper, I can see how this could be undesirable  if we need the atomics in `iceoryx2-pal` and `iceoryx2-bb` to be recognized as the same type in some instances, but do we have this use-case if we change the upper-layer crates use the `iceoryx2-bb` version everywhere? 

Do we have the same requirement as `std` and `core`, where some use crates would like to seamlessly switch between depending on `iceoryx2-bb` and `iceoryx2-pal`?

> get rid of the paste dependency

If `paste` is a problem (even though it is compile-time only), it can be removed by just writing some more boilerplate code by hand. 


> I do not see the benefit of removing the iceoryx2-pal-concurrency-sync dependency in iceoryx2-bb-elementary .

The motivation was to have a less tangled dependency tree. I was aiming to only have one crate in `iceoryx2-bb` depend on `iceoryx2-pal-concurrency-sync`.

iceoryx2-bb/concurrency/Cargo.toml
orecham: Fixed

iceoryx2-bb/concurrency/src/atomic.rs
elfenpiff: @orecham 

> If you mean the use of the "new type" wrapper, I can see how this could be undesirable if we need the atomics in iceoryx2-pal and iceoryx2-bb to be recognized as the same type in some instances, but do we have this use-case if we change the upper-layer crates use the iceoryx2-bb version everywhere?

Actually, I do not mind the wrapper and that the types are recognized as different types. I mind more the implementation overhead and the additional dependency where I do not see the benefit.

> The motivation was to have a less tangled dependency tree. I was aiming to only have one crate in iceoryx2-bb depend on iceoryx2-pal-concurrency-sync.

Understand and agreed. But with my drafted approach it would be the case - or do I overlook something?

iceoryx2-bb/concurrency/src/atomic.rs
orecham: Opted to manually import all aliases to avoid the extra dependency.

iceoryx2-bb/concurrency/src/atomic.rs
orecham: @elfenpiff I removed the `paste` dependency.

iceoryx2-bb/concurrency/src/atomic.rs
```diff
@@ -0,0 +1,118 @@
...
+
+#[derive(Debug, Default)]
+#[repr(transparent)]
+pub struct AtomicI64(InternalAtomicI64);
+
+#[derive(Debug, Default)]
+#[repr(transparent)]
+pub struct AtomicIsize(InternalAtomicIsize);
+
+macro_rules! zero_copy_send_atomic {
```
elBoberido: The name is a bit misleading

iceoryx2-bb/posix/Cargo.toml
```diff
@@ -17,7 +17,9 @@ workspace = true
 default = [
   "lazy_static/spin_no_std"
 ]
-std = []
+std = [
+  "iceoryx2-bb-concurrency/std"
```
elBoberido: Is this really required?

orecham: True, it could be set at the upper level. At the moment strictly not needed even since the types used in this crate do not swap in `std` equivalents (like `Once`).

iceoryx2-bb/concurrency/src/atomic.rs
orecham: Renamed to just `Impl` like done in other crates, it's internal anyway, doesn't need anything fancy I guess.

iceoryx2-bb/concurrency/tests/spin_lock_tests.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    let _watchdog = Watchdog::new();
+    let barrier = Barrier::new(NUMBER_OF_THREADS);
+    let lk = SpinLock::new(0);
+    thread::scope(|s| {
+        for _ in 0..NUMBER_OF_THREADS {
+            s.spawn(|| {
+                barrier.wait();
+                let mut guard = lk.blocking_lock();
+                *guard += 1;
```
elBoberido, line 60: What do you think about looping e.g. 1000 times to increase the chance of races? Also for the other tests.

orecham: @elfenpiff Thoughts? Will change in a follow-up though.

Issue #301, opened 2024-07-17
Move iceoryx2 cxx addendum headers into iceoryx


elfenpiff:
## Brief feature description

In the ffi cxx binding of iceoryx2 are additions that belong into iceoryx_hoofs. They shall be moved there.

See: https://github.com/eclipse-iceoryx/iceoryx/issues/2318

elBoberido:
@elfenpiff shall we rename this issue and indicate that we move the iceoryx_hoof classes to iceoryx2?

elfenpiff:
> @elfenpiff shall we rename this issue and indicate that we move the iceoryx_hoof classes to iceoryx2?

Yes please.

PR #1206, opened 2025-11-30
[#301] Move 'into' files to 'iox2/bb'

elBoberido:
## Notes for Reviewer

This PR moves `iox2/legacy/into.hpp` and its dependencies to `iox2/bb/into.hpp`. Furthermore, the namespace is adjusted to `iox2::bb` and the clang-tidy warnings are fixed.

* [x] Assign PR to reviewer

## PR Reviewer Reminder

* Public API is documented

## References

Relates to #301

### DIFF
--- a/iceoryx2-bb/cxx/include/iox2/bb/into.hpp
+++ b/iceoryx2-bb/cxx/include/iox2/bb/into.hpp
@@ -23,47 +23,47 @@ namespace bb {
 
 /// @brief Helper struct to indicate a lossy conversion, e.g. from an unbounded type into a bounded type
 template <typename D>
-struct lossy { };
+struct Lossy { };
 
 namespace detail {
 /// @brief Helper struct to get the actual destination type 'T' for 'into' with an additional indirection like
-/// 'into<lossy<T>>'
+/// 'into<Lossy<T>>'
 template <typename T>
-struct extract_into_type {
+struct ExtractIntoType {
     using TargetType = T;
 };
 
-/// @brief Helper struct to get the actual destination type 'T' for 'into<lossy<T>>'
+/// @brief Helper struct to get the actual destination type 'T' for 'into<Lossy<T>>'
 template <typename T>
-struct extract_into_type<lossy<T>> {
+struct ExtractIntoType<Lossy<T>> {
     using TargetType = T;
 };
 } // namespace detail
 
 // Using a struct as impl, as free functions do not support partially specialized templates
 template <typename SourceType, typename DestinationType>
-struct FromTrait {
+struct From {
     // AXIVION Next Construct AutosarC++19_03-A7.1.5 : 'auto' is only used for the generic implementation which will always result in a compile error
     static auto from(const SourceType& value IOX2_MAYBE_UNUSED) noexcept {
         static_assert(legacy::always_false_v<SourceType> && legacy::always_false_v<DestinationType>, "\n \
 Conversion for the specified types is not implemented!\n \
-Please specialize 'FromTrait::from'!\n \
+Please specialize 'From::from'!\n \
 -------------------------------------------------------------------------\n \
 template <typename SourceType, typename DestinationType>\n \
-constexpr DestinationType FromTrait::from(const SourceType&) noexcept;\n \
+constexpr DestinationType From::from(const SourceType&) noexcept;\n \
 -------------------------------------------------------------------------");
     }
 };
 
 /// @brief Converts a value of type SourceType to a corresponding value of type DestinationType. This function needs to
 /// be specialized by the user for the types to be converted. If a partial specialization is needed, please have a look
-/// at 'FromTrait'.
+/// at 'From'.
 /// @note If the conversion is potentially lossy 'Destination from<Source, Destination>(...)' should not be used but
 /// instead either one or both of:
-///   - 'Destination from<Source, lossy<Destination>>(...)'
+///   - 'Destination from<Source, Lossy<Destination>>(...)'
 ///   - 'optional<Destination> from<Source, optional<Destination>>(...)'
 /// The 'Destination from<Source, Destination>(...)' implementation should have a 'static_assert' with a hint of the
-/// reason, e.g. lossy conversion and a hint to use 'Destination into<lossy<Destination>>(...)' or
+/// reason, e.g. lossy conversion and a hint to use 'Destination into<Lossy<Destination>>(...)' or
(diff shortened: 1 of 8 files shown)


iceoryx2-bb/cxx/include/iox2/bb/into.hpp
```diff
@@ -27,19 +30,34 @@ namespace detail {
...
 template <typename T>
 struct extract_into_type<lossy<T>> {
-    using type_t = T;
+    using TargetType = T;
 };
 } // namespace detail
 
+// Using a struct as impl, as free functions do not support partially specialized templates
+template <typename SourceType, typename DestinationType>
+struct FromTrait {
```
elBoberido: Renamed `FromImpl` to `FromTrait`. `Impl` was not quite the right name. As alternative to `FromTrait`, just `From` would also work well.

How to handle such changes regarding the breaking changes in the changelog? Technically it is a new class in iceoryx2 and it was not part of the public API.

iceoryx2-bb/cxx/include/iox2/bb/into.hpp
elBoberido: According to our coding style, this should be upper case, but it somehow does not feel right for this type of structs. While one could make a good argument for `lossy`, we have some type traits which are backported from a newer C++ version and which are functional equivalent, e.g. `always_false_v`. In that case, lower snake case makes more sense from my point of view. Maybe something to discuss in the next developer meetup.

iceoryx2-bb/cxx/include/iox2/bb/into.hpp
elfenpiff: We just used it internally, so I would handle it like that. When we reach v1.0 we have to be more careful but then we would also enable some semver compliance checks in the CI.

iceoryx2-bb/cxx/include/iox2/bb/into.hpp
elfenpiff: Sorry, but no! Let's have one rule without any exceptions - because exceptions have this ugly behavior that they never come alone. Soon, there are other exceptions here and there, and it does not feel right somewhere else either - and then we start endless discussions again, delay valuable work for naming nonsense. Another argument is that if we have exceptions, then the exception becomes the new rule, and classes and structs follow random rules because we developers sometimes make mistakes, are confused, and then mix up coding styles - this happened to me constantly in the past.

Please do not deviate here either! Have one naming rule and follow it strictly.

Also backported C++ constructs! Most of the time, we use those only internally, and I think we get quickly used to them. So `lossy` or `Lossy`, `always_false_v` or `AlwaysFalseV` - at first it looks a bit weird, but after remembering our code style and using an IDE where in the description it is highlighted that it is the backported C++ construct, I got used to it.

iceoryx2-bb/cxx/include/iox2/bb/into.hpp
elfenpiff: I would vote for just `From`

iceoryx2-bb/cxx/include/iox2/bb/into.hpp
elfenpiff: Another argument against mixed styles is, even when we contain it in separate folders, we need to manage a more complex infrastructure with inherited clang-tidy files - which sometimes has its problems and we have to invest more of our scarce time.

elfenpiff: > Maybe something to discuss in the next developer meetup.

And again, this takes even more time when it comes to a transition phase for whatever we decide, where we need to add deprecated macros in one version and remove them in the next version. 

For our code base, we have one consistent naming rule!

iceoryx2-bb/cxx/include/iox2/bb/into.hpp
```diff
@@ -12,40 +12,58 @@
...
+struct ExtractIntoType<Lossy<T>> {
+    using TargetType = T;
 };
 } // namespace detail
 
+// Using a struct as impl, as free functions do not support partially specialized templates
+template <typename SourceType, typename DestinationType>
+struct From {
+    // AXIVION Next Construct AutosarC++19_03-A7.1.5 : 'auto' is only used for the generic implementation which will always result in a compile error
+    static auto from(const SourceType& value IOX2_MAYBE_UNUSED) noexcept {
```
ComicSansMS, line 47: `constexpr`?

iceoryx2-bb/cxx/include/iox2/legacy/detail/std_chrono_support.inl
```diff
@@ -17,28 +17,28 @@
...
+From<std::chrono::milliseconds, legacy::units::Duration>::from(const std::chrono::milliseconds& value) noexcept {
+    return legacy::units::Duration::fromMilliseconds(value.count());
 }
 
-inline units::Duration
-FromImpl<std::chrono::seconds, units::Duration>::fromImpl(const std::chrono::seconds& value) noexcept {
-    return units::Duration::fromSeconds(value.count());
+inline legacy::units::Duration
+From<std::chrono::seconds, legacy::units::Duration>::from(const std::chrono::seconds& value) noexcept {
+    return legacy::units::Duration::fromSeconds(value.count());
```
ComicSansMS, line 38: I guess these can also be `constexpr`.

internal/scripts/clang_tidy_scan.sh
```diff
@@ -114,7 +114,7 @@ ADDED_FILES=""
 FILE_LIST=""
 if [[ $FILES_MODE == true || $DIFF_TO_MAIN_MODE == true ]]; then
     if [[ $DIFF_TO_MAIN_MODE == true ]]; then
-        FILES_TO_SCAN=$(git diff --name-only --diff-filter=AM ${MAIN_ORIGIN}/main HEAD)
+        FILES_TO_SCAN=$(git diff --name-only --diff-filter=ACMRT ${MAIN_ORIGIN}/main HEAD)
```
ComicSansMS, line 117: Nice!

ComicSansMS (approved):

iceoryx2-bb/cxx/include/iox2/bb/into.hpp
elBoberido: Maybe my memories are fooling me, but I think we had issues because some implementations could not be `constexpr` and some tools complained. I can try to add `constexpr` again and wait until someone complains again.

elfenpiff (approved):

PR #1207, opened 2025-12-01
[#301] Move 'Duration' files to 'iox2/bb'

elBoberido:
## Notes for Reviewer

This PR moves `iox2/legacy/duration.hpp` and its dependencies to `iox2/bb/duration.hpp`. Furthermore, the namespace is adjusted to `iox2::bb` and the clang-tidy warnings are fixed.

* [x] Assign PR to reviewer

## PR Reviewer Reminder

* Public API is documented

## References

Relates to #301

### DIFF
--- a/iceoryx2-bb/cxx/tests/src/std_chrono_support_tests.cpp
+++ b/iceoryx2-bb/cxx/tests/src/std_chrono_support_tests.cpp
@@ -25,75 +25,75 @@ using namespace iox2::bb::duration_literals;
 constexpr uint64_t NANOSECS_PER_MILLISECOND = Duration::NANOSECS_PER_MILLISEC;
 constexpr uint64_t NANOSECS_PER_SECOND = Duration::NANOSECS_PER_SEC;
 
-class StdChrono_test : public Test { };
+class StdChronoTest : public Test { };
 
 // BEGIN CONSTRUCTION TESTS
 
 TEST(StdChrono_test, ConstructFromChronoMillisecondsZero) {
     ::testing::Test::RecordProperty("TEST_ID", "40b02547-8a9d-4ae6-90b2-72e76e5143f0");
     constexpr uint64_t EXPECTED_MILLISECONDS { 0U };
-    Duration sut = into<Duration>(std::chrono::milliseconds(EXPECTED_MILLISECONDS));
+    const Duration sut = into<Duration>(std::chrono::milliseconds(EXPECTED_MILLISECONDS));
     EXPECT_THAT(sut.as_nanos(), Eq(0U));
 }
 
 TEST(StdChrono_test, ConstructFromChronoMillisecondsLessThanOneSecond) {
     ::testing::Test::RecordProperty("TEST_ID", "ccbcd8df-c146-48ea-a9a0-40abaff31e68");
     constexpr uint64_t EXPECTED_MILLISECONDS { 44U };
-    Duration sut = into<Duration>(std::chrono::milliseconds(EXPECTED_MILLISECONDS));
+    const Duration sut = into<Duration>(std::chrono::milliseconds(EXPECTED_MILLISECONDS));
     EXPECT_THAT(sut.as_nanos(), Eq(EXPECTED_MILLISECONDS * NANOSECS_PER_MILLISECOND));
 }
 
 TEST(StdChrono_test, ConstructFromChronoMillisecondsMoreThanOneSecond) {
     ::testing::Test::RecordProperty("TEST_ID", "04313e7e-2954-4741-ad0e-6f5a9e5aebce");
     constexpr uint64_t EXPECTED_MILLISECONDS { 1001 };
-    Duration sut = into<Duration>(std::chrono::milliseconds(EXPECTED_MILLISECONDS));
+    const Duration sut = into<Duration>(std::chrono::milliseconds(EXPECTED_MILLISECONDS));
     EXPECT_THAT(sut.as_nanos(), Eq(EXPECTED_MILLISECONDS * NANOSECS_PER_MILLISECOND));
 }
 
 TEST(StdChrono_test, ConstructFromChronoMillisecondsMax) {
     ::testing::Test::RecordProperty("TEST_ID", "553d4ab1-ff8a-437e-8be8-8105e62850c6");
     constexpr uint64_t EXPECTED_MILLISECONDS { std::numeric_limits<int64_t>::max() };
-    Duration sut = into<Duration>(std::chrono::milliseconds(EXPECTED_MILLISECONDS));
+    const Duration sut = into<Duration>(std::chrono::milliseconds(EXPECTED_MILLISECONDS));
     EXPECT_THAT(sut.as_millis(), Eq(EXPECTED_MILLISECONDS));
 }
 
 TEST(StdChrono_test, ConstructFromNegativeChronoMillisecondsIsZero) {
     ::testing::Test::RecordProperty("TEST_ID", "181bc67d-1674-44eb-8784-710306efde30");
-    Duration sut = into<Duration>(std::chrono::milliseconds(-1));
+    const Duration sut = into<Duration>(std::chrono::milliseconds(-1));
     EXPECT_THAT(sut.as_nanos(), Eq(0U));
 }
 
 TEST(StdChrono_test, ConstructFromChronoNanosecondsZero) {
     ::testing::Test::RecordProperty("TEST_ID", "3044b00f-a765-417b-a16b-da01c16f7ed0");
     constexpr uint64_t EXPECTED_NANOSECONDS { 0U };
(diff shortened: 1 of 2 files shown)


iceoryx2-bb/cxx/include/iox2/bb/duration.hpp
```diff
@@ -0,0 +1,821 @@
...
+
+#include "iox2/legacy/attributes.hpp"
+#include "iox2/legacy/log/logstream.hpp"
+#include "iox2/legacy/type_traits.hpp"
+
+#include <cmath>
+
+namespace iox2 {
+namespace bb {
+class Duration;
```
elBoberido, line 26: If it is okay, I would omit the additional `unit` namespace. It was always an outlier in iceoryx_hoofs and it was the only class in that namespace. No hard feelings about this and I will add it if desired.

iceoryx2-bb/cxx/include/iox2/bb/duration.hpp
```diff
@@ -0,0 +1,821 @@
...
+    constexpr auto operator*=(const T& rhs) noexcept -> Duration&;
+
+    // END ARITHMETIC
+
+    // BEGIN CONVERSION
+
+    /// @brief returns the duration in nanoseconds
+    /// @note If the duration in nanoseconds is larger than an uint64_t can represent, it will be clamped to the
+    /// uint64_t max value.
+    constexpr auto to_nanoseconds() const noexcept -> uint64_t;
```
elBoberido: The class was modeled after the Rust `Duration` but the naming of the methods is a bit off. I'd like to rename the `to_` methods to `as_`. We have something similar with the `SemanticString::as_string` method.

Furthermore, I would like to shorten the name a little bit and use `nanos`, `micros`, etc. instead of `nanoseconds`, `microseconds`, etc. Also no hard feelings about this but if this change is not desired, I would rename the newly introduced `subsec_nanos`, etc. to `subsecond_nanoseconds`, etc. in order to have the same naming style.

The main reason for this is to get more code into one line without hitting the 120 character limit.

ComicSansMS: My main gripe with that name is that `as_nanoseconds` still returns a `uint64_t`. With that name I would expect it to return a `nanoseconds` type, like `std::chrono::nanoseconds`.

iceoryx2-bb/cxx/include/iox2/bb/duration.hpp
```diff
@@ -0,0 +1,821 @@
...
+
+/// @brief Less than operator
+/// @param[in] lhs is the left hand side of the comparison
+/// @param[in] rhs is the right hand side of the comparison
+/// @return true if duration is less than rhs
+// AXIVION Next Line AutosarC++19_03-A8.4.7 : Each argument is larger than two words
+constexpr auto operator<(const Duration& lhs, const Duration& rhs) noexcept -> bool {
+    return (lhs.m_seconds < rhs.m_seconds)
+           || ((lhs.m_seconds == rhs.m_seconds) && (lhs.m_nanoseconds < rhs.m_nanoseconds));
+}
```
ComicSansMS, line 361: There is a neat little trick to do comparisons like this using [`std::tie`](https://en.cppreference.com/w/cpp/utility/tuple/tie):

`std::tie(lhs.m_seconds, lhs.m_nanoseconds) < std::tie(rhs.m_seconds, rhs.m_nanoseconds);`

It requires pulling in all of `<tuple>`, so it may not be a good fit here, but I thought I'd share it in case you haven't seen it before.

iceoryx2-bb/cxx/include/iox2/bb/duration.hpp
```diff
@@ -0,0 +1,821 @@
...
+    // one bit of the result represents 2^32 nanoseconds;
+    // just shifting left by 32 bits would result in an overflow, therefore blocks of full seconds must be extracted of
+    // the result;
+    // this cannot be done by dividing through NANOSECS_PER_SEC, since that one is base 1_000_000_000 and the result is
+    // base 2^32, therefore the least common multiple can be used to get blocks of full seconds represented with the LSB
+    // representing 2^32 nanoseconds;
+    // this can then safely be converted to seconds as well as nanoseconds without loosing precision
+
+    // least common multiple of 2^32 and NANOSECONDS_PER_SECOND;
+    // for the following calculation it is not important to be the least common multiple, any common multiple will do
```
ComicSansMS, line 667: I'll take your word for it that this part works... 😅

iceoryx2-bb/cxx/include/iox2/bb/duration.hpp
elfenpiff: I am fine with the shortened variant.

elBoberido: @ComicSansMS then `as_nanos` should help to prevent the confusion :)

The idea behind our constructs was to not rely directly on the STL, since our customers would have to also certify those parts of the STL. For interoperability with the STL, we provide the `std_chrono_support.hpp`, which implements `into<std::chrono::*>`. This decouples the STL usage from our functional headers and it's up on the user to decide it a dependency to `chrono` is desired, together with the costs of certification.

iceoryx2-bb/cxx/tests/src/duration_tests.cpp
```diff
@@ -12,24 +12,20 @@
 //
```
ComicSansMS, line 12: Our usual naming convention for test names seems to be to use snake_case for the test names and CamelCase for the test case name, while this file (and the chrono_support_tests below) uses CamelCase throughout for both.

elfenpiff: I agree. Could you please convert the test names into snake case so that it is consistent with all other test cases.

iceoryx2-bb/cxx/tests/src/std_chrono_support_tests.cpp
```diff
@@ -12,90 +12,89 @@
...
 
 constexpr uint64_t NANOSECS_PER_MILLISECOND = Duration::NANOSECS_PER_MILLISEC;
 constexpr uint64_t NANOSECS_PER_SECOND = Duration::NANOSECS_PER_SEC;
 
-class StdChrono_test : public Test { };
+class StdChronoTest : public Test { };
 
 // BEGIN CONSTRUCTION TESTS
 
 TEST(StdChrono_test, ConstructFromChronoMillisecondsZero) {
```
elfenpiff, line 32: Same here, please use lower_snake_case for test names

iceoryx2-bb/cxx/include/iox2/bb/duration.hpp
elBoberido: That's neat. It's a pity that tuples are a library feature in C++

iceoryx2-bb/cxx/tests/src/duration_tests.cpp
elBoberido: Good catch. Since this is imported code, would you be fine by leaving this as is, before legacy `string`, `optional` and `expected` in the public API is replaced by their iceoryx2 counterparts? 
Alternatively, I would move the test files back to the legacy folder for the time being.

ComicSansMS: I'm not sure I understand. How is the replacing of `optional` et al. going to impact `Duration`? Or do you want to replace `Duration` as well eventually?

elBoberido: I meant that I would like to replace the legacy `string`, `optional` and `expected` in the public API with the their iceoryx2 counterparts before the release. Putting effort into renaming all the test names does not help with replacing the other legacy components and there is not that much time until the release. Since the test names doe not have any effect, except consistency, I would like to postpone this work until the important, public facing changes are finished. This means either leaving the test file where it is right now or moving the test file back to the legacy folder.

elBoberido: @ComicSansMS I assume the :+1: means I can keep the file where it is and adjust the test names later :)

In this case, I will continue to move the tests together with the code out of the legacy folders and fix the test names once there are no legacy components in the public API. I also added this as action item to the #301 issue.

ComicSansMS: That's fine with me, but I suspect @elfenpiff might have an opinion here as well.

iceoryx2-bb/cxx/include/iox2/bb/duration.hpp
elBoberido: I was quite confident when I wrote that code ... and we have test :sweat_smile: 

Maybe a mathematician can double check the formula @FerdinandSpitzschnueffler @elfenpiff

Issue #940, opened 2025-08-11
C++ `expected` vocabulary data type


ComicSansMS:
## Brief feature description

Provide as part of the C++ base library an implementation of a vocabulary type that can act as a drop-in replacement for `std::expected`. This targets deployments where no safety certified C++17 standard library is available to provide `expected`.

## Action items

- [x] thin wrapper around `legacy::expected` with `std::expected` compatible API
- [ ] proper implementation without `legacy::expected`

elfenpiff:
One side constraint. The `expected` must compile with C++14.

PR #1236, opened 2025-12-09
[#940] Introduce new 'Expected'

elBoberido:
## Notes for Reviewer

This PR adds an `Expected` which closely follows the API of `std::expected`. For now, it is just a thin wrapper around `legacy::expected` from iceoryx_hoofs. If there is time before the v0.8.0 release, the dependency to `legacy::expected` will be removed by a native implementation.

Some basic tests are added to verify that the wrapper work as intended.

* [x] Assign PR to reviewer

## PR Reviewer Reminders

* Public API is documented

## References

Relates to #940

### DIFF
--- a/doc/release-notes/iceoryx2-unreleased.md
+++ b/doc/release-notes/iceoryx2-unreleased.md
@@ -27,6 +27,8 @@
   [#865](https://github.com/eclipse-iceoryx/iceoryx2/issues/865)
 * Add a C++ string container type with fixed compile-time capacity
   [#938](https://github.com/eclipse-iceoryx/iceoryx2/issues/938)
+* Add new C++ `Expected` vocabulary data type
+  [#940](https://github.com/eclipse-iceoryx/iceoryx2/issues/940)
 * Add a C++ vector container type with fixed compile-time capacity
   [#951](https://github.com/eclipse-iceoryx/iceoryx2/issues/951)
 * Use `epoll` instead of `select` for the `WaitSet` on Linux


iceoryx2-bb/cxx/include/iox2/container/expected.hpp
elBoberido, line 1: @ComicSansMS @elfenpiff @FerdinandSpitzschnueffler this is just a thin wrapper around the `legacy::expected` in order to have a minimal C++23 std::expected compatible API.

The idea is to be able to swap our `Expected` out with the `std::expected` or any other 3rd-party implementation which is compatible with the `std::expected`.

I'll add some test for each method but I guess I will not have time to have full re-implementation of the `expected`, so it will be just a thin wrapper for the v0.8.0 release. We can get rid of the `legacy::expected` without breaking the API later on.

Please check if the minimal implementation is fine. 

Oh, and shall we keep the name `Expected` or rename it to `Result` ;)

iceoryx2-bb/cxx/include/iox2/container/expected.hpp
```diff
@@ -0,0 +1,155 @@
...
+
+#include "iox2/legacy/expected.hpp"
+#include <type_traits>
+
+namespace iox2 {
+namespace container {
+
+struct InPlaceT { };
+// NOLINTNEXTLINE(readability-identifier-naming), for consistency with C++23 code using std::expected
+constexpr InPlaceT in_place {};
```
elBoberido, line 25: This is analogous to `iox2::container::nullopt`. Shall we change all constants to upper camel case?

ComicSansMS: Analogous to `optional`, if this is intended as a drop-in replacement for [the `std` version](https://en.cppreference.com/w/cpp/utility/in_place.html), it needs to remain lower case.

This same type would also be used for `optional` (and possibly other types), so it would move to a different header eventually.

elBoberido: Yeah, I would move it to a different header. It is currently just for convenience in this header.

Out of curiosity, couldn't we just use `constexpr auto NULLOPT = std::nullopt;`? Bear with me, I'm more a manager than a coder nowadays :sweat_smile:

ComicSansMS: That's what we do when opting into using the standard facilities. The assumption was that the implementation we deploy on either won't have a `std::nullopt`, or won't have it as part of the safety-critical subset.

iceoryx2-bb/cxx/include/iox2/container/expected.hpp
elfenpiff: Please keep the name.

iceoryx2-bb/cxx/include/iox2/container/expected.hpp
elfenpiff: To establish compatibility, wouldn't it also be possible to create aliases with the `std` syntax. So we use `InPlaceT` here to follow our guidelines strictly and add a compatibility header where we establish the syntax with the target library.

At former places, we also had a different style guide that deviated from the STL, and with such an approach, we can follow our style and would also provide a migration for any other style, because I bet that the base libraries out there do not always follow STL naming conventions.

elBoberido: @elfenpiff once everything is ported to the new `Expected` I'll add cmake options to switch to the C++ counterparts. Once that works, I'll check if your proposal works.

elfenpiff (approved):

Issue #1234, opened 2025-12-09
Replace legacy types in public API with iceoryx2 counterparts


elBoberido:
## Brief feature description

iceoryx2 uses in its API still some legacy classes from `iceoryx_hoofs` which have some counterparts in the iceoryx2 repo.

## Detailed information

- [ ] replace `legacy::string` with `StaticString`
- [ ] replace `legacy::optional` with `Optional`
- [ ] replace `legacy::expected` with `Expected`

PR #1237, opened 2025-12-09
[#1234] Replace 'legacy::expected' with new 'Expected'

elBoberido:

This PR replaces the legacy `expected` from iceoryx_hoofs with the new `Expected` in the public API.

* [x] Assign PR to reviewer

## PR Reviewer Reminders

* Public API is documented

## References

Relates to #1234

### DIFF
--- a/doc/release-notes/iceoryx2-unreleased.md
+++ b/doc/release-notes/iceoryx2-unreleased.md
@@ -56,8 +56,6 @@
   [#1223](https://github.com/eclipse-iceoryx/iceoryx2/issues/1223)
 * Add socket directory configuration in platform
   [#1232](https://github.com/eclipse-iceoryx/iceoryx2/issues/1232)
-* Replace legacy types in public API with iceoryx2 counterparts
-  [#1234](https://github.com/eclipse-iceoryx/iceoryx2/issues/1234)
 
 ### Bugfixes
 
@@ -113,6 +111,8 @@
     [#1133](https://github.com/eclipse-iceoryx/iceoryx2/issues/1133)
 * Support C++14 for the C++ Bindings
     [#1167](https://github.com/eclipse-iceoryx/iceoryx2/issues/1167)
+* Replace legacy types in public API with iceoryx2 counterparts
+  [#1234](https://github.com/eclipse-iceoryx/iceoryx2/issues/1234)
 
 ### Workflow
 
@@ -295,28 +295,56 @@ CMake package.
 1. **C++:** Replace `iox::optional` from `iceoryx_hoofs` with
    `iox2::container::Optional`
 
-  The new `Optional` in iceoryx2 has a reduced API compared to the one from
-  `iceroyx_hoofs`. The functional interface, which deviated from the STL was
-  removed.
+   The new `Optional` in iceoryx2 has a reduced API compared to the one from
+   `iceroyx_hoofs`. The functional interface, which deviated from the STL was
+   removed.
 
-  ```cpp
-  // old
-  ret_val.and_then([](auto& val) { /* do something with val */ })
-         .or_else([]() { /* do something else */ });
+   ```cpp
+   // old
+   ret_val.and_then([](auto& val) { /* do something with val */ })
+          .or_else([]() { /* do something else */ });
 
-  // new
-  if (ret_val.has_value()) {
-    // do something with ret_val.value()
-  } else {
-    // do something else
-  }
+   // new
+   if (ret_val.has_value()) {
+     // do something with ret_val.value()
+   } else {
+     // do something else
+   }
 
-  // old
-  auto val = ret_val.expect("There should be a value");
+   // old
+   auto val = ret_val.expect("There should be a value");
 
-  // new
-  if (!ret_val.has_value()) {
-    // error handling or terminate
-  }
-  auto val = ret_val.value();
-  ```
+   // new
+   if (!ret_val.has_value()) {
+     // error handling or terminate
+   }
+   auto val = ret_val.value();
+   ```
+
+9. Replace `iox::expected` from `iceoryx_hoofs` with `iox2::container::Expected`
+
+   The new `Expected` in iceoryx2 has a reduced API compared to the one from
+   `iceroyx_hoofs`. The functional interface, which deviated from the STL was
+   removed.
+
+   ```cpp
+   // old
+   ret_val.and_then([](auto& val) { /* do something with val */ })
+          .or_else([]() { /* do something else */ });
+
+   // new
+   if (ret_val.has_value()) {
+     // do something with ret_val.value()
+   } else {
+     // do something with ret_val.error()
+   }
+
+   // old
+   auto val = ret_val.expect("There should be a value");
+
+   // new
+   if (!ret_val.has_value()) {
+     // error handling based on ret_val.error() or terminate
+   }
+   auto val = ret_val.value();
+   ```


examples/cxx/health_monitoring/src/central_daemon.cpp
```diff
@@ -31,7 +31,7 @@ auto main() -> int {
     auto service_name_2 = ServiceName::create("service_2").expect("");
 
     auto node = NodeBuilder()
-                    .name(NodeName::create("central daemon").expect(""))
+                    .name(NodeName::create("central daemon").value())
```
elBoberido: @ComicSansMS @elfenpiff @FerdinandSpitzschnueffler with the minimal STL compatible `Expected`, we lose the functional interface, similar to the new `Optional`. The question is now how to handle this cases where we previously used just `expect("foo")`?

We can just call `value` and with our implementation we can just abort but other implementations might trigger UB. In this simple case, we could just do the proper error handling, but there are other cases where this becomes quite cumbersome.

An idea would be to have a helper function like this
```
template <typename T, typename E>
auto unwrap_or_panic(Expected<T, E> exp) -> T;
```

What do you think? Ideas welcome.

iceoryx2-cxx/src/node_name.cpp
```diff
@@ -79,23 +84,23 @@ void NodeName::drop() noexcept {
...
 
     iox2_node_name_h handle {};
     const auto ret_val = iox2_node_name_new(nullptr, value, value_len, &handle);
     if (ret_val == IOX2_OK) {
-        return iox2::legacy::ok(NodeName { handle });
+        return { container::in_place, NodeName { handle } };
     }
 
-    return iox2::legacy::err(iox2::bb::into<bb::SemanticStringError>(ret_val));
+    return { container::unexpect, iox2::bb::into<bb::SemanticStringError>(ret_val) };
```
elBoberido: @ComicSansMS @elfenpiff @FerdinandSpitzschnueffler these helper structs are unfortunately not possible anymore, since the conversion methods would have to be part of the respective `expected` implementation.

`ok` and `err` felt a little bit more obvious for the happy and sad path. Maybe it is just a matter of getting used to `in_place` and `unexpect`. An option would be to introduce `IOX2_OK` and `IOX2_ERR` macros but unfortunately, there is already a `IOX2_OK` define from the C bindings.

What is your opinion?

examples/cxx/health_monitoring/src/central_daemon.cpp
ComicSansMS: > We can just call value and with our implementation we can just abort but other implementations might trigger UB.

What did those other implementations do before when calling `expect()`? My intention was that the `value` methods in the new optional would be the drop-in replacement for `expect()`.

iceoryx2-cxx/src/node_name.cpp
ComicSansMS: The STL expected provides [`std::unexpected`](https://en.cppreference.com/w/cpp/utility/expected/unexpected.html), which I guess is the equivalent to `legacy::err`? I think that would be viable here.

For the `ok` case, `std::expected` provides an implicit conversion constructor to the ok value (number 6 in [this list](https://en.cppreference.com/w/cpp/utility/expected/expected.html)). The `in_place` is only required in cases where you need to emplace construct from a constructor with multiple arguments. I consider this to be a rare case, in fact it is so rare that I did not include the `in_place` constructor for the current version of `Optional`.

examples/cxx/health_monitoring/src/central_daemon.cpp
elBoberido: It did exactly what the new `Optional` is doing with `value` if it does not contain a value, with the notable exception that one got an error string on the terminal before the `abort`. When using the std::optional, one would get an exception.

iceoryx2-cxx/src/node_name.cpp
elBoberido: Nice. I somehow overlooked the implicit conversion for the happy path. The `std::unexpected` is exactly the same implementation like `legacy::err` ... just a little bit more to type. I will add those to #940

What is your opinion on having a `template <typename E> using err = unexpect<T>;` alias?

examples/cxx/health_monitoring/src/central_daemon.cpp
ComicSansMS: So the main difference is that `expect` will provide a custom message before terminating. Otherwise using `value` is functionally equivalent. In particular, there is no way that `value` will trigger UB?

iceoryx2-cxx/src/node_name.cpp
ComicSansMS: > What is your opinion on having a template <typename E> using err = unexpect<T>; alias?

I suppose you really mean `unexpected` there, not [`std::unexpect`](https://en.cppreference.com/w/cpp/utility/expected/unexpect_t.html) (the value corresponding to the constructor tag type). I guess there is no harm in adding that, other than that it is not the idiomatic way of writing this in C++. It won't impact the replacability of the type at least, if it's provided by the same header that provides the `iox2::Expected` typedef under the use-std-expected opt-in.

examples/cxx/health_monitoring/src/central_daemon.cpp
elBoberido: No, there is a check and `abort` will be called. I also replaced `expect("")` with `value()` in the last commit. Was just curious if there is another idiomatic way with `std::expected` or `std::optional`

iceoryx2-cxx/src/node_name.cpp
elBoberido: I will leave the `Unexpected` for now and if we notice that our fingers hurt from the long typing, we can introduce the shortcut :)

examples/cxx/health_monitoring/src/central_daemon.cpp
ComicSansMS: > Was just curious if there is another idiomatic way with `std::expected` or `std::optional`

No, unfortunately not, as far as I'm aware.

iceoryx2-cxx/src/node_name.cpp
elBoberido: Okay, the handling with `Unexpected` is a bit cumbersome with C++14. I'll introduce the `err` free function to prevent having to specify the template parameter.

ComicSansMS: Ah, yes, I forgot that we need deduction guides for this. `err` free function sounds good!

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
```diff
@@ -331,6 +330,20 @@ class Optional {
     constexpr auto reset() noexcept -> void {
         m_value.reset();
     }
+
+  private:
+    friend auto operator==(const Optional<T>& lhs, NulloptT /* unused */) noexcept -> bool {
```
elfenpiff, line 335: This looks so wrong. Is this a free function defined in a templatized class?

Can we please not do this! The first thing I thought was, is this a friend method? What does this mean? And why only for `NulloptT` and not for the other comparision functions.

elfenpiff (commented):
Please apply also all API changes to all C++ examples in the iceoryx2 book

elfenpiff (approved):

iceoryx2-bb/cxx/include/iox2/container/optional.hpp
elfenpiff: Ah, forget it ... I read abou the hidden friends idiom. Lets role with it.

Issue #1244, opened 2025-12-11
Make compilation of tunnel backends opt-in


orecham:
## Brief Description

Compiling tunnel backends pulls in a lot of external dependencies. This increases build times for the default build via:

```rust
cargo build --workspace
```

Since tunneling over networks is optional functionality and also the choice of backend is better left up to the user, tunnel backends should not be compiled in by default.

PR #1245, opened 2025-12-11
[#1244] Make tunnel backend compilation opt-in

orecham:
## Notes for Reviewer

The zenoh tunnel backend `iceoryx2-tunnel-zenoh` is removed from the workspace and made an optional dependency to crates that use it (`iceoryx2-cli`). 

When the corresponding feature flag is enabled, the backend is compiled.
Building the workspace with  `cargo build --workspace`  will not build the backend.

* [ ] Assign PR to reviewer

## PR Reviewer Reminders

* Public API is documented

## References

Closes #

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,6 @@ members = [
     "iceoryx2-tunnel/backend",
     "iceoryx2-tunnel/conformance-tests",
     "iceoryx2-tunnel/tunnel",
-    "iceoryx2-tunnel/zenoh",
     "iceoryx2-tunnel/end-to-end-tests",
 
     "iceoryx2-userland/record-and-replay",
@@ -52,6 +51,7 @@ members = [
 
     "component-tests/rust",
 ]
+exclude = ["iceoryx2-tunnel/zenoh"]
 
 [workspace.package]
 categories = ["network-programming"]
@@ -104,7 +104,6 @@ iceoryx2-tunnel-end-to-end-tests = { version = "0.7.0", path = "iceoryx2-tunnel/
 iceoryx2-tunnel = { version = "0.7.0", path = "iceoryx2-tunnel/tunnel" }
 iceoryx2-tunnel-backend = { version = "0.7.0", path = "iceoryx2-tunnel/backend" }
 iceoryx2-tunnel-conformance-tests = { version = "0.7.0", path = "iceoryx2-tunnel/conformance-tests" }
-iceoryx2-tunnel-zenoh = { version = "0.7.0", path = "iceoryx2-tunnel/zenoh" }
 iceoryx2-userland-record-and-replay = { version = "0.7.0", path = "iceoryx2-userland/record-and-replay" }
 
 anyhow = { version = "1.0.86" }
@@ -154,7 +153,6 @@ windows-sys = { version = "0.48.0", features = [
     "Win32_System_SystemServices",
     "Win32_System_ProcessStatus",
 ] }
-zenoh = { version = "1.3.4", features = ["default", "unstable"] }
 
 [profile.release]
 strip = true
--- a/iceoryx2-cli/Cargo.toml
+++ b/iceoryx2-cli/Cargo.toml
@@ -39,8 +39,11 @@ name = "iceoryx2_cli"
 path = "lib/src/lib.rs"
 
 [features]
-default = ["tunnel_zenoh"]
-tunnel_zenoh = ["iceoryx2-tunnel-zenoh"]
+default = []
+tunnel_zenoh = [
+  "iceoryx2-tunnel-zenoh",
+  "zenoh"
+]
 
 [dependencies]
 iceoryx2-services-discovery = { workspace = true }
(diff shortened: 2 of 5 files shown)


iceoryx2-cli/Cargo.toml
```diff
@@ -72,8 +75,8 @@ dialoguer = { workspace = true }
 iceoryx2-tunnel = { workspace = true }
 # Zenoh does not build on FreeBSD
 [target.'cfg(not(target_os = "freebsd"))'.dependencies]
-iceoryx2-tunnel-zenoh = { workspace = true, optional = true }
-zenoh = { workspace = true }
+iceoryx2-tunnel-zenoh = { path = '../iceoryx2-tunnel/zenoh', optional = true }
```
elBoberido, line 78: This will potentially break for pubishing

Cargo.toml
```diff
@@ -154,7 +153,6 @@ windows-sys = { version = "0.48.0", features = [
     "Win32_System_SystemServices",
     "Win32_System_ProcessStatus",
 ] }
-zenoh = { version = "1.3.4", features = ["default", "unstable"] }
```
elBoberido, line 157: Could this be kept by making the dependency optional?

Cargo.toml
```diff
@@ -104,7 +104,6 @@ iceoryx2-tunnel-end-to-end-tests = { version = "0.7.0", path = "iceoryx2-tunnel/
 iceoryx2-tunnel = { version = "0.7.0", path = "iceoryx2-tunnel/tunnel" }
 iceoryx2-tunnel-backend = { version = "0.7.0", path = "iceoryx2-tunnel/backend" }
 iceoryx2-tunnel-conformance-tests = { version = "0.7.0", path = "iceoryx2-tunnel/conformance-tests" }
-iceoryx2-tunnel-zenoh = { version = "0.7.0", path = "iceoryx2-tunnel/zenoh" }
```
elBoberido, line 107: This might also be problematic for publishing

Cargo.toml
orecham: @elBoberido I tried that, but even if the dependency is optional, it is built when running either of:

```console
cargo build
cargo build --workspace
```

With the approach in this PR, the dependency is only built when building the crate that uses it with the corresponding feature activated.

iceoryx2-cli/Cargo.toml
orecham: @elBoberido ~~Maybe but it's not clear to me. But since the crates are publish individually, not as a workspace,  maybe it works?~~

EDIT: I looked into it, you are probably correct.

orecham: Maybe we can do this in our script:
1. Publish `iceoryx2-tunnel-zenoh`
2. Update this Cargo.toml to depend on the uploaded version in crates.io
3. Publish this crate

elBoberido: I think I have seen another solution somewhere. Need to find it.

iceoryx2-tunnel/zenoh/Cargo.toml
```diff
@@ -1,28 +1,34 @@
...
-iceoryx2-tunnel-backend = { workspace = true }
-iceoryx2-bb-log = { workspace = true }
+iceoryx2 = { path = "../../iceoryx2" }
+iceoryx2-tunnel-backend = { path = "../backend" }
+iceoryx2-bb-log = { path = "../../iceoryx2-bb/log" }
 
-serde_json = { workspace = true }
-zenoh = { workspace = true }
+serde_json = { version = "1.0", default-features = false, features = ["alloc"] }
+zenoh = { version = "1.3.4", features = ["default", "unstable"] }
```
dkroenke, line 29: Can you try to update zenoh to 1.7.1? Maybe there is no effort in this otherwise we do it later.

orecham: @dkroenke Updated it in https://github.com/eclipse-iceoryx/iceoryx2/pull/1299

iceoryx2-cli/Cargo.toml
elBoberido: @orecham I might have found a nice middle ground. With

```toml
default-members = [
    "iceoryx2",
]
```

and adding the `preserve_order` feature to the `toml` dependency (only for Rust 1.83), a simple `cargo build` will only build the `iceoryx2` crate, which is pretty fast. About 14 seconds. It might make sense to add `iceoryx2-ffi/c`, since that one does not add much compilation time and is useful when using cmake right after cargo. A `cargo build --workspace` still pulls in Zenoh and the compile time increases massively.

I think your approach is the only one that works with `cargo build --workspace`. I couldn't find the solution I was thinking of. Maybe a hallucination :sweat_smile: 

In theory, during publishing, the path should be ignored and only the version should be used. But there are other downsides. We would not be able to do a `cargo publish --dry-run --workspace` anymore. This would increase the risk of introducing circular dependencies which will then only be noticed during an actual release.

What do you think about the `default-members` approach? Shall we have that as first step?

elBoberido: I also have another idea, which at least would make the `cargo publish --dry-run --workspace` go away.

Not sure how feasible this would be, but instead of having all available backend in the `iox2-tunnel` cli tool, how about making the `iox2-tunnel` tool act like the `iox2` tool and just offer to launch anything with `iox2-tunnel-*`?

With this, one could move the `iox2-tunnel-zenoh` cli tool into `iceoryx2-tunnel/zenoh` folder and nothing in the workspace would depend on something in `iceoryx2-tunnel/zenoh` anymore. It would even be possible to easily move that part to another repo, if desired.

orecham:
A different approach will be used to tackle this. A separate PR will be opened.

PR #1247, opened 2025-12-11
[#1234] Replace legacy string with static string

FerdinandSpitzschnueffler:
## Notes for Reviewer

This PR replaces the legacy string from the iceoryx_hoofs import with the new StaticString from iceoryx2 in the public API.

For the StaticString, some additional methods were implemented to ease the port.

* [x] Assign PR to reviewer

## PR Reviewer Reminders

* Public API is documented

## References

Relates to #1234

### DIFF
--- a/iceoryx2-bb/cxx/include/iox2/bb/detail/path_and_file_verifier.hpp
+++ b/iceoryx2-bb/cxx/include/iox2/bb/detail/path_and_file_verifier.hpp
@@ -241,7 +241,7 @@ inline auto PathAndFileVerifier<StringCapacity>::does_end_with_path_separator(
         return false;
     }
     // AXIVION Next Construct AutosarC++19_03-A3.9.1: Not used as an integer but as actual character
-    const char last_character { name.unchecked_access()[name.size() - 1U] };
+    const char last_character { *name.code_units().back_element() };
 
     // NOLINTNEXTLINE(readability-use-anyofallof)
     for (const auto separator : platform::IOX2_PATH_SEPARATORS) {
--- a/iceoryx2-bb/cxx/include/iox2/bb/file_name.hpp
+++ b/iceoryx2-bb/cxx/include/iox2/bb/file_name.hpp
@@ -51,13 +51,8 @@ class FileName : public SemanticString<FileName,
 namespace detail {
 inline auto file_name_does_contain_invalid_characters(
     const container::StaticString<platform::IOX2_MAX_FILENAME_LENGTH>& value) noexcept -> bool {
-    const auto value_size = value.size();
-
-    for (uint64_t i { 0 }; i < value_size; ++i) {
-        // AXIVION Next Construct AutosarC++19_03-A3.9.1: Not used as an integer but as actual character
-        // NOLINTNEXTLINE(readability-identifier-length)
-        const char c { value.unchecked_access()[i] };
-
+    // NOLINTNEXTLINE(readability-use-anyofallof) not yet supported in all compilers
+    for (const char c : value.unchecked_access()) { // NOLINT(readability-identifier-length)
         const bool is_small_letter { ASCII_A <= c && c <= ASCII_Z };
         const bool is_capital_letter { ASCII_CAPITAL_A <= c && c <= ASCII_CAPITAL_Z };
         const bool is_number { ASCII_0 <= c && c <= ASCII_9 };
--- a/iceoryx2-bb/cxx/include/iox2/bb/file_path.hpp
+++ b/iceoryx2-bb/cxx/include/iox2/bb/file_path.hpp
@@ -53,13 +53,8 @@ namespace detail {
 inline auto
 file_path_does_contain_invalid_characters(const container::StaticString<platform::IOX2_MAX_PATH_LENGTH>& value) noexcept
     -> bool {
-    const auto value_size = value.size();
-
-    for (uint64_t i { 0 }; i < value_size; ++i) {
-        // AXIVION Next Construct AutosarC++19_03-A3.9.1: Not used as an integer but as actual character
-        // NOLINTNEXTLINE(readability-identifier-length)
-        const char c { value.unchecked_access()[i] };
-
+    // NOLINTNEXTLINE(readability-identifier-length)
+    for (const char c : value.unchecked_access()) {
         const bool is_small_letter { ASCII_A <= c && c <= ASCII_Z };
         const bool is_capital_letter { ASCII_CAPITAL_A <= c && c <= ASCII_CAPITAL_Z };
         const bool is_number { ASCII_0 <= c && c <= ASCII_9 };
--- a/iceoryx2-bb/cxx/include/iox2/bb/semantic_string.hpp
+++ b/iceoryx2-bb/cxx/include/iox2/bb/semantic_string.hpp
@@ -29,10 +29,10 @@ enum class SemanticStringError : uint8_t {
 };
 
 template <uint64_t Capacity>
-using DoesContainInvalidCharacter = bool (*)(const typename container::StaticString<Capacity>& value);
+using DoesContainInvalidCharacter = bool (*)(const container::StaticString<Capacity>& value);
 
 template <uint64_t Capacity>
-using DoesContainInvalidContent = bool (*)(const typename container::StaticString<Capacity>& value);
+using DoesContainInvalidContent = bool (*)(const container::StaticString<Capacity>& value);
 
 /// @brief The SemanticString is a string which has an inner syntax and restrictions
 ///         to valid content. Examples are for instance
--- a/iceoryx2-bb/cxx/include/iox2/container/static_string.hpp
+++ b/iceoryx2-bb/cxx/include/iox2/container/static_string.hpp
@@ -29,10 +29,10 @@
 namespace iox2 {
 namespace bb {
 template <uint64_t Capacity>
-using DoesContainInvalidCharacter = bool (*)(const typename container::StaticString<Capacity>& value);
+using DoesContainInvalidCharacter = bool (*)(const container::StaticString<Capacity>& value);
 
 template <uint64_t Capacity>
-using DoesContainInvalidContent = bool (*)(const typename container::StaticString<Capacity>& value);
+using DoesContainInvalidContent = bool (*)(const container::StaticString<Capacity>& value);
 
 template <typename, uint64_t Capacity, DoesContainInvalidContent<Capacity>, DoesContainInvalidCharacter<Capacity>>
 class SemanticString;
@@ -213,10 +213,9 @@ class StaticString {
             if (index > m_parent->m_size) {
                 return false;
             }
-            auto number_of_characters_to_move = m_parent->m_size - index;
-            std::memmove(
-                &m_parent->m_string[index + sub_str_size], &m_parent->m_string[index], number_of_characters_to_move);
-            std::memcpy(&m_parent->m_string[index], sub_str->m_string, static_cast<size_t>(sub_str_size));
+            std::copy_backward(
+                &m_parent->m_string[index], &m_parent->m_string[m_parent->m_size], &m_parent->m_string[new_size]);
+            std::copy(&sub_str->m_string[0], &sub_str->m_string[sub_str_size], &m_parent->m_string[index]);
 
             m_parent->m_string[new_size] = '\0';
             m_parent->m_size = new_size;
@@ -691,7 +690,7 @@ class StaticString {
 
         auto const length = std::min(static_cast<uint64_t>(count), m_size - pos);
         StaticString sub_str;
-        std::memcpy(sub_str.m_string, &m_string[pos], length);
+        std::copy(&m_string[pos], &m_string[pos + length], &sub_str.m_string[0]);
         sub_str.m_string[length] = '\0';
         sub_str.m_size = length;
         return sub_str;


iceoryx2-bb/cxx/include/iox2/container/static_string.hpp
```diff
@@ -343,6 +358,37 @@ class StaticString {
         return ret;
     }
 
+    /// Constructs a StaticString from a C-string literal. Users must ensure that the input string represents a valid
```
FerdinandSpitzschnueffler: @ComicSansMS Could you check the added API for the StaticString and if it follows your ideas?

iceoryx2-ffi/c/src/api/attribute.rs
```diff
@@ -68,9 +68,6 @@ pub unsafe extern "C" fn iox2_attribute_key(
             buffer.cast(),
             copied_key_length,
         );
-        copied_key_length
-    } else {
-        0
```
elBoberido, line 73: I'm wondering why this is removed. It might be useful to some of our users.

iceoryx2-bb/cxx/include/iox2/container/static_string.hpp
```diff
@@ -343,6 +358,37 @@ class StaticString {
...
+                break;
+            }
+            ret.push_back(character);
+        }
+        return ret;
+    }
+
+    /// Constructs a StaticString from the first `count` characters of a null terminated C-style string.
+    /// This unchecked function allows for uncontrolled memory access. Users of this must ensure that the input string
+    /// is properly null terminated and represents a valid UTF-8 encoding.
```
ComicSansMS: I feel the description should be more explicit. We take only up to `count` characters or truncate if the string is shorter.

Technically, null termination is not required if the string length does not exceed the `count`.

An important subtlety of truncating functions is that the *truncated* string needs to represent a valid UTF-8 encoding, not the input string. Since the truncation could slice a multi-byte encoded code point.

iceoryx2-bb/cxx/include/iox2/container/static_string.hpp
```diff
@@ -454,6 +500,22 @@ class StaticString {
         return !(lhs == rhs);
     }
 
+    friend auto operator<(StaticString const& lhs, StaticString const& rhs) -> bool {
+        return lhs.compare(rhs) < 0;
+    }
```
ComicSansMS: A not-so-obvious property of the UTF-8 encoding is that this actually orders strings lexicographically _by code points_. This may or may not be worth pointing out in a comment, but I think it's pretty cool.

iceoryx2-bb/cxx/include/iox2/container/static_string.hpp
```diff
@@ -477,18 +539,132 @@ class StaticString {
         return ret;
     }
 
+    /// Creates a substring containing the characters from pos until count; if pos+count is greater than the
+    /// size of the original string the returned substring only contains the characters from pos until size().
+    ///
+    /// @return an Optional containing the substring
+    ///         nullopt if pos is greater than the size of the original string
```
ComicSansMS: This should probably move into `ConstAccessorCodeUnits`. In that case, the function will gain a failure condition where it returns nullopt if the resulting string would not be a valid UTF-8 encoded string (which can never happen with our current restriction of disallowing all code points above U+0080, but the entire API is designed to be forward-compatible in that way).

We also want a `substr` function for `StaticString` eventually, but that will require the introduction of a code point index, which we currently don't have anywhere on the interface, and is a headache we probably don't want to get into at this point. 😅

iceoryx2-bb/cxx/include/iox2/container/static_string.hpp
```diff
@@ -477,18 +539,132 @@ class StaticString {
...
+    }
+
+    /// Finds the first occurence of a character equal to one of the characters of the given character sequence
+    /// and returns its position.
+    ///
+    /// @return an Optional containing the position of the first character equal to one of the characters of the given
+    ///         character sequence
+    ///         nullopt if no character is found or if pos is greater than size()
+    template <typename T>
+    auto find_first_of(T const& str, SizeType pos = 0U) const -> IsStaticStringOrCharArray<T, Optional<SizeType>> {
```
ComicSansMS: This belongs inside `ConstAccessorCodeUnits`.

iceoryx2-bb/cxx/include/iox2/container/static_string.hpp
```diff
@@ -477,18 +539,132 @@ class StaticString {
...
+    }
+
+    /// Finds the last occurence of a character equal to one of the characters of the given character sequence
+    /// and returns its position.
+    ///
+    /// @return an Optional containing the position of the last character equal to one of the characters of the given
+    ///         character sequence
+    ///         nullopt if no character is found
+    template <typename T>
+    auto find_last_of(T const& str, SizeType pos = N) const -> IsStaticStringOrCharArray<T, Optional<SizeType>> {
```
ComicSansMS: Also `ConstAccessorCodeUnits`.

iceoryx2-bb/cxx/include/iox2/container/static_string.hpp
```diff
@@ -477,18 +539,132 @@ class StaticString {
...
+        return nullopt;
+    }
+
+    /// Inserts a StaticString, obtained by str.substr(s_index, count), into the StaticString at position index. The
+    /// insertion fails if the capacity would be exceeded or the provided indices are larger than the respective
+    /// string lengths.
+    ///
+    /// @return true if the insertion was successful, otherwise false
+    template <typename T>
+    auto insert(SizeType index, T const& str, SizeType s_index, SizeType count = T::capacity()) ->
```
ComicSansMS: This can't stay here, for the same reasons that `substr` is problematic. It should probably move to `UncheckedAccessorCodeUnits`, although it is a different beast than the other accessors that we currently have. 

I am somewhat conflicted to what extent we do want to offer convenience functions for code unit-based string manipulations. If this function is useful now, I am not opposed to adding it, but before we start adding a whole bunch of string functions for code units, we should probably just figure out the design for code point indices and have all those functions in the main class.

iceoryx2-bb/cxx/include/iox2/container/static_string.hpp
```diff
@@ -26,6 +28,19 @@
...
+
+template <typename>
+struct IsStaticString : std::false_type { };
+
+template <uint64_t N>
+struct IsStaticString<StaticString<N>> : std::true_type { };
+
+template <typename T, typename ReturnType>
+using IsStaticStringOrCharArray =
+    typename std::enable_if_t<IsStaticString<T>::value || legacy::is_char_array<T>::value, ReturnType>;
```
ComicSansMS, line 58: Unlike `IsStaticString` above, this does not actually answer the question whether `T` is a static string or char array, so I don't like the name too much. How about `RequireStaticStringOrCharArray`?

iceoryx2-bb/cxx/include/iox2/container/static_string.hpp
```diff
@@ -477,18 +539,132 @@ class StaticString {
...
+    auto compare(StaticString const& other) const -> int64_t {
+        auto const other_size = other.size();
+        auto const res = memcmp(&m_string[0], &other.m_string[0], std::min(m_size, other_size));
+        if (res == 0) {
+            if (m_size < other_size) {
+                return -1;
+            }
+            return ((m_size > other_size) ? 1 : 0);
+        }
+        return res;
```
ComicSansMS, line 677: A not-so-obvious property of the UTF-8 encoding is that this actually orders strings lexicographically _by code points_. This may or may not be worth pointing out in a comment, but I think it's pretty cool.

iceoryx2-bb/cxx/include/iox2/container/detail/string_internal.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+/// Struct to get a pointer to the char array of the string literal
+template <uint64_t N>
+// Provides uniform and safe access (in combination with GetSize) to string like constructs
+// NOLINTNEXTLINE(hicpp-avoid-c-arrays, cppcoreguidelines-avoid-c-arrays, modernize-avoid-c-arrays)
+struct GetData<char[N]> {
+    static auto call(const CharArray<N>& data) -> const char* {
+        return &data[0];
+    }
+};
```
ComicSansMS, line 78: Why did you put all of these into `struct`s with a `call()` function instead of just having them as free functions?

ComicSansMS (changes requested):

iceoryx2-ffi/c/src/api/attribute.rs
FerdinandSpitzschnueffler: I removed the return value because it's currently ignored everywhere it's used in our code base and examples. Should I undo that?

iceoryx2-bb/cxx/include/iox2/container/static_string.hpp
FerdinandSpitzschnueffler: I'll move the `substr` method but save the headache for later ;)

iceoryx2-bb/cxx/include/iox2/container/static_string.hpp
FerdinandSpitzschnueffler: I only added the functions we need for the `SemanticString` and the path and file verifier. I'll move them into the respective `*AccessorCodeUnits` classes and make them private so that future design decisions for the code points do not break the public API.

iceoryx2-bb/cxx/include/iox2/bb/detail/path_and_file_verifier.hpp
```diff
@@ -52,57 +52,57 @@ enum class RelativePathComponents : uint8_t {
...
+    static auto is_valid_path_to_file(const typename container::StaticString<StringCapacity>& name) noexcept -> bool;
+
+    /// @brief returns true if the provided name is a valid path, otherwise false
+    /// @param[in] name the string to verify
+    static auto is_valid_path_to_directory(const container::StaticString<StringCapacity>& name) noexcept -> bool;
+
+    /// @brief returns true if the provided name ends with a path separator, otherwise false
+    /// @param[in] name the string which may contain a path separator at the end
+    static auto does_end_with_path_separator(const container::StaticString<StringCapacity>& name) noexcept -> bool;
+};
```
ComicSansMS, line 97: This is a class template that only contains static functions. This disables template argument deduction for the capacity. You always have to call this like:

```
StaticString<42> str;
PathAndFileVerifier<42>::does_end_with_path_separator(str);
```

If you just make them free functions instead, the capacity will be deduced from the arguments:

```
StaticString<42> str;
does_end_with_path_separator(str);
```

If grouping the functions together is still desirable, you can replace the `class` with a `namespace`.

iceoryx2-bb/cxx/include/iox2/bb/detail/path_and_file_verifier.hpp
```diff
@@ -234,12 +235,13 @@ inline auto is_valid_path_to_directory(const legacy::string<StringCapacity>& nam
...
 template <uint64_t StringCapacity>
-inline auto does_end_with_path_separator(const legacy::string<StringCapacity>& name) noexcept -> bool {
+inline auto PathAndFileVerifier<StringCapacity>::does_end_with_path_separator(
+    const container::StaticString<StringCapacity>& name) noexcept -> bool {
     if (name.empty()) {
         return false;
     }
     // AXIVION Next Construct AutosarC++19_03-A3.9.1: Not used as an integer but as actual character
-    const char last_character { name[name.size() - 1U] };
+    const char last_character { name.unchecked_access()[name.size() - 1U] };
```
ComicSansMS: Arguably more readable: `*name.code_units().back_element()`.

iceoryx2-bb/cxx/include/iox2/bb/file_name.hpp
```diff
@@ -48,15 +49,14 @@ class FileName : public SemanticString<FileName,
...
 
 namespace detail {
-inline auto
-file_name_does_contain_invalid_characters(const legacy::string<platform::IOX2_MAX_FILENAME_LENGTH>& value) noexcept
-    -> bool {
+inline auto file_name_does_contain_invalid_characters(
+    const container::StaticString<platform::IOX2_MAX_FILENAME_LENGTH>& value) noexcept -> bool {
     const auto value_size = value.size();
 
     for (uint64_t i { 0 }; i < value_size; ++i) {
```
ComicSansMS: Since you don't need the index for anything, you could also loop `for (char const c : value.unchecked_access())` here, or use `std::find_if` over `value.unchecked_access().begin()/.end()`.

iceoryx2-bb/cxx/include/iox2/bb/file_path.hpp
```diff
@@ -50,14 +51,14 @@ class FilePath : public SemanticString<FilePath,
 
 namespace detail {
 inline auto
-file_path_does_contain_invalid_characters(const legacy::string<platform::IOX2_MAX_PATH_LENGTH>& value) noexcept
+file_path_does_contain_invalid_characters(const container::StaticString<platform::IOX2_MAX_PATH_LENGTH>& value) noexcept
     -> bool {
     const auto value_size = value.size();
 
     for (uint64_t i { 0 }; i < value_size; ++i) {
```
ComicSansMS: Similar to above, this could be a `std::find_if` or a range-based `for`.

iceoryx2-bb/cxx/include/iox2/bb/semantic_string.hpp
```diff
@@ -29,10 +29,10 @@ enum class SemanticStringError : uint8_t {
 };
 
 template <uint64_t Capacity>
-using DoesContainInvalidCharacter = bool (*)(const legacy::string<Capacity>& value);
+using DoesContainInvalidCharacter = bool (*)(const typename container::StaticString<Capacity>& value);
 
 template <uint64_t Capacity>
-using DoesContainInvalidContent = bool (*)(const legacy::string<Capacity>& value);
+using DoesContainInvalidContent = bool (*)(const typename container::StaticString<Capacity>& value);
```
ComicSansMS: I don't think you need the `typename` in these two cases.

iceoryx2-bb/cxx/include/iox2/container/static_string.hpp
```diff
@@ -24,8 +26,36 @@
 #include <type_traits>
 
 namespace iox2 {
+namespace bb {
+template <uint64_t Capacity>
+using DoesContainInvalidCharacter = bool (*)(const typename container::StaticString<Capacity>& value);
+
+template <uint64_t Capacity>
+using DoesContainInvalidContent = bool (*)(const typename container::StaticString<Capacity>& value);
```
ComicSansMS: Also in these two cases: I think you can get rid of the `typename`.

iceoryx2-bb/cxx/include/iox2/container/static_string.hpp
```diff
@@ -154,6 +189,40 @@ class StaticString {
...
+                return false;
+            }
+
+            if (index > m_parent->m_size) {
+                return false;
+            }
+            auto number_of_characters_to_move = m_parent->m_size - index;
+            std::memmove(
+                &m_parent->m_string[index + sub_str_size], &m_parent->m_string[index], number_of_characters_to_move);
+            std::memcpy(&m_parent->m_string[index], sub_str->m_string, static_cast<size_t>(sub_str_size));
```
ComicSansMS: Instead of `memmove` and `memcpy`, consider using `std::copy_backward` and `std::copy` here and in `substr()`. `memcpy` has special properties that we don't rely on here and `copy` should give us at least as good codegen on an optimizing compiler.

iceoryx2-bb/cxx/include/iox2/bb/detail/path_and_file_verifier.hpp
FerdinandSpitzschnueffler: I moved the free functions into the class to make that class a friend of `UncheckedAccessorCodeUnits`. I wasn't able to make only the functions friend that need to be, except by moving their declarations into the `StaticString` file and I wanted to avoid this. Suggestions are very welcome...

iceoryx2-bb/cxx/include/iox2/container/detail/string_internal.hpp
FerdinandSpitzschnueffler: I don't have a strong opinion on this, I just tend to use the specialization approach here instead of function template overloading. Would there be any advantages of using free functions?

iceoryx2-cxx/tests/src/log_tests.cpp
```diff
@@ -73,8 +75,8 @@ TEST(Log, custom_logger_works) {
     log(LogLevel::Trace, "hello", "world");
     log(LogLevel::Debug, "goodbye", "hypnotoad");
     log(LogLevel::Info, "Who is looking for freedom?", "The Hoff!");
-    log(LogLevel::Warn, "Blümchen", "Bassface");
-    log(LogLevel::Error, "Blümchen should record a single with", "The almighty Hypnotoad");
+    log(LogLevel::Warn, "Bluemchen", "Bassface");
+    log(LogLevel::Error, "Bluemchen should record a single with", "The almighty Hypnotoad");
```
elBoberido, line 79: Out of curiosity, is currently only ASCII supported?

iceoryx2-bb/cxx/tests/src/semantic_string_tests.cpp
```diff
@@ -87,7 +89,6 @@ const std::vector<std::string> TestValues<FilePath>::VALID_VALUES { { "file" },
                                                                     { "123.456" },
                                                                     { ".hidden_me" },
                                                                     { "/some/file/path" },
-                                                                    { "./relative/path" },
```
elBoberido, line 90: Out of curiosity, why was this removed=

iceoryx2-ffi/c/src/api/attribute.rs
elBoberido: No strong feelings about this, but others might have a use case. I'll leave it to you.

elBoberido (approved):
Looks good to me. Since this PR touches quite some files and there are also others PRs which would result in merge conflicts, I would be fine with merging this and fixing the small remaining issues in a follow up PR.

I'll leave it to you to decide.

iceoryx2-bb/cxx/include/iox2/container/detail/string_internal.hpp
ComicSansMS: Conventional wisdom shows that specialization is almost never the right solution for function templates. A big part of the reason is that even with specialization, you still get overloading and they don't play nicely together. Between the two, overloading has the overall best ergonomics, so it is usually preferable. The typical case for wrapping them in a class type like this is when you want to opt-out of argument-dependent lookup (the C++20 ranges library does this).

But here you are wrapping the functions in class types because specialization with free functions would break your use case. The best answer to that is to just ditch specialization and go with overloading. It will remove noise from the implementation and improve usability for the callers because they get template argument deduction for the template parameters.

iceoryx2-bb/cxx/include/iox2/bb/detail/path_and_file_verifier.hpp
ComicSansMS: The long-distance friendship here is not nice. I appreciate the effort of trying to keep `substr()` and `insert()` out of the public interface, but the flipside of that is that we now have to deal with the beautiful world of friendship. 😅

I suspect the reason you could not make this work is that a `friend` declaration is always just that, a declaration. That means, if the compiler does not already know about the `does_end_with_separator` function template, the `friend` declaration will declare that function in the context of the class template. When it then encounters the free function definition later, it will be unable to match that to the earlier declaration. The only way to clear up that confusion is to give friendship to all instantiations of the function template, which is too broad, because we only need those instantiations with a matching capacity.

The somewhat unintuitive solution to this is to forward declare the function template before the class template definition. So what you end up with is something like this:

```cpp
// forward-declare the class first, because it is used in the function signature
template<uint64_t Cap>
class StaticString;

// forward-declare the function template
template<uint64_t Cap>
auto does_end_with_path_separator(StaticString<Cap> const& s) -> bool; 

template<uint64_t Cap>
class StaticString {
    // befriend the specific instance of the function template with the matching capacity
    friend auto does_end_with_path_separator<Cap>(StaticString<Cap> const&) -> bool;
};

template<uint64_t Cap>
auto does_end_with_path_separator(StaticString<Cap> const& s) -> bool {
   // this part can now access the private parts of StaticString
    ...
}
```
Does this help?

iceoryx2-cxx/tests/src/log_tests.cpp
FerdinandSpitzschnueffler: Yes, only the code points from 1 to 128 are currently supported.

ComicSansMS: I guess we could just make the `Entry` class used by the test logger not use `StaticString` for storing its contents? Or does that class need to be sent over an iceoryx2 communication channel at some point?

iceoryx2-bb/cxx/tests/src/semantic_string_tests.cpp
FerdinandSpitzschnueffler: The tests that used this string created file paths like `f./relative/pathile` which is invalid on Windows (because of the `f.`).

FerdinandSpitzschnueffler:
The open points will be addressed in a follow-up PR: #1258

iceoryx2-bb/cxx/include/iox2/bb/detail/path_and_file_verifier.hpp
FerdinandSpitzschnueffler: @ComicSansMS Addressed in https://github.com/eclipse-iceoryx/iceoryx2/pull/1258

iceoryx2-bb/cxx/include/iox2/container/detail/string_internal.hpp
FerdinandSpitzschnueffler: @ComicSansMS Addressed in https://github.com/eclipse-iceoryx/iceoryx2/pull/1258

PR #1286, opened 2026-01-06
[#1285] allow dyn types in waitset

### LINKED ISSUE #1285: Add `?Sized` to waitset attachment type requirements to allow dyn Types

Currently, the attachments of the waitset must all have a compile time fixed size which prevents the usage of `dyn` types in the user code. Add `?Sized` to allow `dyn` types.

elfenpiff:

* [x] Assign PR to reviewer

## PR Reviewer Reminders

* Public API is documented

## References

Closes #1285

### DIFF
--- a/doc/release-notes/iceoryx2-unreleased.md
+++ b/doc/release-notes/iceoryx2-unreleased.md
@@ -11,7 +11,8 @@
     conflicts when merging.
 -->
 
-* Example text [#1](https://github.com/eclipse-iceoryx/iceoryx2/issues/1)
+* Allow `dyn` types as `WaitSet` attachments
+  [#1285](https://github.com/eclipse-iceoryx/iceoryx2/issues/1285)
 
 ### Bugfixes
 


dkroenke:
What is the use case to allow dynamic types for Waitset attachments? In the example codes we pass the attachment (listener etc.) mostly as references which should be itself fixed-size. But maybe there is another reason.

elfenpiff:
@dkroenke 

> What is the use case to allow dynamic types for Waitset attachments? In the example codes we pass the attachment (listener etc.) mostly as references which should be itself fixed-size. But maybe there is another reason.

Let's assume the user has its own trait which also requires SynchronousMultiplexing.

```rust
trait MyTrait: SynchronousMultiplexing {}
```

Now you want to implement a function that attaches a type that implements `MyTrait` to the `WaitSet`. This works since it requires `SynchronousMultiplexing`.

We have two options. The generic/template variant:

```rust
fn attach_to_waitset<T: MyTrait>(some_object: &T) {}
```

which generates a new function for every type and increases the compile time and binary size.

Another option is

```rust
fn attach_to_waitset(some_object: &dyn MyTrait) {}
```

In this case, the compiler must be able to build a vtable and this only works when the trait has not a compile time fixed size but is `?Sized`.

The second option was not possible with our waitset. Here, the use case is that the user has a list of objects that implement `MyTrait`, then they have no other option than to go for the `dyn MyTrait` approach since they do not know the specific type.

Assume, `MyTrait` defines GUI widgets and the user has a list of interactive gui widgets which shall be attached to a waitset so that a background thread reacts whenever the user clicked something. Then we have a list of `Vec<Box<dyn MyTrait>>` and need to attach it to a waitset - which was not possible.

PR #1313, opened 2026-01-22
[#1312] no zero copy send for option result

### LINKED ISSUE #1312: Remove `ZeroCopySend` for `Optional`, `Result` and `Duration`
The Rust types `Optional`, `Result` and `Duration` do not have a stable ABI and a fixed memory layout, therefore they cannot implement `ZeroCopySend`. Remove the `ZeroCopySend` implementation for those types.

elfenpiff:
## Notes for Reviewer

* Introduces an API compatible (to `Option`) `ZeroCopySend` capable `StaticOption`
* remove `Option` and `Duration` from `ZeroCopySend` trait default impl
* Introduces a `StaticDuration` that implements `ZeroCopySend`.

* [x] Assign PR to reviewer

## PR Reviewer Reminders

* Public API is documented

## References

Closes #1312

### DIFF
--- a/doc/release-notes/iceoryx2-unreleased.md
+++ b/doc/release-notes/iceoryx2-unreleased.md
@@ -18,7 +18,7 @@
   [#1289](https://github.com/eclipse-iceoryx/iceoryx2/issues/1289)
 * Add source `NodeId` to request and response header
   [#1308](https://github.com/eclipse-iceoryx/iceoryx2/issues/1308)
-* Introduce `RelocatableOption` and `RelocatableDuration` which are 
+* Introduce `RelocatableOption` and `RelocatableDuration` which are
   `ZeroCopySend`
   [#1312](https://github.com/eclipse-iceoryx/iceoryx2/issues/1312)
 
--- a/iceoryx2-bb/container/tests/relocatable_option_tests.rs
+++ b/iceoryx2-bb/container/tests/relocatable_option_tests.rs
@@ -31,7 +31,7 @@ fn default_created_option_is_empty() {
 fn when_empty_as_option_returns_empty_option() {
     let sut = RelocatableOption::<i32>::default();
 
-    assert_that!(sut.as_option_ref(), eq None);
+    assert_that!(sut.as_option_ref(), eq Option::<&i32>::None);
 }
 
 #[test]
@@ -45,7 +45,7 @@ fn when_with_value_as_option_returns_option_with_reference_to_that_value() {
 fn when_empty_as_option_mut_returns_empty_option() {
     let mut sut = RelocatableOption::<i32>::default();
 
-    assert_that!(sut.as_option_mut(), eq None);
+    assert_that!(sut.as_option_mut(), eq Option::<&mut i32>::None);
 }
 
 #[test]
@@ -59,7 +59,7 @@ fn when_with_value_as_option_returns_option_with_mut_reference_to_that_value() {
 fn when_empty_as_deref_returns_empty_option() {
     let sut = RelocatableOption::<Vec<i32>>::default();
 
-    assert_that!(sut.as_deref().is_none(), eq true);
+    assert_that!(sut.as_deref(), eq RelocatableOption::<&[i32]>::None);
 }
 
 #[test]
@@ -73,7 +73,7 @@ fn when_with_value_as_deref_returns_ref_to_target() {
 fn when_empty_as_deref_mut_returns_empty_option() {
     let mut sut = RelocatableOption::<Vec<i32>>::default();
 
-    assert_that!(sut.as_deref_mut().is_none(), eq true);
+    assert_that!(sut.as_deref_mut(), eq RelocatableOption::<&mut [i32]>::None);
 }
 
 #[test]
@@ -87,7 +87,7 @@ fn when_with_value_as_deref_mut_returns_ref_to_target() {
 fn when_empty_as_ref_returns_empty_option() {
     let sut = RelocatableOption::<i32>::default();
 
-    assert_that!(sut.as_ref().is_none(), eq true);
+    assert_that!(sut.as_ref(), eq RelocatableOption::<&i32>::None);
 }
 
 #[test]
@@ -101,7 +101,7 @@ fn when_with_value_as_ref_returns_ref_to_target() {
 fn when_empty_as_mut_returns_empty_option() {
     let mut sut = RelocatableOption::<i32>::default();
 
-    assert_that!(sut.as_mut().is_none(), eq true);
+    assert_that!(sut.as_mut(), eq RelocatableOption::<&mut i32>::None);
 }
 
 #[test]
@@ -147,7 +147,7 @@ fn some_creates_option_that_contains_the_value() {
 fn inspect_callback_is_not_called_when_empty() {
     let sut = RelocatableOption::<i32>::None;
     let mut callback_was_called = false;
-    sut.inspect(|_| callback_was_called = true);
+    sut.inspect(|_: &i32| callback_was_called = true);
 
     assert_that!(callback_was_called, eq false);
 }
@@ -156,7 +156,7 @@ fn inspect_callback_is_not_called_when_empty() {
 fn inspect_callback_is_called_when_it_contains_a_value() {
     let sut = RelocatableOption::<i32>::Some(778);
     let mut callback_was_called = false;
-    sut.inspect(|v| {
+    sut.inspect(|v: &i32| {
         callback_was_called = true;
         assert_that!(*v, eq 778);
     });
@@ -169,7 +169,7 @@ fn map_of_empty_option_is_empty() {
     let sut = RelocatableOption::<i32>::None;
 
     let mut callback_was_called = false;
-    assert_that!(sut.map(|v| {
+    assert_that!(sut.map(|v: i32| {
         callback_was_called = true;
         v + 1}).is_none(), eq true);
     assert_that!(callback_was_called, eq false);
@@ -179,7 +179,7 @@ fn map_of_empty_option_is_empty() {
 fn map_uses_value_and_creates_new_option() {
     let sut = RelocatableOption::<i32>::Some(5);
 
-    assert_that!(sut.map(|v| v + 1), eq RelocatableOption::Some(6));
+    assert_that!(sut.map(|v: i32| v + 1), eq RelocatableOption::Some(6));
 }
 
 #[test]
@@ -214,7 +214,7 @@ fn take_empties_sut_and_returns_content() {
 fn take_if_does_not_call_callback_when_empty() {
     let mut sut = RelocatableOption::<i32>::None;
     let mut callback_was_called = false;
-    let ret_val = sut.take_if(|_| {
+    let ret_val = sut.take_if(|_: &mut i32| {
         callback_was_called = true;
         false
     });
@@ -228,7 +228,7 @@ fn take_if_does_not_call_callback_when_empty() {
 fn take_if_returns_none_when_callback_returns_false() {
     let mut sut = RelocatableOption::<i32>::Some(551);
     let mut callback_was_called = false;
-    let ret_val = sut.take_if(|v| {
+    let ret_val = sut.take_if(|v: &mut i32| {
         assert_that!(*v, eq 551);
         callback_was_called = true;
         false
@@ -374,7 +374,7 @@ fn value_creates_native_option_with_value() {
 }
 
 #[test]
-fn native_to_static_conversion_works() {
+fn native_to_relocatable_conversion_works() {
     let native = Option::<i32>::Some(551);
     let sut: RelocatableOption<i32> = native.into();
 
@@ -382,7 +382,7 @@ fn native_to_static_conversion_works() {
 }
 
 #[test]
-fn static_to_native_conversion_works() {
+fn relocatable_to_native_conversion_works() {
     let sut = RelocatableOption::<i32>::Some(1823);
     let native: Option<i32> = sut.into();
 


iceoryx2-bb/posix/src/clock.rs
```diff
@@ -185,6 +185,32 @@ impl TimeBuilder {
...
+/// A intermediate [`Duration`] representation with a fixed memory layout which
+/// is shared memory compatible.
+#[repr(C)]
+#[derive(
+    Default, Clone, Copy, Eq, PartialEq, Hash, Debug, ZeroCopySend, Serialize, Deserialize,
+)]
+pub struct StaticDuration {
+    seconds: u64,
+    nanoseconds: u32,
+}
```
elBoberido: I would suggest not to use the prefix static for these kind of types since it is already used with other meanings in Rust, e.g. static variables and static lifetime. Furthermore, what is a non-static duration?

I am thinking about an alternative, but cannot come up with something short and convincing at the same time. There is an option to use `CDuration` or `FfiOption` due to the repr(C) or `AbiDuration` due to the ABI guarantees. Maybe we need a custom abbreviation to prevent getting a wrong notion of what the type is meant to be. Having a namespace instead of a prefix might also be worth to think of, similar to Rusts `std::ffi` module.

Also, would it make sense to move this type out of posix? It can also easily be used on bare-metal. Is there an elementary/time?

Do we want (or guarantee) this to be compatible with the C++ Duration?

iceoryx2-bb/posix/src/clock.rs
```diff
@@ -185,6 +185,32 @@ impl TimeBuilder {
     }
 }
 
+/// A intermediate [`Duration`] representation with a fixed memory layout which
+/// is shared memory compatible.
+#[repr(C)]
+#[derive(
+    Default, Clone, Copy, Eq, PartialEq, Hash, Debug, ZeroCopySend, Serialize, Deserialize,
```
elBoberido, line 192: Are you sure that `Eq` and `PartialEq` are doing the right thing here?

iceoryx2-bb/posix/tests/clock_tests.rs
```diff
@@ -85,3 +85,14 @@ fn clock_time_as_timespec_works() {
...
+
+#[test]
+fn clock_static_duration_roundtrip_conversion() {
+    let secs = 123;
+    let nsecs = 456;
+    let duration = Duration::from_secs(secs) + Duration::from_nanos(nsecs);
+    let sut: StaticDuration = duration.into();
+    let duration_2: Duration = sut.into();
+
+    assert_that!(duration, eq duration_2);
```
elBoberido, line 97: Please also assert that the `StaticDuration::secs` actually contains the seconds and not the nanoseconds and vice versa, e.g. by using `Duration::MAX`, which would also ensure that a conversion between `Duration` and `StaticDuration` is lossless.

iceoryx2-bb/container/src/relocatable_option.rs
```diff
@@ -0,0 +1,465 @@
...
+/// match do_stuff().as_option_ref() {
+///   Some(v) => println!("{v}"),
+///   None => println!("none")
+/// }
+/// ```
+#[repr(C)]
+pub struct StaticOption<T> {
+    data: MaybeUninit<T>,
+    has_contents: u8,
+}
```
elBoberido: If I did not misunderstood [RFC-2195](https://github.com/rust-lang/rfcs/blob/master/text/2195-really-tagged-unions.md), the whole implementation can be simplified to

```rust
#[repr(c)] // or repr(Int) if the width of the tag shall also be defined
pub struct StaticOption<T> {
    Some(T),
    None,
}
```

With this, one could then also `impl<T: Copy> Copy for StaticOption<T> {}` since `Drop` does not need to be implemented manually.

Similar to `StaticDuration`, I would suggest not to use the `static` prefix for these kind of types.

iceoryx2-bb/posix/src/clock.rs
orecham: `RelocatableDuration` ?

iceoryx2-bb/posix/src/clock.rs
elfenpiff: Yes, what concern do you have?

doc/release-notes/iceoryx2-unreleased.md
```diff
@@ -18,6 +18,8 @@
   [#1289](https://github.com/eclipse-iceoryx/iceoryx2/issues/1289)
 * Add source `NodeId` to request and response header
   [#1308](https://github.com/eclipse-iceoryx/iceoryx2/issues/1308)
+* Introduce `StaticOption` and `StaticDuration` which are `ZeroCopySend`
```
elBoberido:
```suggestion
* Introduce `RelocatableOption` and `RelocatableDuration` which are `ZeroCopySend`
```

iceoryx2-bb/container/src/relocatable_option.rs
```diff
@@ -0,0 +1,362 @@
...
+};
+use iceoryx2_log::fatal_panic;
+use serde::{de::Visitor, Deserialize, Serialize};
+
+/// Implementation of an [`Option`] that is shared-memory compatible,
+/// has a stable memory layout and can be used for zero-copy cross-language
+/// communication.
+///
+/// The usage is as close as possible to the original [`Option`], except for
+/// the construction via [`Some`] and [`None`].
```
elBoberido:
```suggestion
/// The usage is as close as possible to the original [`Option`], except for
/// the construction via [`Some`] and [`None`], which needs to be
/// `RelocatableOption::Some` and `RelocatableOption::None`.
```

iceoryx2-bb/container/src/relocatable_option.rs
```diff
@@ -0,0 +1,362 @@
...
+        match self {
+            Self::Some(v) => Some(v),
+            Self::None => None,
+        }
+    }
+
+    /// Returns an [`Option`] with a reference to `T`
+    pub fn as_option_ref(&self) -> Option<&T> {
+        match self {
+            Self::Some(ref v) => Some(v),
```
elBoberido: `ref` is superfluous due to `&self`

iceoryx2-bb/container/src/relocatable_option.rs
```diff
@@ -0,0 +1,362 @@
...
+            Self::None => None,
+        }
+    }
+
+    /// Returns an [`Option`] with a mutable reference to `T`
+    pub fn as_option_mut(&mut self) -> Option<&mut T> {
+        match self {
+            Self::Some(ref mut v) => Some(v),
```
elBoberido: Same here. `v` is already a `&mut T`

iceoryx2-bb/container/src/relocatable_option.rs
```diff
@@ -0,0 +1,362 @@
...
+    pub fn expect(self, msg: &str) -> T {
+        if self.is_none() {
+            let origin = alloc::format!(
+                "RelocatableOption::<{}>::expect()",
+                core::any::type_name::<T>()
+            );
+            fatal_panic!(from origin, "Expect: {msg}");
+        }
+
+        unsafe { self.unwrap_unchecked() }
```
elBoberido: No unsafe necessary
```suggestion
        if let Self::Some(v) = self {
            v
        } else {
            let origin =
                alloc::format!("RelocatableOption::<{}>::expect()", core::any::type_name::<T>());
            fatal_panic!(from origin, "Expect: {msg}");
        }
```

iceoryx2-bb/container/src/relocatable_option.rs
```diff
@@ -0,0 +1,362 @@
...
+        self
+    }
+
+    /// Returns [`true`] if the [`RelocatableOption`] does not contain a value, other
+    /// it returns [`false`].
+    pub fn is_none(&self) -> bool {
+        match self {
+            RelocatableOption::None => true,
+            RelocatableOption::Some(_v) => false,
+        }
```
elBoberido, line 255: Food for thought: `if let Self::None = self ...` would be a little bit less verbose

iceoryx2-bb/container/src/relocatable_option.rs
```diff
@@ -0,0 +1,362 @@
...
+        !self.is_none()
+    }
+
+    /// Maps a `RelocatableOption<T>` to a `RelocatableOption<U>`
+    pub fn map<U, F: FnOnce(T) -> U>(self, f: F) -> RelocatableOption<U> {
+        match self {
+            RelocatableOption::None => RelocatableOption::None,
+            RelocatableOption::Some(v) => RelocatableOption::Some(f(v)),
+        }
+    }
```
elBoberido, line 270: Similar here. But `Self::None => Relocate...` could be used to make it a little bit more verbose ... not necessary, though.

iceoryx2-bb/container/src/relocatable_option.rs
```diff
@@ -0,0 +1,362 @@
...
+
+    /// Replaces the existing value of the [`RelocatableOption`] with the new value.
+    /// The old value is returned.
+    pub fn replace(&mut self, value: T) -> RelocatableOption<T> {
+        let mut new_value = RelocatableOption::Some(value);
+        core::mem::swap(self, &mut new_value);
+        new_value
```
elBoberido: Why not just 
```suggestion
        core::mem::replace(self, Self::Some(value))
```

iceoryx2-bb/container/src/relocatable_option.rs
```diff
@@ -0,0 +1,362 @@
...
+        match self {
+            RelocatableOption::None => RelocatableOption::None,
+            RelocatableOption::Some(v) => {
+                if predicate(v) {
+                    core::mem::take(self)
+                } else {
+                    RelocatableOption::None
+                }
+            }
+        }
```
elBoberido, line 296: This is simpler and does the same
```suggestion
        match self {
            Self::Some(ref v) if predicate(v) => self.take(),
            _ => Self::None,
        }
```

iceoryx2-bb/container/src/relocatable_option.rs
```diff
@@ -0,0 +1,362 @@
...
+                "RelocatableOption::<{}>::unwrap()",
+                core::any::type_name::<T>()
+            );
+            fatal_panic!(
+                from origin,
+                "This should never happen! Accessing the value of an empty RelocatableOption."
+            );
+        }
+
+        unsafe { self.unwrap_unchecked() }
```
elBoberido: This can be done without unsafe
```suggestion
        if let Self::Some(v) = self {
            v
        } else {
            let origin =
                alloc::format!("StaticOption::<{}>::unwrap()", core::any::type_name::<T>());
            fatal_panic!(
                from origin,
                "This should never happen! Accessing the value of an empty RelocatableOption."
            );
        }
```

iceoryx2-bb/container/src/relocatable_option.rs
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    /// Consumes the [`RelocatableOption`] and either returns the contained value,
+    /// if there is one, otherwise `default` is returned.
+    pub fn unwrap_or(self, default: T) -> T {
+        if self.is_none() {
+            default
+        } else {
+            unsafe { self.unwrap_unchecked() }
+        }
```
elBoberido: This can also be done without unsafe
```suggestion
        self.unwrap_or_else(|| default)
```

Also, while the Rust Option also uses `default` as variable name, it is a bit overloaded. Food for thought. How about using `alternative`?

iceoryx2-bb/container/src/relocatable_option.rs
```diff
@@ -0,0 +1,362 @@
...
+    /// if there is one, otherwise `T::default()` is returned.
+    pub fn unwrap_or_default(self) -> T
+    where
+        T: Default,
+    {
+        if self.is_none() {
+            T::default()
+        } else {
+            unsafe { self.unwrap_unchecked() }
+        }
```
elBoberido: This can also be done without unsafe
```suggestion
        self.unwrap_or(T::default())
```

iceoryx2-bb/container/src/relocatable_option.rs
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    /// Consumes the [`RelocatableOption`] and either returns the contained value,
+    /// if there is one or returns the return value of the provided callback.
+    pub fn unwrap_or_else<F: FnOnce() -> T>(self, f: F) -> T {
+        if self.is_none() {
+            f()
+        } else {
+            unsafe { self.unwrap_unchecked() }
+        }
```
elBoberido: This can also be done without unsafe
```suggestion
        if let Self::Some(v) = self {
            v
        } else {
            f()
        }
```

iceoryx2-bb/posix/src/clock.rs
elBoberido: Aaargh, I mistook it for `Ord` and `PartialOrd`

iceoryx2-bb/posix/tests/clock_tests.rs
```diff
@@ -85,3 +85,16 @@ fn clock_time_as_timespec_works() {
...
+fn clock_relocatable_duration_roundtrip_conversion() {
+    let secs = 123;
+    let nsecs = 456;
+    let duration = Duration::from_secs(secs) + Duration::from_nanos(nsecs);
+    let sut: RelocatableDuration = duration.into();
+    let duration_2: Duration = sut.into();
+
+    assert_that!(duration, eq duration_2);
+    assert_that!(sut.as_secs(), eq duration.as_secs());
+    assert_that!(sut.subsec_nanos(), eq duration.subsec_nanos());
```
elBoberido, line 99: Better, but without checking for `Duration::MAX` it is still possible to have a wrong (or incompatible) implementation, e.g. the seconds could be stored in a 32 bit value

iceoryx2/src/service/builder/event.rs
```diff
@@ -481,7 +482,13 @@ impl<ServiceType: service::Service> Builder<ServiceType> {
                     .base
                     .create_node_service_tag(msg, EventCreateError::InternalFailure)?;
 
-                if let Some(ref mut deadline) = self.base.service_config.event_mut().deadline {
+                if let Some(ref mut deadline) = self
+                    .base
+                    .service_config
+                    .event_mut()
+                    .deadline
+                    .as_option_mut()
```
elBoberido: Small micro optimization
```suggestion
                if let Relocatable::Some(ref mut deadline) = self
                    .base
                    .service_config
                    .event_mut()
                    .deadline
```

iceoryx2/src/service/builder/event.rs
```diff
@@ -662,14 +669,14 @@ impl<ServiceType: service::Service> Builder<ServiceType> {
         }
 
         if self.verify_deadline
-            && existing_settings.deadline.map(|v| v.value)
-                != required_settings.deadline.map(|v| v.value)
+            && existing_settings.deadline.as_option_ref().map(|v| v.value)
+                != required_settings.deadline.as_option_ref().map(|v| v.value)
```
elBoberido: Another micro optimization
```suggestion
            && existing_settings.deadline.map(|v| v.value)
                != required_settings.deadline.map(|v| v.value)
```

iceoryx2/src/service/builder/event.rs
```diff
@@ -662,14 +669,14 @@ impl<ServiceType: service::Service> Builder<ServiceType> {
...
         {
             fail!(from self, with EventOpenError::IncompatibleDeadline,
                 "{} since the deadline is {:?} but a deadline of {:?} is required.",
                 msg, existing_settings.deadline, required_settings.deadline);
         }
 
-        Ok(*existing_settings)
+        Ok(existing_settings.clone())
```
elBoberido: In theory, the compiler should optimize the `clone` into a plain copy but this change could also be reverted.

iceoryx2/src/service/static_config/event.rs
```diff
@@ -86,7 +92,7 @@ impl StaticConfig {
     /// to a [`WaitSet`](crate::waitset::WaitSet) are woken up and notified about the missed
     /// deadline.
     pub fn deadline(&self) -> Option<Duration> {
-        self.deadline.map(|v| v.value)
+        self.deadline.as_option_ref().map(|v| v.value.into())
```
elBoberido, line 95: This can also be reverted or if you want to keep the ref, refactored to
```suggestion
        self.deadline.map(|ref v| v.value.into())
```

Same with the other three code snippets below

iceoryx2-bb/container/src/relocatable_option.rs
elfenpiff: I like this more, it is more consistent with the rest of the code.

iceoryx2-bb/container/src/relocatable_option.rs
elfenpiff: Doesn't work since v cannot be borrowed as mutable even with `ref mut v` and `predicate` requires a mut ref.

iceoryx2/src/service/static_config/event.rs
elfenpiff: does not work,  this would return a RelocatableOption but we need an Option

elBoberido: Right. Another `into` would be required

iceoryx2-bb/container/src/relocatable_option.rs
elBoberido: It's weird that the predicate needs a `&mut T`. Wonder what the intent for this is in the Rust Option?

iceoryx2-bb/container/tests/relocatable_option_tests.rs
```diff
@@ -0,0 +1,390 @@
...
+
+    assert_that!(sut.is_some(), eq false);
+    assert_that!(sut.is_none(), eq true);
+}
+
+#[test]
+fn when_empty_as_option_returns_empty_option() {
+    let sut = RelocatableOption::<i32>::default();
+
+    assert_that!(sut.as_option_ref(), eq None);
```
elBoberido: If you really want to check if you get an Option with a reference to the value, the current test is not sufficient. You would need to do a pattern matching in a similar way to this
```suggestion
    assert_that!(sut.as_option_ref(), eq Option::<&i32>::None);
```
Same with the other tests.

iceoryx2-bb/container/tests/relocatable_option_tests.rs
```diff
@@ -0,0 +1,390 @@
...
+}
+
+#[test]
+fn when_with_value_as_option_returns_option_with_reference_to_that_value() {
+    let sut = RelocatableOption::<i32>::Some(1234);
+
+    assert_that!(*sut.as_option_ref().unwrap(), eq 1234);
```
elBoberido, line 41:
```suggestion
    assert_that!(sut.as_option_ref(), eq Some(&1234_i32));
```

iceoryx2-bb/container/tests/relocatable_option_tests.rs
```diff
@@ -0,0 +1,390 @@
...
+    let mut sut = RelocatableOption::<i32>::default();
+
+    assert_that!(sut.as_option_mut(), eq None);
+}
+
+#[test]
+fn when_with_value_as_option_returns_option_with_mut_reference_to_that_value() {
+    let mut sut = RelocatableOption::<i32>::Some(4313);
+
+    assert_that!(*sut.as_option_mut().unwrap(), eq 4313);
```
elBoberido, line 55:
```suggestion
    assert_that!(sut.as_option_mut(), eq Some(&mut 4313_i32));
```

iceoryx2-bb/container/tests/relocatable_option_tests.rs
```diff
@@ -0,0 +1,390 @@
...
+    let sut = RelocatableOption::<Vec<i32>>::default();
+
+    assert_that!(sut.as_deref().is_none(), eq true);
+}
+
+#[test]
+fn when_with_value_as_deref_returns_ref_to_target() {
+    let sut = RelocatableOption::<Vec<i32>>::Some(vec![1, 2, 3]);
+
+    assert_that!(sut.as_deref().unwrap(), eq [1,2,3]);
```
elBoberido, line 69: With deref, it is especially important, else one cannot check if the deref of the unwrapped `Vec` would be used
```suggestion
    aassert_that!(sut.as_deref(), eq RelocatableOption::<&[i32]>::Some(&[1,2,3]));
```

iceoryx2-bb/container/tests/relocatable_option_tests.rs
```diff
@@ -0,0 +1,390 @@
...
+    assert_that!(sut.is_none(), eq false);
+    assert_that!(sut.is_some(), eq true);
+    assert_that!(sut.unwrap(), eq 89928);
+}
+
+#[test]
+fn inspect_callback_is_not_called_when_empty() {
+    let sut = RelocatableOption::<i32>::None;
+    let mut callback_was_called = false;
+    sut.inspect(|_| callback_was_called = true);
```
elBoberido:
```suggestion
    sut.inspect(|_: &i32| callback_was_called = true);
```

iceoryx2-bb/container/tests/relocatable_option_tests.rs
```diff
@@ -0,0 +1,390 @@
...
+
+    assert_that!(callback_was_called, eq false);
+}
+
+#[test]
+fn inspect_callback_is_called_when_it_contains_a_value() {
+    let sut = RelocatableOption::<i32>::Some(778);
+    let mut callback_was_called = false;
+    sut.inspect(|v| {
```
elBoberido:
```suggestion
    sut.inspect(|v: &i32| {
```

iceoryx2-bb/container/tests/relocatable_option_tests.rs
```diff
@@ -0,0 +1,390 @@
...
+
+    assert_that!(callback_was_called, eq true);
+}
+
+#[test]
+fn map_of_empty_option_is_empty() {
+    let sut = RelocatableOption::<i32>::None;
+
+    let mut callback_was_called = false;
+    assert_that!(sut.map(|v| {
```
elBoberido:
```suggestion
    assert_that!(sut.map(|v: i32| {
```

iceoryx2-bb/container/tests/relocatable_option_tests.rs
```diff
@@ -0,0 +1,390 @@
...
+
+    assert_that!(sut_full.is_none(), eq true);
+    assert_that!(sut_empty.is_none(), eq true);
+}
+
+#[test]
+fn take_if_does_not_call_callback_when_empty() {
+    let mut sut = RelocatableOption::<i32>::None;
+    let mut callback_was_called = false;
+    let ret_val = sut.take_if(|_| {
```
elBoberido:
```suggestion
    let ret_val = sut.take_if(|_: &i32| {
```

iceoryx2-bb/container/tests/relocatable_option_tests.rs
```diff
@@ -0,0 +1,390 @@
...
+
+#[test]
+fn value_creates_native_option_with_value() {
+    let sut = RelocatableOption::<i32>::Some(772);
+
+    assert_that!(sut.to_option(), eq Some(772));
+}
+
+#[test]
+fn native_to_static_conversion_works() {
```
elBoberido:
```suggestion
fn native_to_relocatable_conversion_works() {
```

iceoryx2-bb/container/tests/relocatable_option_tests.rs
```diff
@@ -0,0 +1,390 @@
...
+    let native = Option::<i32>::Some(551);
+    let sut: RelocatableOption<i32> = native.into();
+
+    assert_that!(sut.to_option(), eq native);
+}
+
+#[test]
+fn static_to_native_conversion_works() {
```
elBoberido:
```suggestion
fn relocatable_to_native_conversion_works() {
```

iceoryx2-bb/container/tests/relocatable_option_tests.rs
elfenpiff: This is less of a unit test but more of an API tests, so that the function name and the return type are correct. It would be sufficient to test this once for every function since the compiler and the language ensures that this does not change during runtime thanks to no function overload, the rust compiler etc.

So I add this once, for every `as_option*` function.

elBoberido: The idea was also to get rid of the unwraps. We have too many of them already 😅

elfenpiff: Why do you think that `unwrap()` is bad in a test?

elBoberido: In this specific case it's not that bad, because the error message will be that an empty Option is accessed. Not using unwrap would only lead to a slightly better error message. It would be just to train the muscle memory to not use unwrap.

The bigger issue is when the SUT ist not an Option but something else. For example, we have some flaky tests, which I wanted to debug but since there is an unwrap on a Result, the underlying error is hidden and it makes it hard to understand what is going on

PR #1328, opened 2026-01-28
[#1327] Use target switches for POSIX usage instead of feature

### LINKED ISSUE #1327: Remove posix feature from crates
## Brief Description

The `posix` feature defined in some crates is used to switch between implementations of structs that utilize `std` and equivalents that work directly with the POSIX platform abstraction.

This approach does not work so well when building tests for `no_std`. Furthermore, the additional feature switch adds some configuration complexity.

orecham:

* [ ] Assign PR to reviewer

## PR Reviewer Reminders

* Public API is documented

## References

Closes #1327

### DIFF
--- a/doc/release-notes/iceoryx2-unreleased.md
+++ b/doc/release-notes/iceoryx2-unreleased.md
@@ -42,6 +42,8 @@
     NOTE: Add new entries sorted by issue number to minimize the possibility of
     conflicts when merging.
 -->
+* Remove support for Bazel Workspaces
+  [#1263](https://github.com/eclipse-iceoryx/iceoryx2/issues/1263)
 * Adjust test names to naming convention
   [#1273](https://github.com/eclipse-iceoryx/iceoryx2/issues/1273)
 * Move character output abstraction into their own crate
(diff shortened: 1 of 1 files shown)


iceoryx2-bb/loggers/src/console.rs
```diff
@@ -47,7 +47,7 @@ fn duration_since_epoch() -> core::time::Duration {
             .duration_since(std::time::UNIX_EPOCH)
             .unwrap()
     }
-    #[cfg(feature = "posix")]
+    #[cfg(all(not(feature = "std"), any(target_os = "linux", target_os = "nto",)))]
```
orecham, line 50: Doing this switch in this crate is a temporary solution ... An abstraction that takes care of switching between std/posix/bare metal would be better IMO, but we need another layer between the posix abstraction and the upper layers to do this nicely. Left for another time.

iceoryx2-pal/print/src/writer.rs
```diff
@@ -120,35 +120,47 @@ impl Write for Stderr {
...
     fn is_terminal(&self) -> bool {
         true
     }
 }
 
-#[cfg(not(any(feature = "std", feature = "posix")))]
+#[cfg(all(
+    not(feature = "std"),
+    not(any(target_os = "linux", target_os = "nto",))
+))]
```
elBoberido, line 133: Food for thought. Could this be handled in a similar fashion to `platform_override`, which is set in the `build.rs` and then you could just use `not(posix)` instead of `not(any(target_os = "linux", target_os = "nto",))`? This could make it potentially also configurable for ouf-of-tree platforms

orecham: Worth considering. It would also keep the list of targets with POSIX in one place. I wonder if we would then need to add a build.rs to every crate though.

elBoberido: Yes, every crate which would use this `cfg` would need a `build.rs`. Unfortunately, it is not possible to use any constant expression in a `cfg` attribute.

PR #1352, opened 2026-02-13
Iox2 1351 package version api

ComicSansMS:

* [ ] Assign PR to reviewer

## PR Reviewer Reminders

* Public API is documented

## References

Closes #

### DIFF
--- a/iceoryx2-cxx/include/iox2/version.hpp
+++ b/iceoryx2-cxx/include/iox2/version.hpp
@@ -1,0 +1,40 @@
+// Copyright (c) 2026 Contributors to the Eclipse Foundation
+//
+// See the NOTICE file(s) distributed with this work for additional
+// information regarding copyright ownership.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Apache Software License 2.0 which is available at
+// https://www.apache.org/licenses/LICENSE-2.0, or the MIT license
+// which is available at https://opensource.org/licenses/MIT.
+//
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+#ifndef IOX2_VERSION_HPP
+#define IOX2_VERSION_HPP
+
+#include "iox2/internal/iceoryx2.hpp"
+
+#include <cstdint>
+#include <iosfwd>
+
+namespace iox2 {
+
+/// Version number.
+struct PackageVersion {
+    std::uint16_t major;
+    std::uint16_t minor;
+    std::uint16_t patch;
+};
+
+/// Returns the crates version acquired through the internal environment variables set by cargo,
+/// ("CARGO_PKG_VERSION_{MAJOR|MINOR|PATCH}").
+PackageVersion package_version();
+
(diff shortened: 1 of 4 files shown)


iceoryx2-cxx/include/iox2/version.hpp
elBoberido, line 1: Food for thought. In iceoryx classic, we generated a C header from the version set in cmake. See also -> https://github.com/eclipse-iceoryx/iceoryx/blob/main/iceoryx_platform/cmake/iceoryx_versions.h.in

This gave us the option to use the version info for deprecation macros like this -> https://github.com/eclipse-iceoryx/iceoryx/blob/main/iceoryx_hoofs/legacy/include/iceoryx_hoofs/cxx/optional.hpp#L22

It's on my todo list to also introduce this macros for iceoryx2 once we reached 1.0, in order to provide a smoother migration path if we deprecate an API call. So, having an analogous C header would help.

iceoryx2-cxx/include/iox2/version.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+#include "iox2/internal/iceoryx2.hpp"
+
+#include <cstdint>
+#include <iosfwd>
+
+namespace iox2 {
+
+/// Version number.
+struct PackageVersion {
```
elfenpiff, line 24: I want to provide a little background on the package version. It is currently stored in the first 8 bytes of any shared memory object iceoryx2 creates. When a user wants to connect to anything within iceoryx2, the package version is verified first, and only if it matches does the connection proceed.

The only construct lacking this is static service details (based on the static storage concept).

The question is, if we ever change the layout/implementation of the rust package version and if a C++/C/Python user might want to acquire the package version of a running iceoryx2 instance. If so, it would make sense to create a "classical" C binding for it and have just one implementation as the source of truth.

If this is a non-existing use case, this approach is fine. We can always provide CLI tooling to perform this task.

iceoryx2-cxx/include/iox2/version.hpp
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) 2026 Contributors to the Eclipse Foundation
```
elfenpiff, line 1: Could you please rename the file to `package_version.hpp`. Currently, the file name and the class name correspond - nearly everywhere. I would try to keep it consistent. Same goes for `package_version.cpp`

iceoryx2-cxx/include/iox2/version.hpp
elfenpiff, line 24: Why make it a struct with public accessible members?

I think this shouldn't be a generic building block that the user can use in their project to define their package version, but it is more the current package version of the iceoryx2 crate and in this case, the user should never change the version.

So I would turn it into a class and add 3 getter methods `major()`, `minor()` and `patch()`. This also reduces the API surface a bit.

elfenpiff: For testing we can add a c++ header file called `testing.hpp` with a testing name space and add a friend function that is able to instantiate different package versions to test the operators.

PR #1360, opened 2026-02-20
[#1300] Add framework for `no_std` testing

### LINKED ISSUE #1300: Support execution of  test suite on `no_std` POSIX platforms
## Brief Description

Some tests in the test suite still require the `std` module, meaning they cannot be executed on any `no_std` platform, including QNX8 which provides a POSIX API but not an `std` implementation.

The scope of this issue is to modify these tests so that they can be executed on such `no_std` POSIX platforms.

Not Doing:
1. Execution of test suite on all `no_std` platforms - support for this is on the roadmap, however is much more complex and left to a follow-up

orecham:
## Notes for Reviewer

Adds components to support execution of test on `no_std` targets (with a C runtime) and integrates three crates with it:
- `iceoryx2-bb-elementary`
- `iceoryx2-bb-concurrency`
- `iceoryx2-bb-threadsafe`

See the included development documentation for details.

Integration of remaining crates with the `no_std` testing framework will be added in subsequent PRs.

* [ ] Assign PR to reviewer

## PR Reviewer Reminders

* Public API is documented

## References

Relates #1300


.github/workflows/reuse_no_std.yml
```diff
@@ -18,11 +18,16 @@ jobs:
         uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9 # version: v1
         with:
           toolchain: ${{ matrix.toolchain }}
-          targets: armv7r-none-eabihf
+          targets: 
+            - armv7r-none-eabihf
+            - stable-x86_64-unknown-linux-gnu
 
       - name: Setup sccache
```
dkroenke, line 24: The `targets` parameter expects a comma-separated string. Maybe `stable-x86_64-unknown-linux-gnu` will not work as target triple.

```suggestion
          targets: armv7r-none-eabihf, x86_64-unknown-linux-gnu
```

iceoryx2-pal/concurrency-sync/src/strategy/mod.rs
```diff
@@ -0,0 +1,24 @@
...
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Concurrency synchronization primitives with pluggable wait/wake strategies.
+//!
+//! This crate provides synchronization primitives that implement the strategy
+//! pattern, allowing platform-specific waiting and wake-up mechanisms to be
+//! hooked in at runtime such as `WaitOnAddress` on Windows or `futex` on
+//! Linux.
+
+pub mod barrier;
```
orecham, line 20: @elfenpiff I moved these synchronization structs with implementations that take in a closure to hook into OS mechanisms into this `strategy` module. Open to alternative names for the module, but I guess it is the strategy pattern.

Reason is that I had to move the tests for these structs in the `bb` layer (since the `no_std` tests need access to bb components to implement bootstrap the test binaries) and I wanted to make the difference clear to other maintainers.

iceoryx2-bb/concurrency/Cargo.toml
```diff
@@ -21,7 +21,7 @@ iceoryx2-bb-elementary-traits = { workspace = true }
 iceoryx2-pal-concurrency-sync = { workspace = true }
 
 [dev-dependencies]
-iceoryx2-bb-concurrency-tests-common = { workspace = true, features = ["std"] }
-iceoryx2-bb-testing = { workspace = true, features = ["std"] }
+iceoryx2-bb-concurrency-tests-common = { path = "./tests-common", features = ["std"] }
```
orecham: From the [cargo reference](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html):

> Note: When a package is published, only dev-dependencies that specify a version will be included in the published crate. For most use cases, dev-dependencies are not needed when published, though some users (like OS packagers) may want to run tests within a crate, so providing a version if possible can still be beneficial.

When inheriting from the workspace, the version is always added therefore the dev dependencies are included and published. Do we want to publish the tests? If so, will need to find a solution for the (false positive) circular dependency caused by `tests-common`.

It seems to be a false positive for me, since the actual dependency graph should be tests -> tests-common -> library, but cargo only sees crate (tests) -> tests-common -> crate (library).

orecham: Perhaps this is a compromise that could be considered:
1. Publish the `tests-common` crates
2. Depend on the `tests-common` crate as a regular dependency in `conformance-tests`
    1. This could allow implementors of extensions to instantiate the conformance tests for their extensions
4. Depend on `tests-common` by path in the library crates tests (like shown here) so it is not considered by the dependency resolver when publishing

internal/scripts/release/crates_io_publish_script.sh
```diff
@@ -156,7 +164,7 @@ sanity_check_cyclic_dependencies() {
     for CRATE in "${CRATES_TO_PUBLISH[@]}"; do
         ALLOWED_CRATE_DEPENDENCIES+=("${CRATE}")
 
-        local CRATE_DEPENDENCIES=$(cargo tree --package "${CRATE}" --depth 1 --prefix none | grep -v '(\*)' | grep -e '^iceoryx2' | awk '{print $1}' | sort | uniq)
+        local CRATE_DEPENDENCIES=$(cargo tree --package "${CRATE}" --depth 1 --prefix none --edges normal | grep -v '(\*)' | grep -e '^iceoryx2' | awk '{print $1}' | sort | uniq)
```
orecham, line 167: This change makes the sanity check ignore dev-dependencies (https://doc.rust-lang.org/cargo/commands/cargo-tree.html#option-cargo-tree---edges)

iceoryx2-pal/posix/tests/win32_security_attributes_tests.rs
```diff
@@ -19,14 +19,13 @@ mod win32_security_attributes {
...
     };
-    use iceoryx2_pal_testing::assert_that;
     use windows_sys::Win32::Foundation::INVALID_HANDLE_VALUE;
 
     fn roundtrip(orig_mode: posix::mode_t) {
         let attr = from_mode_to_security_attributes(INVALID_HANDLE_VALUE, orig_mode);
         let mode = from_security_attributes_to_mode(&attr);
 
-        assert_that!(orig_mode, eq mode);
+        assert_eq!(orig_mode, mode);
```
orecham, line 28: Removed use of `assert_that!` since `iceoryx2-pal-testing` needs to depend on `iceoryx-pal-posix` for some components.

A different organization of the components that depend on the platform is warranted, but left to another PR.

iceoryx2-pal/concurrency-sync/tests/atomic_tests.rs
orecham, line 1: All tests need to be in building blocks layer or above to work in `no_std`. The solution adopted is to re-export these types in the building blocks layer for testing.

iceoryx2-pal/posix/src/linux/timer.rs
```diff
@@ -0,0 +1,41 @@
...
+// which is available at https://opensource.org/licenses/MIT.
+//
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+#![allow(non_camel_case_types)]
+#![allow(clippy::missing_safety_doc)]
+
+use crate::posix::types::*;
+
+pub unsafe fn timer_create(
```
orecham, line 18: These timer syscalls are exposed to support implementation of Watchdog without use of thread syscalls. The idea is that a timer-based implementation would be easier to support in bare metal.

BEST_PRACTICES.md
```diff
@@ -117,8 +117,7 @@
 * Test at least runtimes with
   `assert_that!(start.elapsed(), time_at_least TIMEOUT)`
 * Do not wait for events (indefinitely), use
-  `assert_that!(|| { some_condition }, block_until_true)`
-    * It starts a `iceoryx2_bb_testing::watchdog::Watchdog` in the background that
-    terminates the test when it deadlocks.
+  `assert_that!(|| { some_condition }, eq true, before Watchdog::default())`
```
orecham, line 120: The `block_until` directive was removed to decouple `assert_that` from the `Watchdog`.

In the end was not needed, but this decoupling is more flexible so will keep it. Open for suggestions on the API if anyone has another preference.

iceoryx2-bb/elementary/src/bump_allocator.rs
```diff
@@ -11,9 +11,9 @@
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
 use crate::math::align;
-use iceoryx2_bb_concurrency::atomic::AtomicUsize;
-use iceoryx2_bb_concurrency::atomic::Ordering;
 use iceoryx2_bb_elementary_traits::allocator::{AllocationError, BaseAllocator};
+use iceoryx2_pal_concurrency_sync::atomic::AtomicUsize;
```
orecham, line 15: This is dependency on the pal crate sub-optimal, may be worth moving these into a crate separate from the layered architecture.

Left to another PR.

iceoryx2-bb/testing-nostd-macros/src/inventory_test_generic.rs
orecham, line 1: This generic test implementation could be decoupled from `inventory` and used in place of the `generic_tests` crate.

Left to another PR.

Cargo.toml
```diff
@@ -120,6 +127,16 @@ iceoryx2-tunnel-conformance-tests = { version = "0.8.999", path = "iceoryx2-tunn
 iceoryx2-tunnel-zenoh = { version = "0.8.999", path = "iceoryx2-tunnel/zenoh" }
 iceoryx2-userland-record-and-replay = { version = "0.8.999", path = "iceoryx2-userland/record-and-replay" }
 
+iceoryx2-bb-testing-nostd = { path = "iceoryx2-bb/testing-nostd/" }
```
orecham, line 134: Versions removed from these crates to disable publishing them to crates.io.

.just/build.just
```diff
@@ -0,0 +1,100 @@
+# =============================================================================
```
elfenpiff, line 13: Please create a follow up issue:

Since this is completely new, could you please add a doc document explaining the philosophy behind just and why we use it instead of the shell scripts.
Also, we should have a consistent strategy and then port the old shell scripts to just.
And a little guide how to write/create just ?scripts? and the best practices - if we have any.

elfenpiff, line 13: Please add the license header to all `just` files and adjust the license header shell script so that also ".just" files are covered - I know - but it is required by the eclipse foundation.

.just/build.just
```diff
@@ -0,0 +1,100 @@
...
+# Build Recipes
+# =============================================================================
+
+[private]
+_build-dispatch target *flags:
+    #!/usr/bin/env bash
+    set -euo pipefail
```
elfenpiff, line 20: Why is this required ... maybe add it to the follow up issue.

doc/development-setup/nostd-testing.md
```diff
@@ -0,0 +1,321 @@
...
+fn my_test_case() {
+    module_a::my_test_case();
+}
+```
+
+A macro for defining generic tests for the `no_std` test framework is also
+available.
+
+```rust
+#[inventory_test_generic(TypeA, TypeB)]
```
elfenpiff, line 188: Can we write a simple shell script that checks that the test cases in `no_std`, in `std` and in the common library are the same. I just would like to have at least a CI check that is red when I forgot to add a test in one of the libraries.

doc/development-setup/nostd-testing.md
```diff
@@ -0,0 +1,321 @@
...
+```
+
+### Custom Framework
+
+Tests using the custom `no_std` framework are executed via the binary produced
+by the crate containing the `no_std` tests. In addition, the `core` and `alloc`
+libraries must be built with `panic=abort`:
+
+```console
+RUSTC_BOOTSTRAP=1 RUSTFLAGS="-C panic=abort" \
```
elfenpiff, line 299: Could we add this to `just` as well and call it `just build_no_std`?

iceoryx2-bb/concurrency/tests-common/src/spin_lock_tests.rs
```diff
@@ -0,0 +1,114 @@
...
+                let mut guard = lk.blocking_lock();
+                *guard += 1;
+            });
+        }
+    });
+    let guard = lk.blocking_lock();
+    assert_that!(*guard, eq NUMBER_OF_THREADS);
+}
+
+#[requires_std("threading", "watchdog", "barrier")]
```
elfenpiff, line 69: What happens when I miss one element? Like for instance `barrier`. Is then the CI red? Or does something not compile?

doc/development-setup/nostd-testing.md
orecham: @elfenpiff It is, this was one of the reason to use `just`. All these details are hidden if you use it (explained at the end of this doc).

iceoryx2-bb/concurrency/tests-common/src/spin_lock_tests.rs
orecham: @elfenpiff This is only for manual tracking (to assist with the work to remove these dependencies). Any test annotated with `requires_std` is completely compiled out. There is some documentation on this macro, should it be improved?

doc/development-setup/nostd-testing.md
orecham: This will be addressed in a follow-up. It is tracked here: https://github.com/eclipse-iceoryx/iceoryx2/issues/1300#issuecomment-3971327234

Could also be a separate issue.

.just/build.just
orecham: @elfenpiff Perhaps some tasks do not need some of them, but as a default it seems sane to me. What is the concern?

.just/build.just
orecham: @elfenpiff Done

doc/development-setup/nostd-testing.md
orecham: https://github.com/eclipse-iceoryx/iceoryx2/issues/1404

.just/build.just
orecham: https://github.com/eclipse-iceoryx/iceoryx2/issues/1408

iceoryx2-bb/testing/src/lifetime_tracker.rs
```diff
@@ -11,77 +11,48 @@
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
 use iceoryx2_bb_elementary_traits::zero_copy_send::ZeroCopySend;
-use std::sync::{Mutex, MutexGuard};
 
-static CREATION_COUNTER: Mutex<usize> = Mutex::new(0);
-static DROP_COUNTER: Mutex<usize> = Mutex::new(0);
+#[cfg(feature = "std")]
```
elfenpiff, line 15: I think this can be removed, I do not see any mutex being used in this file.

iceoryx2-bb/testing/src/lifetime_tracker.rs
```diff
@@ -11,77 +11,48 @@
...
+use std::sync::MutexGuard;
 
-static DROP_ORDER: Mutex<Vec<usize>> = Mutex::new(vec![]);
-static TRACKING_LOCK: Mutex<LifetimeTrackingState> = Mutex::new(LifetimeTrackingState {});
+#[cfg(not(feature = "std"))]
```
elfenpiff, line 18: I think this can be removed, I do not see any mutex being used in this file.

Is clippy checking this file?

iceoryx2-bb/threadsafe/tests-common/src/trigger_queue_tests.rs
```diff
@@ -0,0 +1,343 @@
...
+    let used_handle = UnnamedSemaphoreHandle::new();
+
+    let sut = Sut::new(&mtx_handle, &free_handle, &used_handle);
+
+    let start = Time::now().unwrap();
+    assert_that!(sut.timed_pop(TIMEOUT), is_none);
+    assert_that!(start.elapsed().unwrap(), time_at_least TIMEOUT);
+}
+
+#[requires_std("threading")]
```
elfenpiff, line 157: Is there a list of all possible "requires" elements? If I would have written this test, I would have written `#[requires_std("thread")]`

iceoryx2-pal/concurrency-sync/src/lazy_lock.rs
```diff
@@ -87,23 +87,23 @@ impl<T, F: FnOnce() -> T> LazyLock<T, F> {
     ///
     /// This method will block the calling thread if another initialization
     /// routine is currently running.
-    pub fn force(&self) -> &T {
+    pub fn force(this: &Self) -> &T {
```
elfenpiff, line 90: Is clippy checking this code? I am wondering why you write `this: &Self` instead of `&self`?

iceoryx2-pal/posix/tests/win32_security_attributes_tests.rs
elfenpiff: Please create an issue for that.

iceoryx2-pal/testing/src/watchdog.rs
```diff
@@ -10,53 +10,160 @@
...
+        }
+    }
+
+    /// Called by the kernel when `SIGALRM` fires.
+    ///
+    /// `panic!` is not async-signal-safe (it may allocate and unwind), so
+    /// [`posix::abort()`] is used instead to terminate immediately.
+    unsafe extern "C" fn handler(_sig: int) {
+        // write() syscall is permitted
+        cerrln!("Watchdog expired. Aborting.");
```
elfenpiff: Can we have a consistent error message when the watchdog expires? You could create a `fn watchdog_expired_msg() -> &str` call and call it here and in line 157 so that both WatchDog variants behave identical from the outside look and feel.

internal/scripts/release/crates_io_publish_script.sh
elfenpiff: But do we also need to adjust the crates.io publish call? We already had a 0.x.1 release exactly because of this issue.

Cargo.toml
elfenpiff: Can you please add this as a comment above and below this section so that we do not mix it by accident or reintroduce version numbers.

Something like

```rust
########
### START: the crates shall never be published therefore they do not have version numbers
########

...

########
### END: the crates shall never be published therefore they do not have version numbers
########
```

iceoryx2-bb/threadsafe/tests-common/src/trigger_queue_tests.rs
orecham: @elfenpiff It‘s just a string, anything can be written

iceoryx2-pal/concurrency-sync/src/lazy_lock.rs
orecham: @elfenpiff It‘s to match the std version, this was taken from that

.just/build.just
elfenpiff: No concern but if it makes sense we should add a hint to the docs that every shell script should start with:

```sh
#!/usr/bin/env bash   <- for platform independence
set -euo pipefail  <- for whatever??
```

iceoryx2-bb/concurrency/tests-common/src/spin_lock_tests.rs
elfenpiff: Aaah, then I got the notion wrong. I thought that this macro checks some internal list and only deactivates the test if the provided elements are not available.

iceoryx2-bb/testing/src/lifetime_tracker.rs
orecham: @elfenpiff It's the return value of `start_tracking()`

iceoryx2-pal/testing/src/watchdog.rs
orecham: @elfenpiff Can do it, but in the signal handler we cannot build a dynamic string like in line 157. should we have this basic string like here?

internal/scripts/release/crates_io_publish_script.sh
orecham: @elfenpiff I wondered too, but @elBoberido has the opinion that if the current one passes there should be no problem.

iceoryx2-pal/testing/src/watchdog.rs
orecham: @elfenpiff ~~Scratch that, it is not trivial to achieve whilst also satisfying the async signal safe requirements.~~

I added something that should work.

.just/build.just
orecham: @elfenpiff Not sure about that.. adding entries for small things like this is not a convention that makes sense to me. If we added entries for everything at this triviality level the doc would be quite bloated / less likely to be read / unmaintainable. Would leave it up to the script writer. What's your reasoning? Maybe I am not following the concern.

internal/scripts/release/crates_io_publish_script.sh
elBoberido: @orecham you got me wrong. With the current one I meant the current check including the dev dependencies. But since there is now the `--workspace` option for publishing, it should be fine as long as the dry run works. Maybe we need to publish to a kellnr instance to be safe, before we do the next release.

Cargo.toml
elBoberido: My obligatory

> Using `BEGIN` and `END` will allow code folding on some editors

iceoryx2-pal/testing/src/watchdog.rs
elBoberido: @orecham @elfenpiff unfortunately it is only available with Rust 1.93 but `fmt::from_fn` is supposed to be used with `format!` without allocating and should therefore be signal safe. So nothing for now but when we starte certification, this is probably something we need to use.

internal/scripts/release/crates_io_publish_script.sh
elfenpiff: @elBoberido @orecham I am wondering if we have to add the option `--edges normal` also to the publish command?

I am pretty sure, that the cargo publish command once failed due to circular dependencies in the `dev-dependencies` section.

.just/build.just
elfenpiff: > If we added entries for everything at this triviality level the doc would be quite bloated
> What's your reasoning?

I agree, but I think this is not a triviality since I never used this and was also unaware of its existence.

elfenpiff (approved):
The remaining things:

* the dev doc with the pipefail thingy (which can also be skipped if you think it's trivial)
* the publish command adjustment to avoid circular dependency problems

can be done in a follow-up PR.

@orecham You did an awesome job here!

PR #1364, opened 2026-02-23
Set gil_used = false on pymodule

mscheltienne:
I haven't followed the proper contribution process- so feel free to close and re-open with similar changes.
This PR makes the iceoyx2 compatible with free-threaded python. Currently the import triggers:

```
.venv/lib/python3.14t/site-packages/iceoryx2/__init__.py:15: in <module>
    from ._iceoryx2 import *
E   RuntimeWarning: The global interpreter lock (GIL) has been enabled to load module 'iceoryx2._iceoryx2', which has not declared that it can run safely without the GIL. To override this behavior and keep the GIL disabled (at your own risk), run with PYTHON_GIL=0 or -Xgil=0.
```

With those changes- an import of `iceoryx2` doesn't force re-enable the GIL. The added jobs should yield:

```
  - iceoryx2-*-cp38-abi3-manylinux_*.whl
  - iceoryx2-*-cp38-abi3-macosx_*.whl
  - iceoryx2-*-cp38-abi3-win_amd64.whl
  - iceoryx2-*-cp314-cp314t-manylinux_*.whl
  - iceoryx2-*-cp314-cp314t-macosx_*.whl
  - iceoryx2-*-cp314-cp314t-win_amd64.whl
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -43,8 +43,45 @@ jobs:
           path: target/ff/python/dist
           retention-days: 7
 
+  build-python-wheels-freethreaded:
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-latest
+          - macos-latest          # aarch64
+          - windows-latest
+        python:
+          - "3.14t"
+    defaults:
+      run:
+        shell: bash               # required for windows
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # version v6.0.1
+
+      - name: Set Up Free-Threaded Python
+        uses: actions/setup-python@83679a892e2d95755f2dac6acb0bfd1e9ac5d548 # version: v6.0.1
+        with:
+          python-version: ${{ matrix.python }}
+
+      - name: Build Wheel
+        run: |
+          pip install "maturin>=1.8,<2.0"
+          cd iceoryx2-ffi/python
+          maturin build --release --target-dir ../../target/ff/python --out ../../target/ff/python/dist
+
+      - name: Upload Wheel
+        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # version: v6.0.0
+        with:
+          name: wheels-freethreaded-${{ runner.os }}-${{ runner.arch }}-py${{ matrix.python }}-${{ github.sha }}
+          path: target/ff/python/dist
+          retention-days: 7
+
   prepare-python-dist:
-    needs: [build-python-wheels]
+    needs: [build-python-wheels, build-python-wheels-freethreaded]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # version v6.0.1


elfenpiff:
@mscheltienne 

> I haven't followed the proper contribution process- so feel free to close and re-open with similar changes.

Thanks for the contribution. We usually prefer that the one who contributed the code also merges it and then becomes a member of the glorious hall of fame of iceoryx2 contributors.

For this, you would need to:
1. sign the eclipse committer agreement (ECA), see: https://www.eclipse.org/legal/eca/ here you need to create an account
2. create an issue that describes your change
3. add the issue number to every commit message as prefix like `[#123] some commit message`

But if you prefer, I can also take over the PR and create the issue and commits under my name.

mscheltienne:
Thanks- but please take over the PR under the name- I have too limited bandwidth at the moment.

elBoberido:
@mscheltienne do you have time to sign the ECA? We can take care of the PR but your initial contribution still has your copyright and needs to be covered with the ECA.

mscheltienne:
> Email addresses with dots (.) or plus signs (+) before the @ are not allowed. (6658s) 

I mean.. my email is `firstname.lastname@...` - any guidance on creating the account on your platform?

iceoryx2-ffi/python/src/lib.rs
```diff
@@ -115,7 +115,7 @@ pub(crate) use service_type::IpcService;
 pub(crate) use service_type::LocalService;
 
 /// iceoryx2 Python language bindings
-#[pymodule]
+#[pymodule(gil_used = false)]
```
elfenpiff, line 118: @mscheltienne I inspected our code and saw that we use the python gil in

* https://github.com/eclipse-iceoryx/iceoryx2/blob/main/iceoryx2-ffi/python/src/type_storage.rs#L32
* https://github.com/eclipse-iceoryx/iceoryx2/blob/main/iceoryx2-ffi/python/src/service_builder_blackboard.rs#L120
* https://github.com/eclipse-iceoryx/iceoryx2/blob/main/iceoryx2-ffi/python/src/service_builder_blackboard.rs#L137

If those 3 files are fixed we can apply your patch.

mscheltienne: This is way out of my area of expertise, so I will let you investigate further. Note that we would like to use the SW you developed for us in 3.14t which makes this change important.

Here are a couple of pointers from Claude's analysis which will make more sense to you and might help carve the right way forward here:

> As illustrated above, attaching and detaching from the runtime uses exactly the same code in the free-threaded build as is used in the GIL-enabled build to acquire and release the GIL. It is an unfortunate naming issue that PyGILState_Ensure and PyGILState_Release has "GIL" in the name of the function, despite the lack of a GIL on the free-threaded build. It's likely that the C API in future Python version will fix this naming issue.

From: https://py-free-threading.github.io/porting-extensions/

```
---                                                                                                                                                                                                                            
What Python::with_gil Actually Does in Free-Threaded Python                                                                                                                                                                    
                                                                                                                                                                                                                                
At the bottom of the stack, Python::with_gil calls PyGILState_Ensure() — the same C API call in both GIL-enabled and free-threaded builds. PyO3 does not differentiate.                                                        

What changes is what that C function does at the CPython level:                                                                                                                                                                
                                                                                                                                                                                                                                
┌──────────────────────┬──────────────────────────────────────────────────────────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐                           
│                      │                    GIL-enabled Python                    │                                       Free-threaded Python (3.13t/3.14t)                                       │
├──────────────────────┼──────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ PyGILState_Ensure()  │ Acquires the GIL (blocks until no other thread holds it) │ Attaches the thread to the interpreter runtime (non-blocking, multiple threads can be attached simultaneously) │
├──────────────────────┼──────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ PyGILState_Release() │ Releases the GIL                                         │ Detaches the thread from the runtime                                                                           │
├──────────────────────┼──────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Effect               │ Serializes all Python access                             │ Merely registers the thread so it may call C API functions                                                     │
└──────────────────────┴──────────────────────────────────────────────────────────┴────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘

The naming is unfortunate — as the Python free-threading porting guide notes: "It is an unfortunate naming issue that PyGILState_Ensure and PyGILState_Release has 'GIL' in the name of the function, despite the lack of a GIL
  on the free-threaded build."

So Python::with_gil in free-threaded mode = "attach this thread to the interpreter, run the closure, then detach." It does not serialize anything.

---
Risks

Site 1: TypeStorage::clone() — v.clone_ref(py)

clone_ref increments the Python object's reference count. In free-threaded CPython, reference counting uses a split scheme: a per-thread local count plus a shared atomic count. Py_INCREF is thread-safe in free-threaded
builds. No data-race risk here.

The real issue is performance and unnecessary overhead: every TypeStorage::clone() call does a full PyGILState_Ensure → attach → Py_INCREF → PyGILState_Release → detach cycle, even though every caller is already inside a
#[pymethods] function that is already attached. It works, but it's wasteful.

Sites 2 & 3: __set_key_eq_cmp_func callbacks — key_eq_func.call1(py, ...)

These closures are captured and stored, then invoked later from deep inside iceoryx2's Rust core during key comparison. At that point there is no Python token available — with_gil/attach is the only way to call back into
Python.

The risk here is real but narrow: calling a Python callable without GIL protection means another thread could be simultaneously mutating the same Python objects. Specifically:

1. The key_eq_func PyObject itself is safe to hold across threads (Py<T> is Send + Sync), and calling call1 is safe because the thread is attached.
2. However, the Python function being called (the key_cmp closure in blackboard_extensions.py) does ctypes.cast(lhs, ctypes.POINTER(...)).contents and then field comparison. In free-threaded Python, if another thread were
simultaneously modifying the key data in shared memory, you'd have a data race on the iceoryx2 shared memory side, not the Python side. But iceoryx2's internal locking (lock-free data structures) ensures the key comparison
callback is called on stable data — the pointers lhs and rhs point to data that won't be concurrently modified during comparison.
3. The Parc<T> (Arc<Mutex<T>>) wrapping the builder means the builder itself won't be concurrently accessed from multiple threads.

Bottom line: these callbacks are safe in practice because the iceoryx2 core guarantees the pointed-to data is stable during comparison, and the Python callable is a pure function over those pointers.

---
Summary

┌───────────────────────────────────┬────────────────────────────────────────┬─────────────────────────────────────────────────────────────────────────┬──────────────────────────────────────────────────────────────┐
│               Site                │  What happens in free-threaded Python  │                                  Risk                                   │                        Action needed?                        │
├───────────────────────────────────┼────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┤
│ type_storage.rs:32                │ Attach → Py_INCREF (atomic) → Detach   │ None (ref counting is thread-safe). Unnecessary attach/detach overhead. │ Optional refactor for cleanliness — pass py from caller      │
├───────────────────────────────────┼────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┤
│ service_builder_blackboard.rs:120 │ Attach → call Python callable → Detach │ Low — the callable is a pure function over stable pointers              │ None — this is the correct pattern for Rust→Python callbacks │
├───────────────────────────────────┼────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┤
│ service_builder_blackboard.rs:137 │ Same as above                          │ Same as above                                                           │ None                                                         │
└───────────────────────────────────┴────────────────────────────────────────┴─────────────────────────────────────────────────────────────────────────┴──────────────────────────────────────────────────────────────┘

The PR's gil_used = false annotation is safe to apply as-is. The three Python::with_gil calls are compatible with free-threaded Python — they simply become attach/detach operations. There is no correctness issue; the
with_gil calls don't need to be "fixed" for the PR to be correct, despite the reviewer's comment suggesting otherwise. They may be refactored for performance or code cleanliness, but they are not blockers.
```

elfenpiff: I googled a bit and this documentation seems to contradict the claude code argument: https://mejrs.github.io/pyo3/internal/doc/pyo3/index.html#the-gil-dependent-types

> For example &[PyAny](https://mejrs.github.io/pyo3/internal/doc/pyo3/prelude/struct.PyAny.html). These are only ever seen as references, with a lifetime that is only valid for as long as the GIL is held, which is why using them doesn’t require a [Python<'py>](https://mejrs.github.io/pyo3/internal/doc/pyo3/marker/struct.Python.html) token.

This needs some more effort to support this in iceoryx2, especially since we have users that require a python 3.8 ABI - see `iceoryx2-ffi/python/Cargo.toml`

Another thing is, that the `gil_used = false` is set to true as soon as the gil is once used, like in the code snippets I listed above, therefore your patch would not establish a gil free context.

mscheltienne: Again, I'm out of my expertise field and without the bandwidth to properly document myself, so I'll put the AI's answer in a dropdown- maybe it's useful in the future to move this ahead. Short version is that `iceoryx2` usage falls under the GIL-indepenent types in the PyO3's internal type taxonomy- and that `gil_used = false` is not set to true as soon as the GIL is once used; but instead during the module loading. This second point would be coherent with the warning message `RuntimeWarning: The global interpreter lock (GIL) has been enabled to load module 'iceoryx2._iceoryx2'`.

<details>
<summary>Claude's reply</summary>

```
Claim 1: The &PyAny documentation contradicts the analysis                                                                                  

The linked page describes PyO3's internal type taxonomy — "GIL-dependent types" (&PyAny, &PyList, etc.) vs "GIL-independent types"          
(Py<PyAny>, aka PyObject). This is about PyO3's type system design, not about runtime GIL behavior.                                         
                                                                                                                                            
The iceoryx2 Python bindings use PyObject (= Py<PyAny>), which is a GIL-independent type. It can be stored, moved across threads, and held
without the GIL. You just need a Python<'_> token to operate on it (increment refcount, call it, etc.). The old &PyAny GIL-dependent
references are part of the legacy "GIL Refs" API (deprecated since PyO3 0.21 in favor of Bound<'py, T>). This documentation does not
contradict anything — it's describing a different concept entirely.

Claim 2: "gil_used = false is set to true as soon as the GIL is once used"

This is incorrect. Here's what actually happens at each layer:

PyO3 layer: #[pymodule(gil_used = false)] compiles to a static, const module slot declaration. From the PyO3 source:

// This is const — baked into the binary at compile time
pub const fn with_gil_used(self, gil_used: bool) -> Self {
    self.push(ffi::Py_mod_gil,
        if gil_used { ffi::Py_MOD_GIL_USED }
        else { ffi::Py_MOD_GIL_NOT_USED })
}

This slot gets passed to PyModuleDef_Init() during module load. There is no runtime mechanism that flips it back.

CPython layer: The interpreter reads the Py_mod_gil slot at import time only. If a module declares Py_MOD_GIL_NOT_USED, the GIL stays
disabled. If it's missing or set to Py_MOD_GIL_USED, the GIL gets permanently re-enabled for the process. This is a one-way decision at
import time.

PyGILState_Ensure() at runtime: In free-threaded Python, this function attaches the thread to the interpreter runtime — it does NOT
re-enable the GIL. As the free-threading porting guide explicitly notes: "It is an unfortunate naming issue that PyGILState_Ensure and
PyGILState_Release has 'GIL' in the name of the function, despite the lack of a GIL on the free-threaded build."

So: Python::with_gil → PyGILState_Ensure → thread attach. No GIL re-enablement.
```

</details>

We can shelve that for now and come back to it when we have more bandwidth- I believe I can ignore the warning and skip the GIL re-enabling in `3.14t` with an environment variable configuration, but that has downside (since it's a global configuration that will affect every package and prevent us from spotting other dependencies that might otherwise re-enable the GIL and might not be thread-safe).

elfenpiff:
@mscheltienne I will close this PR and the progress is tracked via the issue #1368 .

Issue #1223, opened 2025-12-05
ServiceVariant based on shared memory files in an in-memory fs.


elfenpiff:
### Discussed in https://github.com/eclipse-iceoryx/iceoryx2/discussions/1220

It is currently not possible to contain all shared memory files in a separate directory like `/dev/shm/whatever` since the `shm_open` call does not support it in POSIX systems.

However, one can use explicitly use `open` to create a file and map it into the process space. If the file is stored in an in-memory fs like tmpfs then we should get the identical performance and the user should be able to define the folder for all files via the global iceoryx2 configuration.

PR #1372, opened 2026-02-25
[#1370] Fix cleanup with 'dev_permissions' feature flag

### LINKED ISSUE #1370: Resources cannot always be cleaned up with 'dev_permissions' feature flag
Before posting the bug, take a look at the
[FAQ](https://github.com/eclipse-iceoryx/iceoryx2/blob/main/FAQ.md)
for a possible solution.

## Required information

**Operating system:**

* Linux, 2.19
* other OSes are probably also affected

**Rust version:**

any

**Cargo version:**

any

**iceoryx2 version:**

Probably present since the introduction of the `dev_permissions` feature flag

**Detailed log output:**

> 1 [E] SharedMemory { name: FileName { value: StaticString<255> { len: 80, data: "iox2_0354a209029e7d094a819e2d4030ea331e6caaf0_95202013074162211780128991149.data" } }, has_ownership: AtomicBool(true), memory_mapping: MemoryMapping { file_descriptor: Some(FileDescriptor { value: 6, is_owned: true }), file_path: None, base_address: 0x7f12e807c000, size: 2470, offset: 0 }, memory_lock: None, mapping_offset: 0 }
| Failed to cleanup shared memory since the permissions could not be adjusted (InsufficientPermissions).
> 2 [E] "Service::remove_node_from_service(NodeId(UniqueSystemId { value: 15948803921282728757882316717, pid: 59309, creation_time: Time { clock_type: Monotonic, seconds: 14295, nanoseconds: 864586393 } }), ServiceId(RestrictedFileName { value: StaticString<64> { len: 40, data: \"4eacadf2695a3f4b2eb95485759246ce1a2aa906\" } }))"
| Unable to remove static config of unused service (InternalError).
> 3 [E] "Service::remove_node_from_service(NodeId(UniqueSystemId { value: 15948803921282728757882316717, pid: 59309, creation_time: Time { clock_type: Monotonic, seconds: 14295, nanoseconds: 864586393 } }), ServiceId(RestrictedFileName { value: StaticString<64> { len: 40, data: \"4eacadf2695a3f4b2eb95485759246ce1a2aa906\" } }))"
| Unable to remove static config of unused service (InternalError).
Error: PublishSubscribeOpenError(IsMarkedForDestruction)

When the `dev_permissions` flag is used, any app is entitles to perform the cleanup. But when an app terminates abnormally and the app to cleanup runs under a different user than the one that terminated, the permission cannot be changed and the error with the log from above occurs.

The cleanup works with the `dev_permissions` feature flag even it the resources to be cleaned up were created by an application running under a different user.

Terminal 1:
```
cargo build --example publish_subscribe_publisher --features iceoryx2/dev_permissions
sudo -u testuser -- target/debug/examples/publish_subscribe_publisher
```

Terminal 2:
```
sudo killall -9 publish_subscribe_publisher
target/debug/examples/publish_subscribe_publisher
```

**Underlying issue:**

Changing the file permissions to have write access before deleting the file is required (at least) on Windows. At some locations, the failure to change the file permissions generates only a warning instead of an error. It needs to be checked whether the same approach also works for the `set_permission` call in:

* iceoryx2-cal/src/static_storage/file.rs
* iceoryx2-bb/posix/src/shared_memory.rs

If not, the change of the permissions needs to be abstracted away is the PAL layer and performed only on the platforms that require this.

elBoberido:
## Notes for Reviewer

This PR addresses the issue where resources cannot always be cleaned up, even with the `dev_permissions` flag. The problem occurs when attempting to change file permissions before deletion, as a process running under a different user cannot modify permissions, even if they are already set to 777.

Instead of raising an error, the PR now prints a warning if changing file permissions fails. In such cases, the subsequent deletion attempt will fail on OSes that require files to be writable before deletion, e.g. on Windows.

This behavior, printing a warning instead of an error when setting permissions before deletion, is already used elsewhere in the codebase, so this change does not introduce a regression on Windows.

* [x] Assign PR to reviewer

## PR Reviewer Reminders

* Public API is documented

## References

Closes #1370

### DIFF
--- a/iceoryx2-bb/posix/src/shared_memory.rs
+++ b/iceoryx2-bb/posix/src/shared_memory.rs
@@ -420,16 +420,17 @@ pub struct SharedMemory {
 impl Drop for SharedMemory {
     fn drop(&mut self) {
         if self.has_ownership() {
-            if let Err(e) = self.set_permission(Permission::OWNER_ALL) {
-                warn!(from self,
-                      "Unable to adjust the files permission as preparation to remove the file ({e:?}).");
-            }
+            let set_permission_result = self.set_permission(Permission::OWNER_ALL);
 
             match Self::shm_unlink(&self.name) {
                 Ok(_) => {
                     trace!(from self, "delete");
                 }
                 Err(_) => {
+                    if let Err(e) = set_permission_result {
+                        warn!(from self,
+                              "Unable to adjust the files permission as preparation to remove the file ({e:?}).");
+                    }
                     error!(from self, "Failed to cleanup shared memory.");
                 }
             }
--- a/iceoryx2-cal/src/static_storage/file.rs
+++ b/iceoryx2-cal/src/static_storage/file.rs
@@ -216,21 +216,26 @@ impl crate::named_concept::NamedConceptMgmt for Storage {
             }
         };
 
-        if let Err(e) = file.set_permission(Permission::ALL) {
-            warn!(from origin,
-                  "Unable to adjust the files permission as preparation to remove the file ({e:?}).");
-        }
+        let set_permission_result = file.set_permission(Permission::ALL);
 
         match File::remove(&file_path) {
             Ok(v) => Ok(v),
-            Err(FileRemoveError::InsufficientPermissions)
-            | Err(FileRemoveError::PartOfReadOnlyFileSystem) => {
-                fail!(from origin, with NamedConceptRemoveError::InsufficientPermissions,
-                        "{} due to insufficient permissions.", msg);
-            }
-            Err(v) => {
-                fail!(from origin, with NamedConceptRemoveError::InternalError,
-                        "{} due to unknown failure ({:?}).", msg, v);
+            Err(e) => {
+                if let Err(e) = set_permission_result {
+                    warn!(from origin,
+                          "Unable to adjust the files permission as preparation to remove the file ({e:?}).");
+                }
+                match e {
+                    FileRemoveError::InsufficientPermissions
+                    | FileRemoveError::PartOfReadOnlyFileSystem => {
+                        fail!(from origin, with NamedConceptRemoveError::InsufficientPermissions,
+                                "{} due to insufficient permissions.", msg);
+                    }
+                    _ => {
+                        fail!(from origin, with NamedConceptRemoveError::InternalError,
+                                "{} due to unknown failure ({:?}).", msg, e);
+                    }
+                }
             }
         }
     }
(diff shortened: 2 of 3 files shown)


iceoryx2-bb/posix/src/shared_memory.rs
```diff
@@ -420,17 +420,17 @@ pub struct SharedMemory {
...
-                        trace!(from self, "delete");
-                    }
-                    Err(_) => {
-                        error!(from self, "Failed to cleanup shared memory.");
-                    }
-                },
-                Err(e) => {
-                    error!(from self, "Failed to cleanup shared memory since the permissions could not be adjusted ({:?}).", e);
+            if let Err(e) = self.set_permission(Permission::OWNER_ALL) {
+                warn!(from self,
```
elfenpiff: Can you please print this warning only when `Self::shm_unlink` has failed. If `shm_unlink` succeeds this warning is meaningless.

iceoryx2-cal/src/static_storage/file.rs
```diff
@@ -216,9 +216,10 @@ impl crate::named_concept::NamedConceptMgmt for Storage {
             }
         };
 
-        fail!(from origin, when file.set_permission(Permission::OWNER_ALL),
-                with NamedConceptRemoveError::InternalError,
-                "{} since the permissions could not be adjusted.", msg);
+        if let Err(e) = file.set_permission(Permission::ALL) {
+            warn!(from origin,
```
elfenpiff: Same thing here, please print this warning only when `File::remove` has failed to provide more context on why it have failed. Otherwise, it is somehow meaningless.

iceoryx2-bb/posix/src/shared_memory.rs
elBoberido: This is analogues to other places where the warning is also printed independently of follow up error. Shall it be changed also at those locations?

elBoberido:
@elfenpiff there is still the issue that the sticky bit is set on `/dev/shm`, which means that only the owner can remove the shared memory. So, I guess with `dev_permissions`, one needs to remove the sticky bit from `/dev/shm`, right? This might not be desirable. An alternative would be to have a separate in-memory filesystem for iceoryx2, separate from `/dev/shm` like described in #1223

elfenpiff:
@elBoberido yes you are right. We could achieve this by using the file variant of the dynamic storage, shared memory, resizeable shared memory and zero copy connection in the recommended.rs files when the dev_permissions flag is set.

But that would mean that the users must ensure that `/tmp/` is an in-memory fs.

PR #1378, opened 2026-02-27
[#1374] Use `libc` automatically on platforms that support it

### LINKED ISSUE #1374: Remove `libc_platform` feature
## Brief Description

Propagation of the feature `libc_platform` is error-prone and difficult to maintain. However, the ability to use the `libc` crate instead of `bindgen` is useful in some cases e.g. cross-compilation.

Additionally, the `libc_platform` feature does not really adhere to the convention of "features adding new functionality".

Options to consider:
1. Use `libc` as default on platforms that supports it
2. Leverage a build `cfg` attribute instead of a feature

orecham:
## Notes for Reviewer

~~Removes the feature `libc_platform` and instead allows build to be configured to use `libc` via the environment variable `IOX2_PLATFORM_BINDING`.~~

Removes the feature `libc_platform` and utilizes `libc` automatically where supported.

This removes the difficulty of properly propagating the feature through the crates to enable builds with `libc`.

* [ ] Assign PR to reviewer

## PR Reviewer Reminders

* Public API is documented

## References

Closes #1374


.github/workflows/build-test.yml
orecham, line 1: @dkroenke FYI some tweaks to the CI

FAQ.md
```diff
@@ -560,14 +559,6 @@ sender closes the data segment when going out of scope.
 To circumvent this, you could use the size of the last sent sample as
 `initial_max_slice_len` and use the `Static` allocation strategy.
 
-<!-- markdownlint-disable MD044 'c' needs to be lower-case -->
-### `iceoryx2-ffi-c` Does Not Contain This Feature: `libc_platform`
-<!-- markdownlint-enable MD044 -->
-
-In order to use the `iceoryx2` feature flags when building the `iceoryx2-ffi-c`
-crate standalone, you needs to prefix the feature with `iceoryx2/`,
-e.g. `--features iceoryx2/libc_platform.`.
```
elBoberido, line 570: Shouldn't this just be replaced by `dev_permissions`?

doc/release-notes/iceoryx2-unreleased.md
```diff
@@ -92,27 +92,39 @@
     conflicts when merging.
 -->
 
-* Example text [#1](https://github.com/eclipse-iceoryx/iceoryx2/issues/1)
+* Removed `libc_platform` feature in favor of `platform_binding` build
+  configuration which is set via environment variable
+  [#1374](https://github.com/eclipse-iceoryx/iceoryx2/issues/1374)
```
elBoberido: What do you think about getting rid of `bindgen` on Linux and using the `libc` crate by default? The Android and VxWorks platform are already using a platform abstraction based on the `libc` crate and for cross compilation we also suggest to use the `libc_platform` feature flag. Our Yocto recipes are doing it as well.

This means we have currently double the effort on Linux and maintain two platform abstractions. Since the `libc` based approach works nicely with [cross](https://github.com/cross-rs/cross), we could just use
```bash
cross test --target arm-unknown-linux-gnueabihf --release --package iceoryx2-bb-posix
```

instead of 

```bash
cross test --target arm-unknown-linux-gnueabihf --release --package iceoryx2-bb-posix --features libc_platform
```

It would probably also lead to less github issues created due to problems with cross compilation.

orecham: @elfenpiff Is there any reason to not do this?

elfenpiff: Nope, we need bindgen in general for platforms that are not supported by libc but it is completely fine to remove bindgen completely for the linux platform.

.github/workflows/build-test.yml
```diff
@@ -325,23 +330,21 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         toolchain: [stable, 1.83.0]
-        # NOTE: using 'iceoryx2/libc_platform' to trigger the feature propagation check
-        # by explicitly setting the flag only in the iceoryx2 crate
-        cargo-features-flag: ["", "--features iceoryx2/libc_platform"]
+        platform-binding: ["bindgen", "libc"]
```
dkroenke: It would be good to have the default case covered where we don't pass any setting for the platform binding.

FAQ.md
orecham: @elBoberido This was a section specifically regarding failure to build with `libc` according to the heading, do you want to keep it and generalize?

doc/bazel/README.md
```diff
@@ -107,16 +107,6 @@ when building the project.
...
-| libc_platform      | on, off      | off     |
-| logger_std         | on, off      | on      |
-<!-- markdownlint-disable-next-line MD044 -->
-| logger_posix       | on, off      | off     |
-| logger_bare_metal  | on, off      | off     |
-| logger_buffer      | on, off      | off     |
-| logger_console     | on, off      | on      |
-| logger_file        | on, off      | off     |
-| logger_log         | on, off      | off     |
-| logger_tracing     | on, off      | off     |
```
dkroenke, line 119: This PR removes the `libc_platform` flag only.

doc/release-notes/iceoryx2-unreleased.md
```diff
@@ -96,27 +92,39 @@
...
 
-   set_log_level(LogLevel::Info);
-   info!("some log message")
+    set_log_level(LogLevel::Info);
+    info!("some log message")
+    ```
 
-   // new
-   extern crate iceoryx2_bb_loggers;
+1. Building with `libc` is now default on platforms that support it
```
dkroenke, line 122: Is there a small list or code snippet where one can see which targets build with libc and which need bindgen? (Probably: iceoryx2-pal/posix/bindgen/mod.rs)

orecham: @dkroenke This snippet in `iceoryx2-pal/posix/src/lib.rs`:

```rust
#[cfg(any(
    target_os = "windows",
    target_os = "macos",
    target_os = "freebsd",
    target_os = "nto"
))]
pub(crate) mod internal {
    #![allow(non_upper_case_globals)]
    #![allow(non_camel_case_types)]
    #![allow(non_snake_case)]
    #![allow(unused)]
    #![allow(improper_ctypes)]
    #![allow(unknown_lints)]
    #![allow(unnecessary_transmutes)]
    #![allow(clippy::all)]
    #[cfg(not(bazel_build))]
    include!(concat!(env!("OUT_DIR"), "/posix_generated.rs"));

    #[cfg(bazel_build)]
    pub use iceoryx2_pal_posix_bindgen::*;

    pub const ESUCCES: u32 = 0;
}

#[cfg(any(target_os = "linux", target_os = "android"))]
pub(crate) mod internal {
    pub use libc::*;
}
```

You can look at the modules in the directory `iceoryx2-pal/posix/bindgen` to see which ones have a bindgen config.

doc/release-notes/iceoryx2-unreleased.md
orecham: Removed the option to switch between bindgen and libc. The build will now automatically choose the appropriate approach based on target.

doc/release-notes/iceoryx2-unreleased.md
dkroenke: Would be good to link this in the documentation.

orecham: @dkroenke What do you mean? It's not obvious what documentation would be useful, AFAIK this should be a hidden implementation detail.

iceoryx2-pal/posix/src/linux/mman.rs
```diff
@@ -13,40 +13,38 @@
...
-    crate::internal::shm_unlink(name)
+    libc::shm_unlink(name)
 }
 
 pub unsafe fn shm_list() -> Vec<[i8; 256]> {
     let mut result = vec![];
-    let mut search_path = iceoryx2_pal_configuration::SHARED_MEMORY_DIRECTORY.to_vec();
-    search_path.push(0);
-    let dir = opendir(search_path.as_ptr().cast());
+    let dir = opendir(c"/dev/shm/".as_ptr().cast());
```
elfenpiff: Not every linux distribution/variant has it stored under `/dev/shm` could you please therefore use the `iceoryx2_pal_configuration::SHARED_MEMORY_DIRECTORY` variable instead of hardcoding it in here.

iceoryx2-pal/os-api/build.rs
elBoberido, line 1: Can this file be removed?

iceoryx2-pal/posix/src/lib.rs
```diff
@@ -77,11 +72,16 @@ mod os;
...
 #[path = "stub/os.rs"]
 mod os;
 
-#[cfg(all(not(feature = "libc_platform"), not(target_os = "none")))]
+#[cfg(any(
+    target_os = "windows",
+    target_os = "macos",
+    target_os = "freebsd",
+    target_os = "nto"
+))]
```
elBoberido: Food for thought and smething for a follow up PR. Would it be possible to set something like a `bindgen_based` config option in `bindgen/mod.rs` and use that flag here and further down for the `libc` module import?

iceoryx2-pal/posix/BUILD.bazel
```diff
@@ -33,28 +33,15 @@ cc_library(
...
-    cc_lib = ":iceoryx2-pal-posix-c-lib",
-    bindgen_flags = [
-        "--use-core",
-        "--blocklist-type", "max_align_t",
-        # ignore warnings in the bindgen generated code
-        "--raw-line=#![allow(warnings)]",
-        "--raw-line=#![allow(clippy::all)]",
-    ],
-)
-
```
elBoberido, line 48: I think here we need to become creative since this breaks the QNX build. Let me check tomorrow how this could still be modeled to use bindgen on QNX.

FAQ.md
elBoberido: Yes, this is an easy to make mistake and it's good to have a similar error message in the FAQ, especially since the C bindings might be build by people which do not use Rust

iceoryx2-pal/posix/Cargo.toml
```diff
@@ -14,13 +14,9 @@ version = { workspace = true }
...
 [features]
-# Use https://crates.io/crates/libc for the platform abstraction. This simplifies
-# cross-compilation since bindgen is not required anymore. Unfortunately, the libc crate
-# does not offer all the functionality iceoryx2 requires on all of the supported
-# platforms. Therefore, only a subset of the supported platforms will work with this
-# feature flag.
-libc_platform = ["dep:libc"]
-std = []
+std = [
+  "libc/std"
```
elBoberido, line 19: Out of curiosity, is this needed?

iceoryx2-pal/posix/src/linux/constants.rs
```diff
@@ -15,99 +15,102 @@
...
-pub const FD_SETSIZE: usize = crate::internal::FD_SETSIZE as _;
+pub const CPU_SETSIZE: usize = libc::CPU_SETSIZE as _;
+pub const FD_SETSIZE: usize = libc::FD_SETSIZE as _;
 pub const NULL_TERMINATOR: c_char = 0;
+pub const NAME_MAX: usize = 255;
+
+#[cfg(target_os = "linux")]
 pub const USER_NAME_LENGTH: usize = 255;
+#[cfg(not(target_os = "linux"))]
+pub const USER_NAME_LENGTH: usize = 31;
```
elBoberido: This can be removed, as well as the `cfg` above

iceoryx2-pal/posix/bindgen/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    let out_path = PathBuf::from(env::var("OUT_DIR").unwrap());
+    bindings
+        .write_to_file(out_path.join("posix_generated.rs"))
+        .expect("Couldn't write bindings!");
+
+    cc::Build::new()
+        .file("src/c/socket_macros.c")
+        .compile("libsocket_macros.a");
+}
```
elBoberido, line 50: I think I use this with VxWorks also with the libc crate, since there were a few macros that were not available in the crate. But I can take care of this when the VxWorks branch is updated

iceoryx2-pal/posix/build.rs
```diff
@@ -10,80 +10,45 @@
...
+    //
+    // the target_os check refers for the target compilation which could be
+    // a different platform
+    #[cfg(any(
+        target_os = "windows",
+        target_os = "macos",
+        target_os = "freebsd",
+        target_os = "nto"
+    ))]
+    if target_os != "android" {
```
elBoberido: Shouldn*t this also include `linux`? Else it would pull in `bindgen` with the libclang dependency for cross compilation.

This would be the case when cross-compiling on e.g. macos for linux

iceoryx2-pal/posix/build.rs
```diff
@@ -94,74 +59,20 @@ fn configure_platform_override() {
...
+// #[cfg(any(...))] does not work when cross-compiling
+// when cross compiling, 'target_os' is set to the environment the build script
+// is executed; to get the actual target OS, use the cargo 'TARGET' env variable
+#[cfg(any(
+    target_os = "windows",
+    target_os = "macos",
+    target_os = "freebsd",
+    target_os = "nto"
+))]
+mod bindgen;
```
elBoberido: This would be the same as already described above

iceoryx2-pal/posix/BUILD.bazel
elBoberido: I think this can be done by using `select`
* add `load("@bazel_skylib//lib:selects.bzl", "selects")` at the top
* use this for `iceoryx2-pal-posix`

```
    deps = [
        "//iceoryx2-pal/concurrency-sync:iceoryx2-pal-concurrency-sync",
        "//iceoryx2-pal/configuration:iceoryx2-pal-configuration",
        "//iceoryx2-bb/concurrency:iceoryx2-bb-concurrency",
    ] + select({
        "@platforms//os:qnx": [":iceoryx2-pal-posix-bindgen"],
        "//conditions:default": ["@crate_index//:libc"],
    }),
```

iceoryx2-pal/posix/build.rs
```diff
@@ -10,80 +10,45 @@
...
-
-        builder = configure_builder(target_os.as_str(), builder);
-
-        if std::env::var("DOCS_RS").is_ok() {
-            builder = builder.clang_arg("-D IOX2_DOCS_RS_SUPPORT");
-        }
-
-        let bindings = builder.generate().expect("Unable to generate bindings");
+    // needed for bazel but can be empty for cargo builds
+    println!("cargo:rustc-env=BAZEL_BINDGEN_PATH_CORRECTION=");
```
elBoberido: This can also be removed. It seems it was overlooked at a previous change.

iceoryx2-pal/os-api/build.rs
orecham: This was added (to support bzlmod I think):

```rust
println!("cargo:rustc-check-cfg=cfg(bazel_build)");
```

Not sure if we can remove it.

iceoryx2-pal/posix/Cargo.toml
orecham: @elBoberido The approach I take is to utilize `std` components everywhere when building for `std`. 

I didn't look closely at exactly what changes in `libc`, but in `iceoryx2` for example, building for `no_std` often involves replacing some components with spin-based alternatives which may have an effect on performance (in either direction).

We can decide to use `no_std` flavors of components for both builds but some thinking/analysis is needed, which I would prefer to leave for later.

iceoryx2-pal/posix/build.rs
orecham: @elBoberido Would it be needed for bazel builds of QNX?

iceoryx2-pal/posix/build.rs
orecham: @elBoberido A constant would not work here like it would in `lib.rs` because these `cfg` switches in `build.rs` are for the host platform not the target.

iceoryx2-pal/posix/BUILD.bazel
orecham: @elBoberido Added, although we were not yet supporting bazel builds for QNX. No harm in keeping it there though I guess.

orecham:
@elBoberido @elfenpiff 

I have addressed all issues and manually confirmed the Cargo build for QNX still works on QNX 8.

Bazel was not yet supported for QNX builds and has other issues unrelated to this PR. They can be addressed in a follow-up if we do want to support this, but don't think it is worth the effort.

iceoryx2-pal/posix/build.rs
orecham: I removed it. If we need it for QNX it can be added if we decide to add Bazel support for QNX.

iceoryx2-pal/posix/build.rs
```diff
@@ -10,80 +10,49 @@
 //
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-#[cfg(feature = "libc_platform")]
-fn main() {
-    configure_platform_override();
-}
+const ENV_PLATFORM_PATH: &str = "IOX2_CUSTOM_POSIX_PLATFORM_PATH";
+const BINDGEN_PLATFORMS: &[&str] = &["windows", "macos", "freebsd", "nto"];
+const LIBC_PLATFORMS: &[&str] = &["android", "linux", "vxworks"];
```
orecham, line 15: @dkroenke FYI added these

iceoryx2-pal/posix/Cargo.toml
orecham: Having said that, realized I was not properly doing what I intended.

See: https://github.com/eclipse-iceoryx/iceoryx2/pull/1378/changes/d48ee438d808be67a274fed783ad104a11c7fc2f

iceoryx2-pal/os-api/build.rs
elBoberido: I think it was required due to bindgen. Since this is Linux only, I think we do not need it anymore

iceoryx2-pal/posix/BUILD.bazel
elBoberido: No, but it seems some people managed to make it work with QNX

iceoryx2-pal/posix/build.rs
elBoberido: Even for bazel on QNX this is not required anymore. It is a remnant of the approach before the bazel integration was changed to bzlmod.

iceoryx2-pal/posix/src/linux/constants.rs
```diff
@@ -15,99 +15,98 @@
 
 use crate::posix::types::*;
 
-pub const CPU_SETSIZE: usize = crate::internal::__CPU_SETSIZE as _;
-pub const FD_SETSIZE: usize = crate::internal::FD_SETSIZE as _;
+pub const CPU_SETSIZE: usize = libc::CPU_SETSIZE as _;
+pub const FD_SETSIZE: usize = libc::FD_SETSIZE as _;
 pub const NULL_TERMINATOR: c_char = 0;
+pub const NAME_MAX: usize = 255;
```
elBoberido, line 21: It seems this constant was added to the latest libc crate version
```suggestion
pub const NAME_MAX: usize = libc::NAME_MAX as _;;
```

iceoryx2-pal/posix/src/linux/constants.rs
```diff
@@ -128,244 +127,236 @@ pub const RLIMIT_RTTIME: __rlim_t = 15;
 pub const RLIMIT_NLIMITS: __rlim_t = 16;
```
elBoberido, line 127: It seems these constants are now also added to the libc crate. Up to you if you want to use those. The only issue is with RLIMIT_NLIMITS, which is deprecated, since it seems to not be stable across OS versions.

iceoryx2-pal/posix/src/linux/constants.rs
```diff
@@ -128,244 +127,236 @@ pub const RLIMIT_RTTIME: __rlim_t = 15;
...
+pub const SIGPROF: int = libc::SIGPROF as _;
+pub const SIGSYS: int = libc::SIGSYS as _;
+pub const SIGTRAP: int = libc::SIGTRAP as _;
+pub const SIGURG: int = libc::SIGURG as _;
+pub const SIGVTALRM: int = libc::SIGVTALRM as _;
+pub const SIGXCPU: int = libc::SIGXCPU as _;
+pub const SIGXFSZ: int = libc::SIGXFSZ as _;
 pub const SIG_ERR: sighandler_t = sighandler_t::MAX;
 pub const SIG_DFL: int = 0;
 pub const SIG_IGN: int = 1;
```
elBoberido, line 169: These are now also available via the libc crate

iceoryx2-pal/posix/BUILD.bazel
```diff
@@ -50,13 +51,28 @@ rust_bindgen_library(
...
     srcs = glob(["src/**/*.rs"]),
-    rustc_flags = ["--cfg=bazel_build"],
+    rustc_flags = select({
+        "@platforms//os:linux":   ["--cfg=bazel_build", "--cfg=platform_binding=\"libc\""],
+        "@platforms//os:android": ["--cfg=bazel_build", "--cfg=platform_binding=\"libc\""],
+        "@platforms//os:vxworks": ["--cfg=bazel_build", "--cfg=platform_binding=\"libc\""],
+        "@platforms//os:macos":   ["--cfg=bazel_build", "--cfg=platform_binding=\"bindgen\""],
+        "@platforms//os:windows": ["--cfg=bazel_build", "--cfg=platform_binding=\"bindgen\""],
+        "@platforms//os:freebsd": ["--cfg=bazel_build", "--cfg=platform_binding=\"bindgen\""],
+        "@platforms//os:qnx":     ["--cfg=bazel_build", "--cfg=platform_binding=\"bindgen\""],
```
elBoberido, line 61: How about defaulting to `libc`? This leads to less effort when a new platform based on libc is added.

iceoryx2-pal/posix/BUILD.bazel
```diff
@@ -50,13 +51,28 @@ rust_bindgen_library(
...
         "//iceoryx2-bb/concurrency:iceoryx2-bb-concurrency",
-    ],
+    ] + select({
+        "@platforms//os:linux":   ["@crate_index//:libc"],
+        "@platforms//os:android": ["@crate_index//:libc"],
+        "@platforms//os:vxworks": ["@crate_index//:libc"],
+        "@platforms//os:macos":   [":iceoryx2-pal-posix-bindgen"],
+        "@platforms//os:windows": [":iceoryx2-pal-posix-bindgen"],
+        "@platforms//os:freebsd": [":iceoryx2-pal-posix-bindgen"],
+        "@platforms//os:qnx":     [":iceoryx2-pal-posix-bindgen"],
```
elBoberido, line 74: Same here.

iceoryx2-pal/posix/build.rs
```diff
@@ -94,74 +63,32 @@ fn configure_platform_override() {
...
-        _ => panic!("Unsupported target OS: {}", target_os),
+fn configure_platform_binding(target_os: &str) {
+    println!("cargo:rustc-check-cfg=cfg(platform_binding, values(\"libc\", \"bindgen\"))");
+    if BINDGEN_PLATFORMS.contains(&target_os) {
+        println!("cargo:rustc-cfg=platform_binding=\"bindgen\"");
+    } else if LIBC_PLATFORMS.contains(&target_os) {
+        println!("cargo:rustc-cfg=platform_binding=\"libc\"");
+    } else {
+        panic!("Unsupported target OS: {}", target_os);
     }
```
elBoberido, line 81: Food for thought. To simplify the process to add new platforms. What do you think about using only a flag for `bindgen` and `#[cfg(not(bindgen_foo))]`. With this, less places need to be touched when a new platform which is based on libc is added.

Either way, this is fine for now.

iceoryx2-pal/os-api/src/lib.rs
elBoberido, line 1: The `os-ap/src/c` folder could be delelted

iceoryx2-pal/posix/build.rs
```diff
@@ -10,80 +10,49 @@
...
+    // the target_os check refers for the target compilation which could be
+    // a different platform
+    #[cfg(any(
+        target_os = "linux",
+        target_os = "windows",
+        target_os = "macos",
+        target_os = "freebsd",
+        target_os = "nto"
+    ))]
+    if BINDGEN_PLATFORMS.contains(&target_os.as_str()) {
```
elBoberido, line 47: I'm not sure anymore why we had this complex setup, but it seems the `cfg` could be removed. Further investigation is needed though and I wouldn't change it in this PR.

iceoryx2-pal/posix/src/linux/constants.rs
orecham: https://github.com/eclipse-iceoryx/iceoryx2/issues/1388

iceoryx2-pal/posix/src/linux/constants.rs
orecham: https://github.com/eclipse-iceoryx/iceoryx2/issues/1388

iceoryx2-pal/posix/src/linux/constants.rs
orecham: https://github.com/eclipse-iceoryx/iceoryx2/issues/1388

iceoryx2-pal/posix/build.rs
orecham: @elBoberido I actually like the current approach. The suggested change would set up the semantic that anything that is not `bindgen` must use `libc`. The current approach is more self-documenting IMO.

If you feel strongly about this, I would not object to it being changed later.

iceoryx2-pal/posix/build.rs
orecham: From what I recall, the reason was to allow builds on android, which cannot build `bindgen`. The reason why we wanted to do that is unclear.

I will leave it to another PR where this can be investigated in more detail.

iceoryx2-pal/posix/BUILD.bazel
orecham: I would leave these for another PR.

iceoryx2-pal/posix/build.rs
elBoberido: No problem. We can keep it as is

iceoryx2-pal/posix/build.rs
elBoberido: This was even from before the Android MVP ... I have to check the history of the file.

elBoberido: You were right. It was introduced with the Android MVP. But the logic behind that was flawed since bindgen was pulled in anyway. It also needs to be put behind a cfg in the Cargo.toml file. Well, something for our future selves if we ever encounter any problem.

PR #1391, opened 2026-03-05
[#1368] Support for free-threaded python 3.14

### LINKED ISSUE #1368: Add Support for Python Free-Threading
## Brief Feature Description

See: [https://github.com/eclipse-iceoryx/iceoryx2/pull/1364](https://github.com/eclipse-iceoryx/iceoryx2/pull/1364) for context.

Starting with Python 3.13, CPython introduces an experimental free-threaded build (PEP 703) that removes the Global Interpreter Lock (GIL). To support this build configuration, PyO3 modules must explicitly declare that they do not rely on the GIL.

This can be done by annotating the module with:

```rust
#[pymodule(gil_used = false)]
```

However, this requires that:

* No APIs that depend on `Python<'py>` or otherwise require the GIL are used.
* All existing GIL-dependent code paths are refactored accordingly.
* The module is fully safe to execute without the GIL.

The following files currently contain GIL-bound code that would need to be reviewed and potentially refactored:

* `iceoryx2-ffi/python/src/type_storage.rs` (L32)
* `iceoryx2-ffi/python/src/service_builder_blackboard.rs` (L120)
* `iceoryx2-ffi/python/src/service_builder_blackboard.rs` (L137)

## Compatibility & Packaging Considerations

Supporting free-threaded Python introduces additional packaging constraints:

* Free-threaded Python builds do not support `abi3`
* If we want to support both:

  * Regular CPython (with GIL)
  * Free-threaded CPython builds

We may need to:

* Drop `abi3`
* Build version-specific wheels
* Make the Python ABI configuration adjustable in `iceoryx2-ffi/python/Cargo.toml`
* Potentially introduce Cargo feature flags to differentiate build targets

Additionally, we must ensure continued compatibility with users still running Python 3.8+.

berkgercek:
This PR makes a few small changes after upgrading to PyO3 0.28.2, which has interface changes that make logic clearer about how the GIL in Python is handled. There should be no change to functionality from a user standpoint, and the Python tests pass locally after changing.

## Notes for Reviewer

- Upgrades `pyo3` to `0.28.2` from `0.25.0`
- Changes `PyObject` types to `Py<PyAny>` which was the underlying type for the now-deprecated `PyObject`
- Uses `Python::attach` instead of `Python::with_gil` [as per the rework of the PyO3 interface](https://github.com/PyO3/pyo3/issues/3987).
- Specifies `from_py_object` and `skip_from_py_object` for classes to adapt to the [phase-out of blanket implementation](https://pyo3.rs/main/conversions/traits#-phase-out-of-frompyobject-blanket-implementation-for-cloneable-pyclasses-). Skipped for classes only returned or passed by reference.
- Adds a `-i` flag to poetry which is passed through to Maturin for building for a specific interpreter version. Defaults to empty (no specific interpreter)
- Adds an additional job to release workflow which builds for `3.14t` with a matrix for future versions of Python with free-threading

## References

Closes #1368

Relates to #1364

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -84,7 +84,7 @@ jobs:
           retention-days: 7
 
   prepare-python-dist:
-    needs: [build-python-wheels]
+    needs: [build-python-wheels, build-python-freethreaded-wheels]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # version v6.0.1


elfenpiff (approved):

elfenpiff:
@berkgercek Thank you for the contribution!

PR #1497, opened 2026-03-29
[#1496] Use compiler builtins for 'SourceLocation' to get location at invokation point

### LINKED ISSUE #1496: Implement C++20 'source_location' based on compiler source location builtins
## Brief feature description

C++20 introduced `source_location`, which can be used to get a source location at the invocation point, without the usage of macros and has the benefit of obtaining this information at the invocation point.

## Detailed information

iceoryx2 has a `SourceLocation` implementation but it is based on the `__FILE__`, `__LINE__` and `__FUNCTION__` and can only be used ergonomically by hiding them in another macro. We do this with the log and assert macros and that works pretty well. Unfortunately, at places like `Optional::value` or `Expected::value`, this leads to the situation that we get the file, line and function information of the `value` function of the `Optional` or `Expected`, instead of where the `value` function is called. This makes the information useless for debugging.

By using compiler source location builtins, we can implement the C++20 `source_location` behavior with C++14. For this, we can use `__builtin_FILE()`, `__builtin_LINE()` and `__builtin_FUNCTION()`, which are supported by GCC, CLang and MSVC.

See also https://clang.llvm.org/docs/LanguageExtensions.html#source-location-builtins.

## Tasks

* [x] use compiler source location builtins in `iox2::legacy::er::SourceLocation` and move it to `iox2::bb::detail::SourceLocation`
* [x] replace `IOX2_CURRENT_SOURCE_LOCATION` with `iox2::bb::detail::SourceLocation::current()`
* [x] add `SourceLocation = SourceLocation::current()` to `Optional` and `Expected` functions like `value` and `error`

elBoberido:
## Notes for Reviewer

This PR re-implements the C++20 `source_location` with compiler builtins, in order to be able to use it with C++14.

With the introduction of the new capabilities, the `Expected::value()`, `Expected::error()` and `Optional::value()` methods were updated in order to show the line of the invocation point where the invalid access was triggered, instead of the line where the check is done. This will help a lot at debugging since the user now sees where the core of the issue lies, instead of having an error message with a line from within the `Expected` or `Optional`. 

The new `SourceLocation` was also added to the logging infrastructure.

## PR Reviewer Reminders

* Public API is documented

## References

Closes #1496

### DIFF
--- a/iceoryx2-bb/cxx/tests/src/source_location_tests.cpp
+++ b/iceoryx2-bb/cxx/tests/src/source_location_tests.cpp
@@ -1,5 +1,4 @@
-// Copyright (c) 2020 - 2023 by Apex.AI Inc. All rights reserved.
-// Copyright (c) 2025 Contributors to the Eclipse Foundation
+// Copyright (c) 2026 Contributors to the Eclipse Foundation
 //
 // See the NOTICE file(s) distributed with this work for additional
 // information regarding copyright ownership.


elBoberido:
@ComicSansMS would be great if you also could have a look at the PR.

There is still a case where we are not able to show the actual invocation point. This is when the `operator->` or `operator*` is used. We could talk about removing this operators and requiring the users to use the `value()` method but I'm not sure if the users would be that happy about the usability. On the other hand, it would maybe make the user more aware of a potential violation instead of a simple `*` or `->`. What do you think?

cc @elfenpiff

iceoryx2-bb/cxx/include/iox2/legacy/detail/log/logstream.inl
```diff
@@ -86,13 +86,13 @@ inline constexpr LogRaw raw(const void* const data, const uint64_t size) noexcep
 // AXIVION Next Construct AutosarC++19_03-A3.9.1 : See at declaration in header
```
ComicSansMS: This violation should go away with the change.

iceoryx2-bb/cxx/include/iox2/legacy/detail/log/logstream.inl
```diff
@@ -86,13 +86,13 @@ inline constexpr LogRaw raw(const void* const data, const uint64_t size) noexcep
...
 // NOLINTNEXTLINE(readability-function-size)
 inline LogStream::LogStream(
-    Logger& logger, const char* file, const int line, const char* function, LogLevel logLevel) noexcept
+    Logger& logger, const char* file, const uint32_t line, const char* function, LogLevel logLevel) noexcept
     : m_logger(logger) {
     m_logger.createLogMessageHeader(file, line, function, logLevel);
 }
 
 // AXIVION Next Construct AutosarC++19_03-A3.9.1 : See at declaration in header
```
ComicSansMS: Same as line 86.

ComicSansMS:
> There is still a case where we are not able to show the actual invocation point. This is when the `operator->` or `operator*` is used. We could talk about removing this operators and requiring the users to use the `value()` method but I'm not sure if the users would be that happy about the usability. On the other hand, it would maybe make the user more aware of a potential violation instead of a simple `*` or `->`. What do you think?

Removing `*` and `->` from `Optional` would mean that the type can no longer be used as a drop-in replacement for `std::optional`. This was originally a design goal of `Optional`.

The disadvantage with keeping the operators as is is slightly worse diagnostics in uses of those operators that are pretty much always programming errors. I would keep them to not needlessly break drop-in compatibility, but I don't feel strongly about this.

iceoryx2-bb/cxx/include/iox2/legacy/detail/log/logstream.inl
elBoberido: done

iceoryx2-bb/cxx/include/iox2/legacy/detail/log/logstream.inl
elBoberido: done

elBoberido:
> Looks good. Are we still missing unit tests for the newly added source location type, or are they located somewhere else?

Thanks for the hint. They were indeed missing. I added new ones.

> Removing `*` and `->` from `Optional` would mean that the type can no longer be used as a drop-in replacement for `std::optional`. This was originally a design goal of `Optional`.

I was unsure how far the compatibility needs to go, since we also do not support the monadic operations, yet.

> The disadvantage with keeping the operators as is is slightly worse diagnostics in uses of those operators that are pretty much always programming errors. I would keep them to not needlessly break drop-in compatibility, but I don't feel strongly about this.

Also no strong feelings about this. I hoped a little bit to come up with a secret functionality only C++ committee members know of and we could add default parameters to these functions ;)

Will keep it as is for now and if necessary, we can change it later.

iceoryx2-bb/cxx/tests/src/source_location_tests.cpp
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) 2020 - 2023 by Apex.AI Inc. All rights reserved.
```
FerdinandSpitzschnueffler: Is the copyright header correct? This seems to be a completely new file with new tests.

elBoberido: Oh, copy&paste error. Fixed

PR #1542, opened 2026-04-13
fix(event): validate lifecycle event IDs against event_id_max_value at create and open time

### LINKED ISSUE #1488: event service builder can specify invalid ids
The event builder can define event ids that may become invalid when they are restricted with `event_id_max_value()`.

Example:

```rust
 let event = node
     .service_builder(&"MyEventName".try_into()?)
     .event()
     .notifier_dead_event(EventId::new(50)) 
     .notifier_dropped_event(EventId::new(54))
     .notifier_created_event(EventId::new(53))
// max event id is 10, despite that, an event id of 50+ could be defined.
     .event_id_max_value(10)
     .open_or_create()?;
```

Solution: Check the defined event id values for consistency when creating the service. If they are contradicting each other, emit a failure.

aki1770-del:
## Notes for Reviewer

This fixes silent acceptance of out-of-range lifecycle event IDs (notifier_created_event,
notifier_dropped_event, notifier_dead_event) in the EventService builder.

Previously, an invalid event ID (exceeding event_id_max_value) was accepted at
service creation time and only rejected later when a Notifier tried to emit it via
notify_with_custom_event_id — far from the misconfiguration site.

Two guard points added:
- **create_impl**: validates lifecycle event IDs before persisting static config.
  Returns new EventCreateError::EventIdExceedsMaxSupportedValue on violation.
- **verify_service_configuration**: validates stored lifecycle event IDs on open path.
  Returns existing DoesNotSupportRequestedMaxEventId, catching any service persisted
  in an invalid state before this fix.

New conformance test: create_fails_when_lifecycle_event_id_exceeds_max_event_id

[testing]: https://github.com/eclipse-iceoryx/iceoryx2/blob/master/doc/website/concepts/best-practice-for-testing.md

## PR Reviewer Reminders

* Public API is documented

## References

Closes #1488

*AI-assisted — authored with Claude, reviewed by Komada.*

### DIFF
--- a/iceoryx2/conformance-tests/src/service_event.rs
+++ b/iceoryx2/conformance-tests/src/service_event.rs
@@ -1298,6 +1298,63 @@ pub mod service_event {
             format!("{}", EventCreateError::InsufficientPermissions), eq "EventCreateError::InsufficientPermissions");
         assert_that!(
             format!("{}", EventCreateError::IsBeingCreatedByAnotherInstance), eq "EventCreateError::IsBeingCreatedByAnotherInstance");
+        assert_that!(
+            format!("{}", EventCreateError::EventIdExceedsMaxSupportedValue), eq "EventCreateError::EventIdExceedsMaxSupportedValue");
+    }
+
+    // Regression test for #1488: lifecycle event IDs exceeding event_id_max_value must be
+    // rejected at service creation time and detected at open time, rather than silently
+    // accepted and only failing at notify_with_custom_event_id call time.
+    #[conformance_test]
+    pub fn create_fails_when_lifecycle_event_id_exceeds_max_event_id<Sut: Service>() {
+        let config = generate_isolated_config();
+        let node = NodeBuilder::new().config(&config).create::<Sut>().unwrap();
+        const MAX: usize = 5;
+
+        // notifier_created_event exceeds max
+        let service_name = generate_service_name();
+        let sut = node
+            .service_builder(&service_name)
+            .event()
+            .event_id_max_value(MAX)
+            .notifier_created_event(EventId::new(MAX + 1))
+            .create();
+        assert_that!(sut, is_err);
+        assert_that!(sut.err().unwrap(), eq EventCreateError::EventIdExceedsMaxSupportedValue);
+
+        // notifier_dropped_event exceeds max
+        let service_name = generate_service_name();
+        let sut = node
+            .service_builder(&service_name)
+            .event()
+            .event_id_max_value(MAX)
+            .notifier_dropped_event(EventId::new(MAX + 1))
+            .create();
+        assert_that!(sut, is_err);
+        assert_that!(sut.err().unwrap(), eq EventCreateError::EventIdExceedsMaxSupportedValue);
+
+        // notifier_dead_event exceeds max
+        let service_name = generate_service_name();
+        let sut = node
+            .service_builder(&service_name)
+            .event()
+            .event_id_max_value(MAX)
+            .notifier_dead_event(EventId::new(MAX + 1))
+            .create();
+        assert_that!(sut, is_err);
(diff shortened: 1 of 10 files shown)


iceoryx2/src/service/builder/event.rs
```diff
@@ -673,6 +695,34 @@ impl<ServiceType: service::Service> Builder<ServiceType> {
                 msg, existing_settings.deadline, required_settings.deadline);
         }
 
+        // Verify that the existing service's lifecycle event IDs are within its own
+        // event_id_max_value bounds.  A service whose lifecycle events exceed its own max
+        // would cause every Notifier creation to fail with EventIdOutOfBounds at emit time —
+        // a silent accept that only surfaces later.  Catch it here instead.
+        let max_event_id = existing_settings.event_id_max_value;
```
elfenpiff: This case can never happen. When exactly this condition is checked in `create_impl`, which causes an error, how should a service with this contract violation be created?

It is possible that I am wrong here, but I also do not see a test case covering the `EventOpenError::DoesNotSupportRequestedMaxEventId` error case.

iceoryx2/conformance-tests/src/service_event.rs
```diff
@@ -1298,6 +1298,63 @@ pub mod service_event {
...
+    // Regression test for #1488: lifecycle event IDs exceeding event_id_max_value must be
+    // rejected at service creation time and detected at open time, rather than silently
+    // accepted and only failing at notify_with_custom_event_id call time.
+    #[conformance_test]
+    pub fn create_fails_when_lifecycle_event_id_exceeds_max_event_id<Sut: Service>() {
+        let config = generate_isolated_config();
+        let node = NodeBuilder::new().config(&config).create::<Sut>().unwrap();
+        const MAX: usize = 5;
+
+        // notifier_created_event exceeds max
```
elfenpiff: In our test philosophie we verify with every test just one aspect of the code. Could you please create 4 separate test cases out of this, so that we also have a clean setup and can exclude that a previous test run affects the next one.

iceoryx2/src/service/builder/event.rs
aki1770-del: > This case can never happen. When exactly this condition is checked in `create_impl`, which causes an error, how should a service with this contract violation be created?
> 
> It is possible that I am wrong here, but I also do not see a test case covering the `EventOpenError::DoesNotSupportRequestedMaxEventId` error case.

You're right — with the create check in place, the open path can't observe a service that violates the contract. Removed the open-time block; only the create-time validation remains.

iceoryx2/conformance-tests/src/service_event.rs
aki1770-del: > In our test philosophie we verify with every test just one aspect of the code. Could you please create 4 separate test cases out of this, so that we also have a clean setup and can exclude that a previous test run affects the next one.

Done — split into four separate #[conformance_test] functions (create_fails_when_notifier_created_event_exceeds_max_event_id, ...dropped..., ...dead..., and create_succeeds_when_lifecycle_event_ids_within_max_event_id). Each gets its own generate_isolated_config() and node, so no cross-test state.

aki1770-del:
Rebased onto current main; this sat conflicting since your review in April, and that is on us. Your two points are carried from the 15 April push: no open-time check, four independent conformance tests. Rebase changes: `From<EventCreateError> for ServiceCreateError` is now an exhaustive match, so the new variant maps to `InternalFailure`; and two existing tests (C++ `service_settings_are_applied`, Python `test_service_builder_configuration_works`) plus two doc examples set lifecycle ids above the `event_id_max_value` they declare, now rejected, so they are brought into range. Locally: clippy clean, conformance tests and doctests green, failure cases fail without the check.

elfenpiff (approved):

PR #1543, opened 2026-04-13
fix(python): divide __slice_len by sizeof(T) to get element count in payload()

### LINKED ISSUE #1533: Python bindings report Slice payload size in bytes instead of number of elements
**iceoryx2 version:**
0.8.1

`publisher.py`
```python
import ctypes
import iceoryx2

print("Launching Publisher")

node = (
    iceoryx2.NodeBuilder.new()
    .name(iceoryx2.NodeName.new("SizeBugRepro"))
    .create(iceoryx2.ServiceType.Ipc)
)

service = (
    node.service_builder(iceoryx2.ServiceName.new("SlicePub"))
    .publish_subscribe(iceoryx2.Slice[ctypes.c_uint64])
    .open_or_create()
)

publisher = (
    service.publisher_builder()
    .initial_max_slice_len(1024)
    .allocation_strategy(iceoryx2.AllocationStrategy.PowerOfTwo)
    .create()
)

counter = 0
while True:
    counter += 1
    node.wait(iceoryx2.Duration.from_secs(1))
    sample = publisher.loan_slice_uninit(5)
    for i in range(0, 5):
        sample.payload()[i] = i + counter
    sample.assume_init().send()
    print(f"Sample #{counter} sent")
```

`subscriber.py`
```python
import ctypes
import iceoryx2

print("Launching Subscriber")

node = (
    iceoryx2.NodeBuilder.new()
    .name(iceoryx2.NodeName.new("SizeBugRepro"))
    .create(iceoryx2.ServiceType.Ipc)
)

service = (
    node.service_builder(iceoryx2.ServiceName.new("SlicePub"))
    .publish_subscribe(iceoryx2.Slice[ctypes.c_uint64])
    .open_or_create()
)

subscriber = (
    service.subscriber_builder()
    .create()
)

while True:
    node.wait(iceoryx2.Duration.from_secs(1))
    sample = subscriber.receive()
    if sample is not None:
        print("Received sample")
        payload = sample.payload()
        print(f"Payload size is {payload.len()}")
```

When the publisher loans a sample of size 5, the subscriber should receive a payload slice of that sime size.
What happens instead is that the sample on the publisher size has a size of 40, so the size in bytes of the payload, instead of the number of elements in the slice.

aki1770-del:
## Notes for Reviewer

The Rust FFI layer's `__slice_len` property returns the raw **byte length** of the
payload buffer. The Python `payload()` function was passing this byte count directly
to `Slice()` as the element count.

For single-byte element types (`c_uint8`) this happened to be correct. For any
element type wider than one byte (`c_uint32`, `c_int64`, a struct), `Slice.len()`
would report `N * sizeof(T)` instead of `N`.

Fix: one line in `payload()` — divide `__slice_len` by `ctypes.sizeof(contained_type)`
before passing to `Slice()`.

New regression test: `test_slice_payload_element_count_is_correct_for_multi_byte_types`
publishes a `Slice[c_uint32]` of 3 elements and asserts `Slice.len() == 3` (not 12).

[testing]: https://github.com/eclipse-iceoryx/iceoryx2/blob/master/doc/website/concepts/best-practice-for-testing.md

## PR Reviewer Reminders

* Public API is documented

## References

Closes #1533

*AI-assisted — authored with Claude, reviewed by Komada.*

### DIFF
--- a/iceoryx2-ffi/python/python-src/iceoryx2/publish_subscribe_extensions.py
+++ b/iceoryx2-ffi/python/python-src/iceoryx2/publish_subscribe_extensions.py
@@ -27,7 +27,10 @@ def payload(self: Any) -> Any:
     assert self.__payload_type_details is not None
     if get_origin(self.__payload_type_details) is Slice:
         (contained_type,) = get_args(self.__payload_type_details)
-        return Slice(self.payload_ptr, self.__slice_len, contained_type)
+        # __slice_len is the raw payload size in bytes; divide by the element
+        # size to obtain the element count expected by Slice. (fix for #1533)
+        number_of_elements = self.__slice_len // ctypes.sizeof(contained_type)
+        return Slice(self.payload_ptr, number_of_elements, contained_type)
 
     return ctypes.cast(self.payload_ptr, ctypes.POINTER(self.__payload_type_details))
 
--- a/iceoryx2-ffi/python/python-src/iceoryx2/publish_subscribe_extensions.py
+++ b/iceoryx2-ffi/python/python-src/iceoryx2/publish_subscribe_extensions.py
@@ -27,9 +27,9 @@ def payload(self: Any) -> Any:
     assert self.__payload_type_details is not None
     if get_origin(self.__payload_type_details) is Slice:
         (contained_type,) = get_args(self.__payload_type_details)
-        # __slice_len is the raw payload size in bytes; divide by the element
-        # size to obtain the element count expected by Slice. (fix for #1533)
-        number_of_elements = self.__slice_len // ctypes.sizeof(contained_type)
+        number_of_elements = self.__payload_size_in_bytes // ctypes.sizeof(
+            contained_type
+        )
         return Slice(self.payload_ptr, number_of_elements, contained_type)
 
     return ctypes.cast(self.payload_ptr, ctypes.POINTER(self.__payload_type_details))
--- a/iceoryx2-ffi/python/python-src/iceoryx2/publish_subscribe_extensions.py
+++ b/iceoryx2-ffi/python/python-src/iceoryx2/publish_subscribe_extensions.py
@@ -27,10 +27,7 @@ def payload(self: Any) -> Any:
     assert self.__payload_type_details is not None
     if get_origin(self.__payload_type_details) is Slice:
         (contained_type,) = get_args(self.__payload_type_details)
-        number_of_elements = self.__payload_size_in_bytes // ctypes.sizeof(
-            contained_type
-        )
-        return Slice(self.payload_ptr, number_of_elements, contained_type)
+        return Slice(self.payload_ptr, self.__slice_len, contained_type)
 
     return ctypes.cast(self.payload_ptr, ctypes.POINTER(self.__payload_type_details))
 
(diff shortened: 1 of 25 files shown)


iceoryx2-ffi/python/python-src/iceoryx2/publish_subscribe_extensions.py
```diff
@@ -27,7 +27,10 @@ def payload(self: Any) -> Any:
     assert self.__payload_type_details is not None
     if get_origin(self.__payload_type_details) is Slice:
         (contained_type,) = get_args(self.__payload_type_details)
-        return Slice(self.payload_ptr, self.__slice_len, contained_type)
+        # __slice_len is the raw payload size in bytes; divide by the element
+        # size to obtain the element count expected by Slice. (fix for #1533)
+        number_of_elements = self.__slice_len // ctypes.sizeof(contained_type)
```
elfenpiff: This fixes the symptoms but not the actual bug. The problem is that - only in python - the `number_of_elements` contain wrongly the number of bytes and not how many `contained_type`s are contained in that slice.

aki1770-del: Good catch — renamed the Rust FFI getter from `__slice_len` to `__payload_size_in_bytes` across all 9 binding files (`Sample`, `SampleMut`, `SampleMutUninit`, `Request*`, `Response*`, `ActiveRequest`). The unit is now explicit at the source; the element-count division stays in the Python layer where `contained_type` is known via the type annotations.

While renaming I noticed `request_response_extensions.py` had the identical byte-vs-element bug at `request_payload()` and `response_payload()` — not covered by the original fix. Fixed those too in the same commit (1e909b4).

elfenpiff: But now the whole API is inconsistent to the C++, C and Rust API where this is always the len.

This creates a lot of problems when using this API across language boundaries. In one language the slice size is measured in bytes in the other in elements. Furthermore, such an API doesn't make sense for a high level abstraction.

A list or array len is in python also measured in number of elements and not number of bytes.

aki1770-del:
Option 2 implemented in 91e3c21. The payload element size is read once from each port factory's static config and threaded through:

- pub-sub: `PortFactoryPublishSubscribe` → `PortFactoryPublisher/Subscriber` → `Publisher/Subscriber` → `Sample/SampleMut/SampleMutUninit`
- req-resp: `PortFactoryRequestResponse` → `PortFactoryClient/Server` → `Client/Server` → `RequestMut/RequestMutUninit/ActiveRequest` → `PendingResponse` → `Response/ResponseMut/ResponseMutUninit`

`__slice_len` is back to a zero-arg getter returning element count, consistent with the C/C++/Rust APIs and Python's `len()`. Python extensions call `self.__slice_len` directly with no division logic. The parallel bug in `request_response_extensions.py` (`request_payload` / `response_payload`) is fixed as a side effect.

aki1770-del:
You refused the byte-based API on 15 April; 96 minutes later we pushed the 23-file element-size threading, which was not what you asked for either, then went silent. This replaces it with what you asked for: `__slice_len` keeps its name and returns the element count, read from the sample header as the C API does. Nine binding files; test and changelog carried. The title still names the old approach. Python suite 535 passed; the test fails without it.

Issue #1573, opened 2026-04-20
Enhance degradation handler


elBoberido:
## Brief feature description

The degradation handler has a limited scope and is only called when a connection is corrupted. The scope should be extended to be able to hook into the send call and execute custom code when a sample cannot be delivered.

## Detailed information

There are scenarios where a custom action needs to be performed when a sample cannot be delivered. Currently, the only two options are to either drop the sample or block until there is enough space in the receiver buffer to deliver the sample.

With the degradation handler (currently named degradation callback), there is already some infrastructure to hook into the system. But currently, the degradation handler is only called on a broken connection and is also not available in the language bindings.

## Tasks

* [ ] add degradation handler to the send call
* [ ] add degradation handler to all language bindings

## Related issues

* #1131 
* #1511

PR #1574, opened 2026-04-20
[DO NOT REVIEW] [WIP] [#1573] Enhance Degradation Handler

elBoberido:

## PR Reviewer Reminders

* Public API is documented

## References

Closes #


iceoryx2-cal/src/zero_copy_connection/common.rs
```diff
@@ -761,6 +789,15 @@ pub mod details {
...
                         "{msg} {ptr:?} via channel {channel_id:?} since the channel is closed.");
                 }
+
+                if do_fail {
+                    fail!(from self, with ZeroCopySendError::ReceiveBufferFull,
+                          "{msg} {ptr:?} via channel {channel_id:?} since the receive buffer is full.");
+                }
+
+                if discare_sample {
+                    return Ok(Some(ptr));
```
elBoberido: @elfenpiff @budrus it would be trivial to give the user to option to discard the oldest sample instead of the newest. With this, the user would be able to also execute custom code when safe overflowing is enabled.

Could be done later in a separate issue. Nothing for the current degradation handler enhancement.

iceoryx2-cal/src/zero_copy_connection/mod.rs
```diff
@@ -129,6 +131,33 @@ impl core::fmt::Display for ZeroCopyReleaseError {
 
 impl core::error::Error for ZeroCopyReleaseError {}
 
+pub enum UnableToDeliverAction {
+    Fail,
+    Block,
+    Retry,
+    DiscardLatestSample,
```
elBoberido, line 139: @elfenpiff I would align this with the `UnableToDeliverStrategy` and `DegradationAction` and have the following options

```rust
enum UnableToDeliverAction {
    DiscardSample,
    Retry,
    Fail,
}

enum UnableToDeliverStrategy {
    DiscardSample, // unchanged
    Retry, // what `Block` does right now
    Fail, // similar to DiscardSample but in addition abort with an error code
    DefereToHandler, // the user decides with  DegradationAction
}

enum DegradationAction {
    Default, // -> leads to UnableToDeliverStrategy::Retry
    Discard (or Ignore), // -> leads to UnableToDeliverStrategy::DiscardSample
    Retry, // -> leads to UnableToDeliverStrategy::Retry
    Warn, // -> leads to UnableToDeliverStrategy::DiscardSample and print warning
    Fail, // -> leads to UnableToDeliverStrategy::DiscardSample and error code
}
```

What do you think?

elfenpiff: * Some thoughts in my hand. Can you add an option, BlockWithTimeout to the ZeroCopyConnectiong thingie?

* Where does the UnableToDeliverAction come into play?

* I would keep UnableToDeliverStrategy::Block. Whenever we have a blocking call we add the explicit block prefix. I would keep it here. Maybe rename it to BlockUntilDelivered. And add here the Option BlockWithTimeoutUntilDelivered(Duration).

* When using UnableToDeliverStrategy::Fail, I think the delivery should not fail instantly but try to deliver the sample to all remaining subscribers/servers/clients and then fail

elBoberido: > * Some thoughts in my hand. Can you add an option, BlockWithTimeout to the ZeroCopyConnectiong thingie?

I would solve this differently. The `wait_while` should provide a elapsed time. The elapsed time would then be passed to the degradation handler via the `DegradationContext` (together with a retry counter). The user can then discard the sample and continue via the handler. As shortcut, I would just add a `retry_timeout` to the publisher builder. This is also easier to implement for the language bindings

> * Where does the UnableToDeliverAction come into play?

This is hidden from the user. I did not want to pass a `DegradationAction` to the `blocking_send`. So the `DegradationAction` is mapped to the `UnableToDeliverAction`.

> * I would keep UnableToDeliverStrategy::Block. Whenever we have a blocking call we add the explicit block prefix. I would keep it here. Maybe rename it to BlockUntilDelivered. And add here the Option BlockWithTimeoutUntilDelivered(Duration).

This makes the language bindings more complicated. I would suggest to replace the `Block` with `Retry` since this is also what one would use to block with the degradation handler.

> * When using UnableToDeliverStrategy::Fail, I think the delivery should not fail instantly but try to deliver the sample to all remaining subscribers/servers/clients and then fail

Good point. At some point we might need to give the user to decide whether to
* fail fast
* best effort (need better name)
* all or nothing (need better name) -> with this, the `deliver_offset` would first check if it can deliver the sample to all connections and call the degradation handler if it is not possible. If the user wants to retry, it would block until all queues have some space for the new sample. It the user wants to fail, it would not deliver the sample to any subscriber.

I think the current behavior is to fail fast when `ZeroCopySendError::InternalError` occurs or the degradation for `ZeroCopySendError::ConnectionCorrupted` is set to fail. Can you please check. If this is the case, what do you think about having `DegradationAction::FailSlow` and `DegradationAction::FailFast`?

elBoberido:
PR showed as merged due to the merge of #1588

Issue #1481, opened 2026-03-25
Implement a byte-wise atomic wrapper to prevent UB


FerdinandSpitzschnueffler:
## Brief feature description

In a multithreaded context, reading a value that may be concurrently written to can lead to undefined behavior. For example:

```rust
let a: T = SomeTValue;
let b = a; // copy
```

Even if the reading of `a` is detected as concurrent with a write and its result is discarded, copying `a` still triggers undefined behavior. To fix this, we introduce a wrapper type that provides byte-wise atomic read and write accesses on the inner type `T`. This wrapper will only guarantee atomicity at the byte level; it will not provide higher-level thread-safety guarantees. Users must still enforce proper synchronization.

elBoberido:
@FerdinandSpitzschnueffler in the end torn-writes and torn-reads are a problem of the user and the user has to take care of it. The wrapper does only ensure that the memory copy is not UB, but does not care about data integrity.

FerdinandSpitzschnueffler:
> [@FerdinandSpitzschnueffler](https://github.com/FerdinandSpitzschnueffler) in the end torn-writes and torn-reads are a problem of the user and the user has to take care of it. The wrapper does only ensure that the memory copy is not UB, but does not care about data integrity.

Exactly!

PR #1577, opened 2026-04-22
[#1481] byte-wise atomic wrapper to prevent UB

FerdinandSpitzschnueffler:
## Notes for Reviewer

To handle padding bytes in the wrapper `ByteAtomic`, the new trait `AtomicCopy` was introduced. It basically provides a method to apply a callback to each offset-size pair of every field of a type. 

## PR Reviewer Reminders

* Public API is documented

## References

Relates to #1481


iceoryx2-bb/derive-macros/src/zero_copy_send_compile_tests.rs
FerdinandSpitzschnueffler, line 1: The changes to this file were made because the tests didn't compile for reasons other than those stated in the test names.

iceoryx2-bb/elementary-traits/src/atomic_copy.rs
```diff
@@ -0,0 +1,110 @@
...
+///
+/// This trait provides a method to apply a callback to each offset-size pair of every field
+/// of the type.
+///
+/// # Safety
+///
+/// * See Safety section of [`crate::zero_copy_send::ZeroCopySend`].
+/// * Implementations of this trait must ensure that the offset and size of each field are
+///   calculated correctly. Otherwise, undefined behavior may occur.
+pub unsafe trait AtomicCopy: ZeroCopySend + Copy + 'static {
```
elfenpiff: I think `ZeroCopySend` is strictly speaking not needed for `AtomicCopy`. First of all, it does not have any dependencies to `ZeroCopySend` or things that trait provides and `AtomicCopy` does not require `#[repr(C)]`. It could be used also in just one process and the member ordering can be anything.

iceoryx2-bb/container/src/string/static_string.rs
```diff
@@ -63,6 +65,21 @@ pub struct StaticString<const CAPACITY: usize> {
     len: u64,
 }
 
+unsafe impl<const CAPACITY: usize> AtomicCopy for StaticString<CAPACITY> {
+    fn __for_each_field<F: FnMut(usize, usize)>(&self, base_offset: usize, callback: &mut F) {
+        let aligned_base_offset = align_to::<Self>(base_offset);
+        callback(
+            aligned_base_offset + core::mem::offset_of!(Self, data),
+            self.len(),
```
elfenpiff: Just for clarity, could you please use here: `core::mem::size_of::<u8>() * self.len()`

iceoryx2-bb/container/src/string/static_string.rs
```diff
@@ -63,6 +65,21 @@ pub struct StaticString<const CAPACITY: usize> {
...
+        );
+        callback(
+            aligned_base_offset + core::mem::offset_of!(Self, terminator),
+            1,
```
elfenpiff: Just for clarity, could you please use here: `core::mem::size_of::<u8>()`

iceoryx2-bb/container/src/string/static_string.rs
```diff
@@ -63,6 +65,21 @@ pub struct StaticString<const CAPACITY: usize> {
...
+        );
+        callback(aligned_base_offset + core::mem::offset_of!(Self, len), 8);
```
elfenpiff: Just for clarity, could you please use here: `core::mem::size_of::<u64>()`

iceoryx2-bb/container/src/byte_atomic.rs
```diff
@@ -0,0 +1,111 @@
...
+
+unsafe impl<T: AtomicCopy, const SIZE: usize> ZeroCopySend for ByteAtomic<T, SIZE> {}
+
+impl<T: AtomicCopy, const SIZE: usize> ByteAtomic<T, SIZE> {
+    /// Creates a new [`ByteAtomic`] that contains the passed value. It fails when the size
+    /// of the value and `SIZE` do not match.
+    pub fn new(value: T) -> Result<Self, ByteAtomicError> {
+        // TODO: remove the following check once size_of::<T>() can be directly used in the
+        // struct definition
+        if size_of::<T>() != SIZE {
```
elfenpiff, line 45: Could we use here the rust static assert from `iceoryx2-bb/elementary/src/static_assert.rs`. This would fail at compile time and would get rid of the runtime check and we could remove the `ByteAtomicError` as well.

iceoryx2-bb/elementary-traits/src/atomic_copy.rs
```diff
@@ -0,0 +1,110 @@
...
+    #[doc(hidden)]
+    /// Iterates over the fields of the type, calculates the offset relative to the provided
+    /// base_offset, and applies the provided callback to each offset-size pair of the field.
+    fn __for_each_field<F: FnMut(usize, usize)>(&self, _base_offset: usize, _callback: &mut F) {}
```
elfenpiff: I am wondering if it would make sense to not provide a default implementation here since the default implementation is wrong if it is not explicilty overridden. Would it be possible to remove it?

iceoryx2-bb/container/src/byte_atomic.rs
```diff
@@ -0,0 +1,111 @@
...
+                data: core::array::from_fn(|i| AtomicU8::new(unsafe { *value_ptr.add(i) })),
+                _inner_type: PhantomData,
+            });
+        }
+
+        // If the passed value is not a scalar, it may contain padding bytes. Reading or copying
+        // these padding bytes would lead to undefined behavior. Therefore, we first set all bytes
+        // to zero and then copy only the fields, i.e. the initialized bytes, of the passed value.
+        let mut bytes = [0u8; SIZE];
+        value.__for_each_field(0, &mut |offset, size| {
```
elfenpiff, line 56: The first parameter is the base offset, why can it be zero? What does the base offset define? Could you clarify this maybe in the comment of the `__for_each_field` documentation.

iceoryx2-bb/container/src/byte_atomic.rs
```diff
@@ -0,0 +1,111 @@
...
+    /// Copies the stored value byte-wise into a [`MaybeUninit<T>`].
+    ///
+    /// # Safety
+    ///
+    /// * When the value is concurrently written to, torn-reads are possible. The user must take care
+    ///   of the data integrity.
+    pub unsafe fn read(&self) -> MaybeUninit<T> {
+        let mut data: MaybeUninit<T> = MaybeUninit::uninit();
+        let data_ptr = data.as_mut_ptr() as *mut u8;
+        for (i, item) in self.data.iter().enumerate() {
```
elfenpiff, line 76: I think we need to distinguish here as well between scalar, where we can just iterate over all `item`s (atomic u8), and a type with padding. Then we need to use `__for_each_field` otherwise we write into the padding bytes, and I have no idea if this is defined behavior - but I would guess it isn't.

iceoryx2-bb/container/src/byte_atomic.rs
elfenpiff: Is it required in a recursion where it increments the offset internally?

iceoryx2-bb/container/src/byte_atomic.rs
FerdinandSpitzschnueffler: No, unfortunately this doesn't work with generics.

iceoryx2-bb/container/src/byte_atomic.rs
FerdinandSpitzschnueffler: I don't think so. Padding bytes can be written to via a reference to the memory (what we basically do here). They probably will be uninitialized again once the user extracts `T` with `assume_init`, but writing should be fine. I'm not 100% sure, but see https://google.github.io/learn_unsafe_rust/advanced_unsafety/uninitialized.html#padding + Miri does not complain.

iceoryx2-bb/elementary-traits/src/atomic_copy.rs
FerdinandSpitzschnueffler: Removed it.

iceoryx2-bb/elementary-traits/src/atomic_copy.rs
FerdinandSpitzschnueffler: Yes, I was able to remove the default implementation and by adding the `__for_each_field` implementation to the scalar types, I could also remove `__is_scalar`.

iceoryx2-bb/container/src/byte_atomic.rs
FerdinandSpitzschnueffler: Yes, the base offset is needed when `__for_each_field` is called recursively, for example in a nested struct. I added comments on that in the trait and the macro implementation.

iceoryx2-bb/container/tests-common/src/byte_atomic_tests.rs
```diff
@@ -0,0 +1,235 @@
...
+        assert_that!(read_value.b, eq new_value.b);
+        assert_that!(read_value.c, eq new_value.c);
+        assert_that!(read_value.d.0, eq new_value.d.0);
+        assert_that!(read_value.d.1, eq new_value.d.1);
+        assert_that!(read_value.d.2, eq new_value.d.2);
+        assert_that!(read_value.d.3, eq new_value.d.3);
+    }
+}
+
+// TODO: The following tests should be run with Miri but using iceoryx2_bb_posix::barrier/thread::*
```
elfenpiff: Can you create an issue that this is handled globally and we have some barrier and thread wrapper so that this can be exchanged. We could put this into `iceoryx2-bb-posix` and depending on the feature flag (maybe miri) we then use the normal std variants.

This would require us to wrap them in a stub thing that has the same API has the `iceoryx2-bb-posix` variants.

elfenpiff (dismissed):

iceoryx2-bb/container/tests-common/src/byte_atomic_tests.rs
orecham: @elfenpiff @FerdinandSpitzschnueffler  ~~Could consider putting it in `iceoryx2-bb-concurrency`, where similar switches for `loom` currently are.~~ Actually that might tangle the dependency tree. Again this requirement is asking for some layer between `iceoryx2-bb-posix` and the upper layers...

BTW `loom` uses `cfg` instead of feature flags for this purpose: https://docs.rs/loom/0.7.2/loom/#running-loom-tests

This avoids polluting the feature set.

FerdinandSpitzschnueffler: Added https://github.com/eclipse-iceoryx/iceoryx2/issues/1601.

elfenpiff (approved):

Issue #1584, opened 2026-04-25
Feature Request: Public API to remove/cleanup orphaned service static configs


0-jake-0:
## Problem

When an iceoryx2 process terminates ungracefully (e.g., `SIGKILL`, crash during service initialization), service static configs (payload type fingerprints) persist indefinitely in `/dev/shm/`. These orphaned configs are not removed by the existing `Node::cleanup_dead_nodes()` mechanism, which only cleans up node monitoring files and node-owned resources.

On restart, if another process tries to open the same service, `open_or_create()` returns `IncompatibleTypes` because the stale config's type fingerprint no longer matches what the new process is trying to create.

### Impact

- **Multi-process deployments** that experience crashes or forced termination must manually clean `/dev/shm/iox2_*` files
- **No graceful in-process recovery path** for orphaned service configs
- **Users forced to use external cleanup scripts** or accept manual intervention on deployment
- **Not suitable for containerized/orchestrated environments** (e.g., Kubernetes) where processes are forcefully terminated regularly

## Example

\`\`\`bash
# Process 1: Start publisher, crash mid-initialization
cargo run --example publisher &
pkill -9 publisher  # Ungraceful kill

# Process 2: Try to subscribe — fails
cargo run --example subscriber
# Error: ServiceConfigMismatch: IncompatibleTypes for service "my/service"

# Workaround: Manual cleanup required
rm -f /dev/shm/iox2_*
cargo run --example subscriber  # Now works
\`\`\`

## Requested Solution

Provide a public API to remove/cleanup orphaned service static configs. Options:

1. **Extend `Node::cleanup_dead_nodes()`** — Optionally also remove service configs for services with no live participants
2. **Add service removal API** — e.g., `Service::<ipc::Service>::remove("my/service")`  
3. **Document the cleanup procedure** — At minimum, guide users on safe cleanup and when service configs can be removed

## Environment
- iceoryx2 version: 0.8.1
- Use case: Streaming data processing with graceful recovery from process crashes
- Deployment: Container/orchestration environments with forceful process termination

Thank you for this excellent library! This is a feature request for improved lifecycle management of orphaned service configs in multi-process deployments.

elBoberido:
@0-jake-0 sorry for the late reply. We are currently quite busy with work for the next release.

[...]

PR #1587, opened 2026-04-26
[#1584] Expose Service::remove() for orphaned static config cleanup

0-jake-0:
Add a `Service::remove` trait method (with a crate-private `remove_static_service_config` helper) so users can clean up orphaned static service configs left behind after an abrupt process termination (e.g. SIGKILL). Mirror the new API in the C and Python FFIs as `iox2_service_remove` and `Service.remove`.

See https://github.com/eclipse-iceoryx/iceoryx2/issues/1584


FerdinandSpitzschnueffler:
@elfenpiff @0-jake-0 #1584 is already closed. Can we close this PR or, if something is missing, reopen the issue?

Issue #1578, opened 2026-04-22
`open_or_create` fails with `ServiceInCorruptedState` / `IsMarkedForDestruction` during concurrent `cleanup_dead_nodes`


erdody:
# `open_or_create` fails with `ServiceInCorruptedState` / `IsMarkedForDestruction` during concurrent `cleanup_dead_nodes`

## ToDos (Updated 28.04.2026)

After a triage and debug session with @elBoberido we confirmed the following issues that need to be fixed:

* See branch: `bob-be-careful-with-that-axe`
  * [x]  1. `process_state.rs:397` and `process_state.rs:552`. Add to `lock_state_file` an error when the `number_of_links()` to a file is zero. This means the file lock cannot be acquired since the file does no longer exist in the file system. This error must be handled in `ProcessGuardLockError`
  * [x] 2. The error `ProcessIsBeingCleanedUpOrCrashedDuringCleanup` from `process_state.rs` must be propagated correctly via the `iceoryx2_cal` monitoring concept to the node cleanup (`remove_stale_resources`) function. Currently, it is translated as `InternalError` and causes a cleanup abort. When the node is waiting for a cleanup to be finished, this error must be propagated so that the waiter is able to wait and does not abort.
  * [x] 3. Retry to open or create a service when the error `ServiceInCorruptedState` or `IsMarkedForDestruction` is encountered. (See the report below for details)
  * [x] 4. When an existing service is opened, the node must check if there are dead nodes connected to the service. If so, it must either clean them up or wait until they are cleaned up, or until the cleaner crashes. Otherwise, it tries to connect to a service that might have no more port resources left, and the connection fails.

* [x] 5. When the cleaner crashes while releasing a port via `DynamicConfig::**::release_**_handle()` it is possible that the underlying index is leaked. This results in a reduced port list size and the user might encounter an `ExceedsMaxSupported**` error when trying to create a port.
* [x] 6. When an application is killed while initializing resources like a data segment or a connection, it is not possible to remove them due to the restricted permissions. This can be observed by calling `ls -alh /dev/shm` with shared memory objects with write-only permissions. The solution is to "aggressively" increase the permissions before removing the resource. This is safe since the cleanup lock was acquired and the ownership belongs fully to the cleanup process.
* [x] 7. ZeroCopyConnection `remove_sender`/`remove_receiver` fails when connection is not initialized
* [x] 8. NamedConcept remove may fail when not-initialized.
[...]

PR #1592, opened 2026-04-30
[#1578] harden cleanup process

elfenpiff:
## Notes for Reviewer

Here is the debugging branch: https://github.com/eclipse-iceoryx/iceoryx2/pull/1590
All adjustments should be applied here - just in clean.

This PR should address todo 1. - 4. from #1578 

## PR Reviewer Reminders

* Public API is documented

## References

Relates to #1578

### DIFF
--- a/iceoryx2-bb/posix/src/directory.rs
+++ b/iceoryx2-bb/posix/src/directory.rs
@@ -372,13 +372,13 @@ impl Directory {
     pub fn remove_empty(path: &Path) -> Result<(), DirectoryRemoveError> {
         if unsafe { posix::rmdir(path.as_c_str()) } == -1 {
             let msg = format!("Unable to remove empty directory \"{path}\"");
-            handle_errno!(DirectoryRemoveError, from "Directory::remove",
+            handle_errno!(DirectoryRemoveError, from "Directory::remove_empty()",
+                success Errno::ENOENT => (),
                 Errno::EACCES => (InsufficientPermissions, "{} due to insufficient permissions.", msg),
                 Errno::EPERM => (InsufficientPermissions, "{} due to insufficient permissions.", msg),
                 Errno::EBUSY => (CurrentlyInUse, "{} since the directory is currently in use.", msg),
                 Errno::EINVAL => (LastComponentIsDot, "{} since the last path component is \".\".", msg),
                 Errno::ELOOP => (LoopInSymbolicLinks, "{} due to a loop in the symbolic links of the path \".\".", msg),
-                Errno::ENOENT => (DanglingSymbolicLink, "{} since the path contains a dangling symbolic link.", msg),
                 Errno::EEXIST => (DirectoryDoesNotExist, "{} since the directory does not exist.", msg),
                 Errno::ENOTDIR => (NotADirectory, "{} since it is not a directory.", msg),
                 Errno::EROFS => (ResidesOnReadOnlyFileSystem, "{} since the directory resides on a read only file system.", msg),
--- a/iceoryx2-cxx/include/iox2/node_state.hpp
+++ b/iceoryx2-cxx/include/iox2/node_state.hpp
@@ -68,7 +68,18 @@ class DeadNodeView {
 
     /// Removes all stale resources of the dead [`Node`]. On error it returns a [`NodeCleanupFailure`].
     /// It returns true if the stale resources could be removed, otherwise false.
-    auto remove_stale_resources() -> iox2::bb::Expected<bool, NodeCleanupFailure>;
+    auto try_remove_stale_resources() -> iox2::bb::Expected<void, NodeCleanupFailure>;
+
+    /// Removes all stale resources of a dead [`Node`]. If another instance
+    /// is already removing the dead [`Node`] it waits until the other instance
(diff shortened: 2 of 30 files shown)


iceoryx2-bb/posix/src/directory.rs
```diff
@@ -372,13 +372,13 @@ impl Directory {
     pub fn remove_empty(path: &Path) -> Result<(), DirectoryRemoveError> {
         if unsafe { posix::rmdir(path.as_c_str()) } == -1 {
             let msg = format!("Unable to remove empty directory \"{path}\"");
-            handle_errno!(DirectoryRemoveError, from "Directory::remove",
+            handle_errno!(DirectoryRemoveError, from "Directory::remove_empty()",
```
elBoberido, line 375: At some point we discussed whether this function should return a bool or enum to indicate whether the directory was removed or just not there. I'm fine with just this change, though.

iceoryx2-cxx/include/iox2/node_state.hpp
```diff
@@ -68,7 +68,18 @@ class DeadNodeView {
 
     /// Removes all stale resources of the dead [`Node`]. On error it returns a [`NodeCleanupFailure`].
     /// It returns true if the stale resources could be removed, otherwise false.
-    auto remove_stale_resources() -> iox2::bb::Expected<bool, NodeCleanupFailure>;
+    auto try_remove_stale_resources() -> iox2::bb::Expected<void, NodeCleanupFailure>;
```
elBoberido, line 71: Since this is a breaking change, shouldn't it be added to the changelog?

iceoryx2/src/node/mod.rs
```diff
@@ -207,6 +210,8 @@ pub enum NodeCreationFailure {
     InsufficientPermissions,
     /// Errors that indicate either an implementation issue or a wrongly configured system.
     InternalError,
+    /// Indicates that another "instance" on the system removed the resource required by the [`Node`].
+    SystemCorrupted,
```
elBoberido, line 214: If the user is supposed to just retry when this happens, then the name is not ideal.

iceoryx2/src/service/port_factory/mod.rs
```diff
@@ -102,6 +107,26 @@ pub trait PortFactory {
     ) -> Result<(), NodeListFailure>;
 }
 
+pub(crate) fn cleanup_dead_nodes_in_service<T: PortFactory>(
```
elBoberido: shouldn't this be `blocking_cleanup_dead_nodes_in_service`

iceoryx2/src/service/port_factory/publish_subscribe.rs
```diff
@@ -142,11 +140,15 @@ impl<
...
     pub(crate) fn new(service: ServiceState<Service, NoResource>) -> Self {
-        Self {
+        let shared_node = service.shared_node.clone();
+        let new_self = Self {
             service: Arc::new(service),
             _payload: PhantomData,
             _user_header: PhantomData,
-        }
+        };
+        cleanup_dead_nodes_in_service(&new_self, shared_node);
```
elBoberido: Shouldn't this respect the `cleanup_dead_nodes_on_creation` setting?

Same with the other services.

iceoryx2/src/node/mod.rs
elfenpiff: Nope, this is absolutely correct. The node creation must always succeed. This error indicates that someone removed internal node resources before the node was even created. This should never happen and even the node list function or removing a dead node from a service would never interfer with a node creation.

You cannot list a node which is not yet being created and such a node cannot connect to a service, since it is not yet created.

iceoryx2/src/service/port_factory/publish_subscribe.rs
elfenpiff: Nope, this cleans up dead nodes whenever it connects to a service, not when a node is created or destroyed. This shall not respect the setting and shall always happen since here the user is directly affected if a dead node is lying around.

The service connection will not fail if the cleanup fails. This could be caused by insufficient permissions for instance, but one should always try to remove the stale resources.

elBoberido: Well, from a user point of view, one might want to turn this off, if the cleanup on node creation or shutdown is turned off

elBoberido (commented):
Looks fine and can be merged once the the markdown issue is fixed.

I would also suggest to consider adding an option to turn the cleanup off.

iceoryx2/src/service/port_factory/publish_subscribe.rs
elfenpiff: Why? This only has benefits and no drawbacks, or do I overlook something here?

When you open an existing service (only then) the service is checked for dead nodes. If there are some, they are removed and with that stale resources, that might block your port builder, are removed.

elBoberido: Well, with the same argument one could ask why the cleanup on node creation/shutdown can be turned off :)

elBoberido (approved):

PR #1594, opened 2026-05-05
[#1593] Make commit message verification more robust

### LINKED ISSUE #1593: Improve commit message verification
## Brief Description

The [current verification script](https://github.com/eclipse-iceoryx/iceoryx2/blob/main/internal/scripts/ci_test_commit_msg.sh) is not detecting commit messages that do not adhere to the guidelines.  
e.g. Commits in: 
* https://github.com/eclipse-iceoryx/iceoryx2/pull/1592/commits
* https://github.com/eclipse-iceoryx/iceoryx2/pull/1583/commits

orecham:
## Notes for Reviewer

1. Add `just` script for commit verification; remove corresponding shell script
2. Properly retrieve commits in branch
3. Ensure subject begin with the issue number `[#XYZ]`
4. Ensure message begins with a capital letter
5. Utilize [`imperative`](https://github.com/crate-ci/imperative/tree/master) crate to check for imperative mood
    1. As per recommendation in the [commit guidelines][commit-guidelines]
    1. Does not catch everything but catches common offenders
 
The imperative mood check can be omitted if there is worry about false positives.

## PR Reviewer Reminders

* Public API is documented

## References

Closes #1593


.just/verify.just
```diff
@@ -51,6 +55,105 @@ _verify-std-propagation-package target:
         rust-script {{VERIFY_STD_PROPAGATION_SCRIPT}} "$MANIFEST"
     fi
 
+[private]
+_verify-commits base="":
+    #!/usr/bin/env bash
+    set -euo pipefail
+
+    GUIDELINES_URL='https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html'
```
elfenpiff: I do not like those guidelines at all, and also the imperative mood.

If the rest of the crew:
@elBoberido @ComicSansMS @dkroenke @FerdinandSpitzschnueffler @budrus 

Like it, I will not intervene here. But the gains are, in my opinion, nearly nonexistent. And multi line commit messages with detailed lists and explanations are just wrong (why add a guideline for something that clearly belongs into the code and not hidden in a commit message)

When I am looking for a specific commit with lazygit, I first take a look at the changes, and if this is not sufficient, I will take a look at the commit message.

But I do not care for wording or spelling; it must be short and summarize what happened in this commit. If it is a capital case, a comma is missing, the line length is 50 or 120, I don't care.
But I care that the CI is not adding another check that lets my PR fail because my commit message didn't start with a capital letter, or was in a non-imperative mood.

orecham: @elfenpiff I like the guideline. The imperative mood check encourages grouping changes into atomic units (whether the content of the commit actually follows this is another question) and if done correctly makes cherry picking easier.

The capitialiation I agree is nitpicky but I enjoy a clean commit log. The friction it adds is not greater than enforicing the issue number is tagged - just reword the commit.

FerdinandSpitzschnueffler: I agree with @orecham and already follow this guideline (apart from the line length - 50 is sometimes too short). Additionally, the checklist in our PR template requires this, so it makes sense to check it. If we decide against, we should also remove this point from the checklist.

elfenpiff: If you like it, then I am fine with it. @orecham would it be possible to set the line length to 110. This is also the line length we have in Rust, C++ and Python. I see no reason why git should be here an exception.

orecham: There is no line check at the moment. We can leave it out to allow some flexibility. The check only fails on the things I deemed worth being strict about.

ComicSansMS: I don't think it is sensible to match the git commit message line length to the code line length.

The reason for the length limits on git commit messages is to ensure that they play safe with tooling. For instance, github cuts overlong commit messages between the 50th and 70th character (depending on the length of the last word). If you stay below the 50 character mark you are guaranteed that the first line will always be fully displayed in the commit overview. Also, some terminals get really confused by implicit line breaks if `git log` has to break a line that is too long for display, which messes up the console output.

The established convention for git seems to be 50/72: 50 characters for the first line, followed by a blank line and then 72 characters for every line afterwards. If we follow a line limit for commit messages at all, it should be this, as that is what all the tooling expects. For instance, this is also what vim's syntax highlighting will use when writing commit messages.

elBoberido: I would be careful to be too strict with this. Especially new contributors often struggle with the commit format and in general I'm happy when they at least have an issue number in the commit message.

I also try to follow the guidelines but I would only enforce the issue number and nothing else.

grafik

orecham: @elBoberido This is a matter of taste, so I guess we will never all agree but we will need to settle on a convention. 

I am in favor of a commit history that reads like steps building upon each other and especially value cherry-pickability (the convention doesn't guarantee this but it does encourage it). Commits like: `[#XYZ] first attempt` are more effort for me to reason about and work with.

elBoberido: @orecham fully agree and I also do this and try to have atomic commits which are building. The issue is just with casual contributors and all the things they have to do to be able to merge a PR. The ECA is already a big hurdle and I fear we scare some of them away if we are too strict with some stuff that is not that important. We have to find a good balance.

elfenpiff: @orecham @elBoberido can we wrap up the discussion here and merge it? The `[#123]` prefix check is definitely valuable.

.just/lint.just
```diff
@@ -40,22 +40,45 @@ _lint-dispatch workspace target *flags:
...
+            echo "Usage: just lint <workspace> <target> [flags]"
+            echo ""
+            echo "Workspace-targeted (require <target>):"
+            echo "  sdk             lint the main workspace"
+            echo "  integrations    lint the integrations workspace"
+            echo "  all             lint everything"
+            echo ""
+            echo "  Targets:"
+            echo "    rust            cargo fmt check / fix"
+            echo "    markdown        markdownlint check / fix"
```
orecham, line 54: TODO: document commits linting check

.just/verify.just
orecham: @elfenpiff @elBoberido  I propose this:

1. Keep the issue number check
1. Keep the imperative mood check 
1. Drop the requirement to start first word with capital letter

elBoberido: I'm all for the issue number check. Because of the reasons mentioned above, I'm skeptical about the other things but I leave it to you two to bikeshed this out.

PR #1582, opened 2026-04-24
[#996] Refactor BumpAllocators

dkroenke:
## Notes for Reviewer

- Consolidate BumpAllocator from iceoryx2-bb-memory into BumpAllocator in iceoryx2-bb-elementary to avoid maintaining two BumpAllocator implementations with almost same functionality
- ~~add Traits for `core::ptr::NonNull` to make use of safe functions `from_ref()` and `from_mut` and use it in the API Adaptions for the `BumpAllocator`~~
- Rename `bump_allocator` module in iceoryx2-cal to `shm_bump_allocator` to better distinguish it from the existing bump_allocator module.

## Pre-Review Checklist for the PR Author

* [x] Add sensible notes for the reviewer
* [x] PR title is short, expressive and meaningful
* [x] Consider switching the PR to a draft (`Convert to draft`)
    * as draft PR, the CI will be skipped for pushes
* [x] Relevant issues are linked in the [References](#references) section
* [x] Branch follows the naming format (`iox2-123-introduce-posix-ipc-example`)
* [x] Commits messages are according to this [guideline][commit-guidelines]
    * [x] Commit messages have the issue ID (`[#123] Add posix ipc example`)
    * Keep in mind to use the same email that was used to sign the [Eclipse Contributor Agreement][eca]
* [x] Tests follow the [best practice for testing][testing]
* [x] Changelog updated [in the unreleased section][changelog] including API breaking changes

[commit-guidelines]: https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
[eca]: http://www.eclipse.org/legal/ECA.php
[testing]: https://github.com/eclipse-iceoryx/iceoryx/blob/master/doc/website/concepts/best-practice-for-testing.md
[changelog]: https://github.com/eclipse-iceoryx/iceoryx2/blob/main/doc/release-notes/iceoryx2-unreleased.md

## PR Reviewer Reminders

* Commits are properly organized and messages are according to the guideline
* Unit tests have been written for new behavior
* Public API is documented
* PR title describes the changes

## References

Closes #996

PR #1611, opened 2026-05-09
[#1578] robust lock-free unique index set

elfenpiff:
## Notes for Reviewer

Introduces a unique index set where the user can specify an owner id whenever they take out an index. The owner can be a PID + epoch. This allows other processes to recover leaked indices of dead processes - therefore it is robust. It never leaks or enters an invalid state if a process dies.

## PR Reviewer Reminders

* Public API is documented

## References

Relates to #1578

### DIFF
--- a/iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
+++ b/iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
@@ -527,10 +527,29 @@ impl RobustUniqueIndexSet {
 }
 
 #[derive(Debug)]
+pub struct StaticRobustUniqueIndexSetData<const CAPACITY: usize> {
+    cells: [AtomicU64; CAPACITY],
+}
+
+impl<const CAPACITY: usize> Default for StaticRobustUniqueIndexSetData<CAPACITY> {
+    fn default() -> Self {
+        Self {
+            cells: [const { AtomicU64::new(0) }; CAPACITY],
+        }
+    }
+}
+
+impl<const CAPACITY: usize> StaticRobustUniqueIndexSetData<CAPACITY> {
+    pub fn allocator(&mut self) -> BumpAllocator {
+        BumpAllocator::new(self.cells.as_mut_ptr().cast())
+    }
+}
+
+#[derive(Debug)]
 #[repr(C)]
 pub struct StaticRobustUniqueIndexSet<const CAPACITY: usize> {
     state: RobustUniqueIndexSet,
-    cells: [AtomicU64; CAPACITY],
+    cells: StaticRobustUniqueIndexSetData<CAPACITY>,
 }
 
 impl<const CAPACITY: usize> Default for StaticRobustUniqueIndexSet<CAPACITY> {
@@ -566,14 +585,13 @@ impl<const CAPACITY: usize> StaticRobustUniqueIndexSet<CAPACITY> {
 
         let mut new_self = Self {
             state: unsafe { RobustUniqueIndexSet::new_uninit(capacity) },
-            cells: core::array::from_fn(|_| AtomicU64::new(0)),
+            cells: StaticRobustUniqueIndexSetData::default(),
         };
 
-        let allocator = BumpAllocator::new(new_self.cells.as_mut_ptr().cast());
         unsafe {
             new_self
                 .state
-                .init(&allocator)
+                .init(&new_self.cells.allocator())
                 .expect("All required memory is preallocated")
         };
 
--- a/iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
+++ b/iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
@@ -96,7 +96,7 @@ use iceoryx2_bb_elementary::enum_gen;
 use iceoryx2_bb_elementary::relocatable_ptr::{PointerTrait, RelocatablePointer};
 use iceoryx2_bb_elementary_traits::allocator::AllocationError;
 use iceoryx2_bb_elementary_traits::relocatable_container::RelocatableContainer;
-use iceoryx2_log::{fail, fatal_panic};
+use iceoryx2_log::{error, fail, fatal_panic};
 
 use crate::mpmc::unique_index_set_enums::{
     ReleaseMode, ReleaseState, UniqueIndexCreationError, UniqueIndexSetAcquireFailure,
@@ -122,6 +122,11 @@ impl OwnerId {
 
         Ok(Self(value))
     }
+}
+
+struct RepairState {
+    generation_counter: u64,
+    borrowed_indices: u64,
 }
 
 #[repr(C)]
@@ -199,29 +204,39 @@ impl RobustUniqueIndexSet {
     /// Returns if the [`RobustUniqueIndexSet`] is locked or not. If the set is locked no more indices
     /// can be borrowed.
     pub fn is_locked(&self) -> bool {
-        self.borrowed_indices.load(Ordering::Relaxed) == u64::MAX
+        self.generation_counter.load(Ordering::Relaxed) == u64::MAX
     }
 
     fn lock(&self) -> ReleaseState {
-        match self.borrowed_indices.compare_exchange(
-            0,
-            u64::MAX,
-            Ordering::Relaxed,
-            Ordering::Relaxed,
-        ) {
-            Ok(_) => ReleaseState::Locked,
-            Err(_) => ReleaseState::Unlocked,
+        if self.is_locked() {
+            return ReleaseState::Locked;
+        }
+
+        loop {
+            let state = self.repair_borrowed_indices();
+
+            if state.borrowed_indices == 0 {
+                match self.generation_counter.compare_exchange(
+                    state.generation_counter,
+                    u64::MAX,
+                    Ordering::Relaxed,
+                    Ordering::Relaxed,
+                ) {
+                    Ok(_) => return ReleaseState::Locked,
+                    Err(_) => continue,
+                }
+            } else {
+                return ReleaseState::Unlocked;
+            }
         }
     }
 
     /// Returns the current len.
     pub fn borrowed_indices(&self) -> usize {
-        let borrowed_indices = self.borrowed_indices.load(Ordering::Relaxed);
-
-        if borrowed_indices == u64::MAX {
+        if self.is_locked() {
             0
         } else {
-            borrowed_indices as _
+            self.borrowed_indices.load(Ordering::Relaxed) as _
         }
     }
 
@@ -278,13 +293,13 @@ impl RobustUniqueIndexSet {
             //////////////////////////////////////
             let current_generation_count = self.generation_counter.load(Ordering::Acquire);
 
+            if current_generation_count == u64::MAX {
+                fail!(from self, with UniqueIndexSetAcquireFailure::IsLocked,
+                    "{msg} since the RobustUniqueIndexSet is locked.");
+            }
+
             for n in 0..self.capacity {
                 let cell = unsafe { &*self.data_ptr.as_ptr().add(n) };
-
-                if self.is_locked() {
-                    fail!(from self, with UniqueIndexSetAcquireFailure::IsLocked,
-                        "{msg} since the RobustUniqueIndexSet is locked.");
-                }
 
                 match cell.compare_exchange(
                     OwnerId::EMPTY.0,
@@ -293,11 +308,6 @@ impl RobustUniqueIndexSet {
                     Ordering::Relaxed,
                 ) {
                     Ok(_) => {
-                        if self.is_locked() {
-                            fail!(from self, with UniqueIndexSetAcquireFailure::IsLocked,
-                                "{msg} since the RobustUniqueIndexSet is locked.");
-                        }
-
                         self.borrowed_indices.fetch_add(1, Ordering::Relaxed);
 
                         //////////////////////////////////////
@@ -305,7 +315,15 @@ impl RobustUniqueIndexSet {
                         //   1. cell & borrowed_indices
                         //   2. generation counter
                         //////////////////////////////////////
-                        self.generation_counter.fetch_add(1, Ordering::Release);
+                        if self.increment_generation_counter(Ordering::Release)
+                            == ReleaseState::Locked
+                        {
+                            // revert the increment, not important, just for consistency
+                            self.borrowed_indices.fetch_sub(1, Ordering::Relaxed);
+                            cell.store(OwnerId::EMPTY.0, Ordering::Relaxed);
+                            fail!(from self, with UniqueIndexSetAcquireFailure::IsLocked,
+                                "{msg} since the RobustUniqueIndexSet is locked.");
+                        }
                         return Ok(n);
                     }
                     Err(_) => continue,
@@ -333,16 +351,40 @@ impl RobustUniqueIndexSet {
         self.verify_init("release()");
 
         let cell = unsafe { &*self.data_ptr.as_ptr().add(index) };
-        cell.store(OwnerId::EMPTY.0, Ordering::Relaxed);
 
-        self.borrowed_indices.fetch_sub(1, Ordering::Relaxed);
+        let cell_content = cell.load(Ordering::Relaxed);
 
-        //////////////////////////////////////
-        // SYNC POINT enforce order of:
-        //   1. cell & borrowed_indices
-        //   2. generation counter
-        //////////////////////////////////////
-        self.generation_counter.fetch_add(1, Ordering::Release);
+        if cell_content == OwnerId::EMPTY.0 {
+            error!(from self, "Release of the not acquired index {index} prevented." );
+        } else {
+            match cell.compare_exchange(
+                cell_content,
+                OwnerId::EMPTY.0,
+                Ordering::Relaxed,
+                Ordering::Relaxed,
+            ) {
+                Ok(_) => {
+                    cell.store(OwnerId::EMPTY.0, Ordering::Relaxed);
+                    self.borrowed_indices.fetch_sub(1, Ordering::Relaxed);
+
+                    //////////////////////////////////////
+                    // SYNC POINT enforce order of:
+                    //   1. cell & borrowed_indices
+                    //   2. generation counter
+                    //////////////////////////////////////
+                    self.increment_generation_counter(Ordering::Release);
+                }
+                Err(v) => {
+                    if v == OwnerId::EMPTY.0 {
+                        error!(from self,
+                            "Prevented race in releasing the index {index} which would have resulted in a double free.");
+                    } else {
+                        error!(from self,
+                            "It seems like the index {index} was modified while releasing it. This could indicate a corrupted system.");
+                    }
+                }
+            }
+        }
 
         if mode == ReleaseMode::LockIfLastIndex {
             self.lock()
@@ -382,17 +424,21 @@ impl RobustUniqueIndexSet {
                     )
                     .is_ok()
             {
-                self.borrowed_indices.fetch_sub(1, Ordering::Relaxed);
-
                 //////////////////////////////////////
                 // SYNC POINT enforce order of:
-                //   1. cell & borrowed_indices
+                //   1. cell
                 //   2. generation counter
                 //////////////////////////////////////
-                self.generation_counter.fetch_add(1, Ordering::Release);
+                if self.increment_generation_counter(Ordering::Release) == ReleaseState::Locked {
+                    return ReleaseState::Locked;
+                }
 
                 if mode == ReleaseMode::LockIfLastIndex {
                     self.lock();
+                } else {
+                    // must be called here explicitly since self.lock() calls this
+                    // already
+                    self.repair_borrowed_indices();
                 }
             }
         }
@@ -400,6 +446,73 @@ impl RobustUniqueIndexSet {
         match self.is_locked() {
             true => ReleaseState::Locked,
             false => ReleaseState::Unlocked,
+        }
+    }
+
+    fn increment_generation_counter(&self, ordering: Ordering) -> ReleaseState {
+        let mut current_generation_count = self.generation_counter.load(Ordering::Relaxed);
+        loop {
+            if current_generation_count == u64::MAX {
+                return ReleaseState::Locked;
+            }
+
+            match self.generation_counter.compare_exchange(
+                current_generation_count,
+                current_generation_count + 1,
+                ordering,
+                Ordering::Relaxed,
+            ) {
+                Ok(_) => return ReleaseState::Unlocked,
+                Err(v) => current_generation_count = v,
+            }
+        }
+    }
+
+    fn repair_borrowed_indices(&self) -> RepairState {
+        loop {
+            //////////////////////////////////////
+            // SYNC POINT enforce order of:
+            //   1. cell
+            //   2. generation counter
+            //////////////////////////////////////
+            let current_generation_count = self.generation_counter.load(Ordering::Acquire);
+
+            if current_generation_count == u64::MAX {
+                return RepairState {
+                    generation_counter: current_generation_count,
+                    borrowed_indices: 0,
+                };
+            }
+
+            let mut count = 0;
+            for n in 0..self.capacity {
+                let cell = unsafe { &*self.data_ptr.as_ptr().add(n) };
+                if cell.load(Ordering::Relaxed) != OwnerId::EMPTY.0 {
+                    count += 1;
+                }
+            }
+
+            self.borrowed_indices.store(count, Ordering::Relaxed);
+
+            //////////////////////////////////////
+            // SYNC POINT enforce order of:
+            //   1. borrowed_indices
+            //   2. generation counter
+            //////////////////////////////////////
+            match self.generation_counter.compare_exchange(
+                current_generation_count,
+                current_generation_count + 1,
+                Ordering::Release,
+                Ordering::Relaxed,
+            ) {
+                Ok(_) => {
+                    return RepairState {
+                        generation_counter: current_generation_count + 1,
+                        borrowed_indices: count,
+                    };
+                }
+                Err(_) => continue,
+            }
         }
     }
 }
--- a/iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
+++ b/iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
@@ -16,6 +16,13 @@
 //! processes. Can be used as a building block for
 //! allocators or lock-free containers.
 //!
+//! The key properties of the robust unique index set:
+//!
+//!  * contains unique indices from 0 to CAPACITY
+//!  * indices can be acquired or released
+//!  * when releasing the last index the user has the option to lock the index set which means that
+//!    new indices can be never acquired again
+//!
 //! # Example
 //!
 //! ## Runtime fixed size RobustUniqueIndexSet
@@ -44,7 +51,7 @@
 //! println!("Acquired index {}", new_index);
 //!
 //! // return the index to the index set
-//! unsafe { index_set.release(new_index, ReleaseMode::default()) };
+//! unsafe { index_set.release(new_index, owner_id, ReleaseMode::default()) };
 //! ```
 //!
 //! ## Compile time FixedSizeUniqueIndexSet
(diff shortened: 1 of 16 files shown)


iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
```diff
@@ -0,0 +1,644 @@
...
+//! # extern crate iceoryx2_bb_loggers;
+//!
+//! use iceoryx2_bb_elementary::bump_allocator::*;
+//! use iceoryx2_bb_lock_free::mpmc::robust_unique_index_set::*;
+//! use iceoryx2_bb_lock_free::mpmc::unique_index_set_enums::*;
+//! use iceoryx2_bb_elementary_traits::relocatable_container::*;
+//!
+//! const CAPACITY: usize = 128;
+//! let mut memory = [0u8; RobustUniqueIndexSet::const_memory_size(CAPACITY)];
+//! let allocator = BumpAllocator::new(memory.as_mut_ptr());
```
elBoberido, line 40: This will break with #1582 and not show up as merge conflict, since it is a new file. We need to ensure that #1582 is rebased after this one is merged or the other way around.

cc @dkroenke

iceoryx2-bb/lock-free/tests-common/src/mpmc_robust_unique_index_set_tests.rs
```diff
@@ -0,0 +1,366 @@
...
+    let owner_id = OwnerId::new(123).unwrap();
+    let mut indices = vec![];
+    for n in 0..CAPACITY {
+        assert_that!(sut.borrowed_indices(), eq n);
+        indices.push(sut.acquire(owner_id).unwrap());
+        assert_that!(sut.borrowed_indices(), eq n + 1);
+    }
+
+    for n in (1..CAPACITY + 1).rev() {
+        assert_that!(sut.borrowed_indices(), eq n);
```
elBoberido: With this, the logic would match more closely to the loop above

```suggestion
    for n in (0..CAPACITY).rev() {
        assert_that!(sut.borrowed_indices(), eq n + 1);
```

Alternatively, `1..=CAPATCITY` would have the same effect as `1..CAPACITY + 1`

iceoryx2-bb/lock-free/tests-common/src/mpmc_robust_unique_index_set_tests.rs
```diff
@@ -0,0 +1,366 @@
...
+#[test]
+pub fn indices_are_unique() {
+    let sut = StaticRobustUniqueIndexSet::<CAPACITY>::new();
+
+    let owner_id = OwnerId::new(123).unwrap();
+    let mut indices = vec![];
+    for _ in 0..CAPACITY {
+        let index = sut.acquire(owner_id).unwrap();
+
+        assert_that!(indices, not_contains index);
```
elBoberido: Food for thought

```suggestion
        assert_that!(indices, excludes index);
```

iceoryx2-bb/lock-free/tests-common/src/mpmc_robust_unique_index_set_tests.rs
```diff
@@ -0,0 +1,366 @@
...
+    for _ in 0..CAPACITY {
+        indices.push(sut.acquire(owner_id).unwrap());
+    }
+
+    for _ in 0..CAPACITY - 1 {
+        assert_that!(sut.release(indices.pop().unwrap(), owner_id, ReleaseMode::LockIfLastIndex), eq Ok(ReleaseState::Unlocked));
+    }
+
+    assert_that!(sut.release(indices.pop().unwrap(), owner_id, ReleaseMode::LockIfLastIndex), eq Ok(ReleaseState::Locked));
+    assert_that!(sut.is_locked(), eq true);
```
elBoberido, line 182: Would it make sense to add a unlocked check before the lock action?

```suggestion
    assert_that!(sut.is_locked(), eq false);
    assert_that!(sut.release(indices.pop().unwrap(), owner_id, ReleaseMode::LockIfLastIndex), eq Ok(ReleaseState::Locked));
    assert_that!(sut.is_locked(), eq true);
```

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
```diff
@@ -0,0 +1,644 @@
...
+impl OwnerId {
+    const EMPTY: OwnerId = OwnerId(0);
+
+    pub fn new(value: u64) -> Result<Self, OwnerIdNewError> {
+        if value == Self::EMPTY.0 {
+            fail!(from "OwnerId::new()", with OwnerIdNewError::InvalidOwnerValue,
+                "Invalid value for OwnerId. The OwnerId cannot be 0.");
+        }
+
+        Ok(Self(value))
```
elBoberido, line 140: As you know, I'm not a big fan of surrogates, but if an empty state needs to be encoded with a 0, the 0 should not be a valid parameter. Please consider using `NonZero<u64>`


```suggestion
    const EMPTY: OwnerId = OwnerId(0);

    pub fn new(value: NonZero<u64>) -> Self {
        Self(value.get())
```

One could think about using an assert to check if the value is really zero in case someone uses `NonZero::new_unchecked(0)`

iceoryx2-bb/lock-free/tests-common/src/mpmc_robust_unique_index_set_tests.rs
```diff
@@ -0,0 +1,366 @@
...
+pub fn recover_locks_the_set_if_release_mode_is_lock() {
+    let sut = StaticRobustUniqueIndexSet::<CAPACITY>::new();
+
+    let owner_id = OwnerId::new(558).unwrap();
+    for _ in 0..CAPACITY {
+        sut.acquire(owner_id).unwrap();
+    }
+
+    assert_that!(sut.recover(ReleaseMode::LockIfLastIndex, |id| id == owner_id), eq ReleaseState::Locked);
+    assert_that!(sut.is_locked(), eq true);
```
elBoberido, line 243: Should thee be a check if the set is now empty?

iceoryx2-bb/lock-free/tests-common/src/mpmc_robust_unique_index_set_tests.rs
```diff
@@ -0,0 +1,366 @@
...
+
+    let owner_id = OwnerId::new(123).unwrap();
+    let mut indices = vec![];
+    for n in 0..CAPACITY {
+        assert_that!(sut.borrowed_indices(), eq n);
+        indices.push(unsafe { sut.acquire(owner_id).unwrap() });
+        assert_that!(sut.borrowed_indices(), eq n + 1);
+    }
+
+    for n in (1..CAPACITY + 1).rev() {
```
elBoberido: Here, `0..CAPACITY` would also match more closely to the loop above

iceoryx2-bb/lock-free/tests-common/src/mpmc_robust_unique_index_set_tests.rs
```diff
@@ -0,0 +1,366 @@
...
+    assert_that!(
+        sut.release(index, owner_id, ReleaseMode::LockIfLastIndex),
+        is_ok
+    );
+
+    assert_that!(sut.recover(ReleaseMode::Default, |id| id == owner_id), eq ReleaseState::Locked);
+}
+
+#[test]
+pub fn acquire_and_release_works_with_uninitialized_memory() {
```
elBoberido: The name is a bit off. When acquire and release are used, the memory is initialized ... twice.

iceoryx2-bb/lock-free/tests-common/src/mpmc_robust_unique_index_set_tests.rs
```diff
@@ -0,0 +1,366 @@
...
+
+    // check if the sut is still in an consistent state
+    let mut ids = vec![];
+    let mut id_counter = [0u64; CAPACITY];
+
+    for id in id_counter.iter_mut().take(CAPACITY) {
+        let e = sut.acquire(owner_id);
+        assert_that!(e, is_ok);
+        *id += 1;
+        ids.push(e.unwrap());
```
elBoberido, line 360: The `ids` are never checked

elBoberido (commented):
Need a fresh mind to review the actual implementation. Everything else is reviewed

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
elfenpiff: The surrogates are unavoidable in the inner implementation because of the atomics. I would not use `NonZero` here for two reasons:

1. `OwnerId` does exactly what `NonZero` does but directly in place when `OwnerId` is constructed
2. The `OwnerId` contract might be extended in the future, maybe `u64::MAX` is added as another surrogate type. With your API, it would break the API; in my case, one could extend this easily.

elBoberido: Fine. But it is nowhere documented that 0 is not allowed.

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
```diff
@@ -0,0 +1,662 @@
...
+            }
+        };
+
+        unsafe {
+            self.data_ptr.init(fail!(from self,
+                when allocator.allocate(layout),
+                "Failed to initialize since the allocation of the data memory failed."))
+        };
+
+        for i in 0..self.capacity + 1 {
```
elBoberido:
```suggestion
        for i in 0..=self.capacity {
```

We should always use the `..=` syntax instead of adding 1. The former cannot overflow. In this specific case, there will also not be an overflow but just to have this in muscle memory.

But I'm also not sure if this code is correct. Shouldn't it be just `0..self.capacity()`?

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
```diff
@@ -0,0 +1,662 @@
...
+            return ReleaseState::Locked;
+        }
+
+        loop {
+            let state = self.repair_borrowed_indices();
+
+            if state.borrowed_indices == 0 {
+                match self.generation_counter.compare_exchange(
+                    state.generation_counter,
+                    u64::MAX,
```
elBoberido: Please replace this e.g. with `Self::GENERATION_COUNTER_LOCK_INDICATOR` or similar with
```rust
impl RobustUniqueIndexSet {

const GENERATION_COUNTER_LOCK_INDICATOR: u64 = u64::MAX;
// ...
}
```

Please also use it in the `is_locked` method.

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
```diff
@@ -0,0 +1,662 @@
...
+    }
+
+    /// Returns the capacity of the [`RobustUniqueIndexSet`].
+    pub fn capacity(&self) -> usize {
+        self.capacity
+    }
+
+    /// The compile time version of [`RobustUniqueIndexSet::memory_size()`]
+    pub const fn const_memory_size(capacity: usize) -> usize {
+        core::mem::size_of::<AtomicU64>() * capacity + core::mem::align_of::<AtomicU64>() - 1
```
elBoberido, line 263: Either the `init` function is wrong or this needs to be `capacity + 1`.

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
```diff
@@ -0,0 +1,662 @@
...
+            //////////////////////////////////////
+            let current_generation_count = self.generation_counter.load(Ordering::Acquire);
+
+            if current_generation_count == u64::MAX {
+                fail!(from self, with UniqueIndexSetAcquireFailure::IsLocked,
+                    "{msg} since the RobustUniqueIndexSet is locked.");
+            }
+
+            for n in 0..self.capacity {
+                let cell = unsafe { &*self.data_ptr.as_ptr().add(n) };
```
elBoberido: A small optimization is possible by moving the conversion of the relative pointer to an absolute pointer out of the loop.

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
```diff
@@ -0,0 +1,662 @@
...
+
+                match cell.compare_exchange(
+                    OwnerId::EMPTY.0,
+                    owner_id.0,
+                    Ordering::Relaxed,
+                    Ordering::Relaxed,
+                ) {
+                    Ok(_) => {
+                        self.borrowed_indices.fetch_add(1, Ordering::Relaxed);
```
elBoberido: If the application is crashing before this is finished, the borrowed_indices count is out of sync.

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
```diff
@@ -0,0 +1,662 @@
...
+
+                        //////////////////////////////////////
+                        // SYNC POINT enforce order of:
+                        //   1. cell & borrowed_indices
+                        //   2. generation counter
+                        //////////////////////////////////////
+                        if self.increment_generation_counter(Ordering::Release)
+                            == ReleaseState::Locked
```
elBoberido: Worst case, a owner id could have been written to a locked index set.

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
```diff
@@ -0,0 +1,662 @@
...
+                        {
+                            // revert the increment, and the reservation not important, just for consistency
+                            self.borrowed_indices.fetch_sub(1, Ordering::Relaxed);
+                            cell.store(OwnerId::EMPTY.0, Ordering::Relaxed);
```
elBoberido: If the application crashes before these two statements, the reverting is not done. If it crashes in between these two statements, the borrowed_indices count is out of sync

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
```diff
@@ -0,0 +1,662 @@
...
+                ordering,
+                Ordering::Relaxed,
+            ) {
+                Ok(_) => return ReleaseState::Unlocked,
+                Err(v) => current_generation_count = v,
+            }
+        }
+    }
+
+    fn repair_borrowed_indices(&self) -> RepairState {
```
elBoberido: It's really hard to tell what this is supposed to do. Without a description I cannot tell if the implementation is correct.

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
```diff
@@ -0,0 +1,662 @@
...
+
+            let mut count = 0;
+            for n in 0..self.capacity {
+                let cell = unsafe { &*self.data_ptr.as_ptr().add(n) };
+                if cell.load(Ordering::Relaxed) != OwnerId::EMPTY.0 {
+                    count += 1;
+                }
+            }
+
+            self.borrowed_indices.store(count, Ordering::Relaxed);
```
elBoberido: Isn't this racy? Another entity could just increment/decrement the `borrowed_indices`

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
```diff
@@ -0,0 +1,662 @@
...
+
+            //////////////////////////////////////
+            // SYNC POINT enforce order of:
+            //   1. borrowed_indices
+            //   2. generation counter
+            //////////////////////////////////////
+            self.increment_generation_counter(Ordering::Release);
+
+            if current_generation_count + 1 == self.generation_counter.load(Ordering::Relaxed) {
```
elBoberido: This check is not enough to ensure that the `borrowed_indices` was not mutated in between the calls to `generate_counter`. For example, the `acquire` could already have incremented the generation counter before `repair_borrowed_indices` was called, and suspended before `borrowed_indices` was incremented but after the cell was written with the new owner id. Now the `acquire` continues and increments a `borrowed_indices` value which is out of sync.

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
```diff
@@ -0,0 +1,662 @@
...
+        let cell = unsafe { &*self.data_ptr.as_ptr().add(index) };
+
+        match cell.compare_exchange(
+            owner_id.0,
+            OwnerId::EMPTY.0,
+            Ordering::Relaxed,
+            Ordering::Relaxed,
+        ) {
+            Ok(_) => {
+                self.borrowed_indices.fetch_sub(1, Ordering::Relaxed);
```
elBoberido: This is racing with the `self.borrowed_indices.store(count, Ordering::Relaxed);` in `repair_borrowed_indices`. After the CAS was successful and assuming the `borrowed_indices` value was one, then the store in `repair_borrowed_indices` will already have set the value to 0 if if was called right after the CAS and before this `fetch_sub`. The `fetch_sub` would therefore cause an underflow.

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
```diff
@@ -0,0 +1,662 @@
...
+            }
+        }
+    }
+
+    /// Returns the current len.
+    pub fn borrowed_indices(&self) -> usize {
+        if self.is_locked() {
+            0
+        } else {
+            self.borrowed_indices.load(Ordering::Relaxed) as _
```
elBoberido: The `borrowed_indices` value is not used in this module and can be out of sync with the actual count. Worst case, a value of 0 could be returned even though, there is a valid active index borrowed.

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
```diff
@@ -0,0 +1,662 @@
...
+            let value = cell.load(Ordering::Relaxed);
+
+            if value == OwnerId::EMPTY.0 {
+                continue;
+            }
+
+            if predicate(OwnerId(value))
+                && cell
+                    .compare_exchange(
+                        value,
```
elBoberido, line 427: So, when the CAS fails, the step is skipped. Should be fine. When the value changed, the cell is occupied with a different ID.

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
```diff
@@ -0,0 +1,662 @@
...
+                //////////////////////////////////////
+                if self.increment_generation_counter(Ordering::Release) == ReleaseState::Locked {
+                    return ReleaseState::Locked;
+                }
+
+                if mode == ReleaseMode::LockIfLastIndex {
+                    self.lock();
+                } else {
+                    // must be called here explicitly since self.lock() calls this
+                    // already
```
elBoberido: I don't understand this comment. The `lock` call is in a different branch and cannot `call this already`

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
elfenpiff: init function is wrong

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
elfenpiff: The variable has been removed.

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
elfenpiff: removed the borrowed_indices member completely and also this call.

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
elfenpiff: `borrowed_indices` is completely removed.

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
elfenpiff: `repair_borrowed_indices` is renamed to `borrowed_indices_and_generation_counter` and has a detailed comment.

furthermore the `borrowed_indices` variable is removed.

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
elfenpiff: `borrowed_indices` was removed

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
elfenpiff: the `borrowed_indices` variable is removed.

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
elfenpiff: I removed the variable

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
```diff
@@ -0,0 +1,664 @@
...
+            }
+
+            //////////////////////////////////////
+            // SYNC POINT enforce order of:
+            //   1. cell
+            //   2. generation counter
+            //////////////////////////////////////
+            let new_generation_count = self.increment_generation_counter(Ordering::Release);
+
+            if current_generation_count + 1 == new_generation_count {
```
elBoberido, line 521: Hehe, again the naming. Only if you if there is another commit required. The word current is dependent on the time and when one looks at it at a later point, the value at the current time does not match the value at the old current time. For this, to prevent ambiguities, I suggest to use something like `generation_count_at_begin` or `initial_generation_count`. Alternatively, having a `let expected_generation_count = current_generation_count + 1;` right before the `increment_generation_counter` call, would also reduce the cognitive load on this line.

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
```diff
@@ -0,0 +1,664 @@
...
+                return SetState {
+                    generation_counter: current_generation_count,
+                    borrowed_indices: 0,
+                };
+            }
+
+            let mut count = 0;
+            for n in 0..self.capacity {
+                let cell = unsafe { &*cell_ptr.add(n) };
+                if cell.load(Ordering::Relaxed) != OwnerId::EMPTY.0 {
```
elBoberido, line 509: I somehow have the feeling that I read somewhere that a `relaxed` load is not guaranteed to get the latest value, just not to get a value that is older than the one from from the last `relaxed` load and the cored do not necessarily have updated the value.

But the `Acquire` from the load of the generation counter should ensure a memory synchronization and since the operation is repeated if the generation count changed, the `relaxed` load should be fine in this context.

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
```diff
@@ -0,0 +1,664 @@
...
+            //////////////////////////////////////
+            // SYNC POINT enforce order of:
+            //   1. cell
+            //   2. generation counter
+            //////////////////////////////////////
+            let new_generation_count = self.increment_generation_counter(Ordering::Release);
+
+            if current_generation_count + 1 == new_generation_count {
+                return SetState {
+                    generation_counter: current_generation_count + 1,
```
elBoberido, line 523: Also nitpicking and just if there are other changes required or with another PR. It would make more sense to use `new_generation_count` instead of `current_generation_count + 1`. The result is the same, but the cognitive load reading the code will be reduced.

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
```diff
@@ -0,0 +1,664 @@
...
+        debug_assert!(
+            self.is_memory_initialized.load(Ordering::Relaxed),
+            "Undefined behavior when calling RobustUniqueIndexSet::{source} and the object is not initialized."
+        );
+    }
+
+    /// Returns if the [`RobustUniqueIndexSet`] is locked or not. If the set is locked no more indices
+    /// can be borrowed.
+    pub fn is_locked(&self) -> bool {
+        self.generation_counter.load(Ordering::Relaxed) == GENERATION_COUNTER_LOCK_INDICATOR
```
elBoberido, line 222: This might not be sufficient. Like already mentioned in the other comment, the load might not get the latest value. But since this call is inherently racy and it could be locked the next nanosecond after the call, it should be fine. When someone would try to mutate the index set with the outdated information, the operation would fail and no harm is done, especially since there is no unlock.

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
```diff
@@ -0,0 +1,664 @@
...
+                            fail!(from self, with UniqueIndexSetAcquireFailure::IsLocked,
+                                "{msg} since the RobustUniqueIndexSet is locked.");
+                        }
+                        return Ok(n);
+                    }
+                    Err(_) => continue,
+                }
+            }
+
+            if current_generation_count == self.generation_counter.load(Ordering::Relaxed) {
```
elBoberido, line 342: Here I would need to check if the relaxed load is sufficient. It might be possible that the load can happen before the for loop, since there is no way that this instruction is reached with an increment of the counter from the same thread.

On the other hand, this is also inherently racy and another thread could change the generation count right after this check. What do you think, should this be made `SeqCst`?

Ideally, this would also be a `Release` ordering, but that does not work with load.

elBoberido (approved):

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
```diff
@@ -0,0 +1,664 @@
...
+                continue;
+            }
+
+            if predicate(OwnerId(value))
+                && cell
+                    .compare_exchange(
+                        value,
+                        OwnerId::EMPTY.0,
+                        Ordering::Relaxed,
+                        Ordering::Relaxed,
```
elBoberido, line 430: Maybe one last food for thought. Should `cell.load` have `Ordering::Acquire` and `compare_exchange` have `Ordering::Release` when successful and `Ordering::Acquire` otherwise?

elBoberido: cc @FerdinandSpitzschnueffler

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
elfenpiff: nope, also with acquire you do not get the newest one. the guarantee is that you get all other memory changes that happen before the atomic was set to the value you have read.

elfenpiff: In this context, the generation_counter.load synced everything

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
elfenpiff: > On the other hand, this is also inherently racy and another thread could change the generation count right after this check.

Actually, this is not a problem. Lets assume this would happen, how would it differ to the scenarion where the generation counter changed as soon as the function returned to the user.

iceoryx2-bb/lock-free/src/mpmc/robust_unique_index_set.rs
elfenpiff: But why.

We do not sync here any memory. We need to ensure that compare exchange comes after the cell load but since it is a dependent operation on value, the return value of cell.load, it must come before

elBoberido: Because we might not get the latest value with the latest relaxed load but only one which is not older than the last time it was read on the same thread. At least I think I have read something in this regard but my memories are blurry. I could be wrong on this. If I'm not wrong, we might not release an index due to the load giving us an oudated value. Since there are other sync point, even if I'm not wrong, this might be only hypothetical.

elfenpiff: > Because we might not get the latest value with the latest relaxed load but only one which is not older than the last time it was read on the same thread.

I agree, this is correct.

> If I'm not wrong, we might not release an index due to the load giving us an oudated value.

This is correct. But this can also happen when the index is invalidated right after the load. In those cases the user has to just retry. And the important thing is, since it is a multithreaded program, from the outside the user could not observe any difference.

Except when they use a surrounding sync mechanism, like we do with the file locks.

PR #1615, opened 2026-05-11
[#1578] integrate robust unique index set into container

elfenpiff:

## PR Reviewer Reminders

* Public API is documented

## References

Relates to #1578

### DIFF
--- a/iceoryx2-bb/concurrency/src/cell.rs
+++ b/iceoryx2-bb/concurrency/src/cell.rs
@@ -41,6 +41,13 @@ impl<T> Cell<T> {
         self.0.as_ptr()
     }
 
+    pub fn get(&self) -> T
+    where
+        T: Copy,
+    {
+        unsafe { *self.0.as_ptr() }
+    }
+
     pub fn set(&self, value: T) {
         self.0.set(value)
     }
--- a/iceoryx2-bb/concurrency/src/cell.rs
+++ b/iceoryx2-bb/concurrency/src/cell.rs
@@ -45,7 +45,7 @@ impl<T> Cell<T> {
     where
         T: Copy,
     {
-        unsafe { *self.0.as_ptr() }
+        self.0.get()
     }
 
     pub fn set(&self, value: T) {
--- a/iceoryx2-bb/lock-free/src/mpmc/container.rs
+++ b/iceoryx2-bb/lock-free/src/mpmc/container.rs
@@ -321,9 +321,6 @@ impl<T: Copy + Debug> Container<T> {
 
             let element_generation_counter =
                 &*self.element_generation_counter_ptr.as_ptr().add(index as _);
-            //////////////////////////////////////
-            // SYNC POINT with reading data values
-            //////////////////////////////////////
             let current_generation_count = element_generation_counter.load(Ordering::Relaxed);
 
             // Another process died and the entry was released by `Self::recover()` but the
@@ -334,6 +331,10 @@ impl<T: Copy + Debug> Container<T> {
             // contains data. If this succeeds we won the race.
             //
             // If this fails the entry was released successfully and we need to increment it by 1.
+
+            //////////////////////////////////////
+            // SYNC POINT all atomic operations with reading data values
+            //////////////////////////////////////
             if Self::contains_data(current_generation_count) {
                 if element_generation_counter
                     .compare_exchange(
@@ -416,7 +417,10 @@ impl<T: Copy + Debug> Container<T> {
         state
     }
 
-    /// Recovers and releases all entries the dead [`OwnerId`] owned.
+    /// Recovers and releases all entries the dead [`OwnerId`] owned. It assumes that the dead owner
+    /// maybe died while adding some entry, therefore it removes all entries where the
+    /// [`OwnerId`] does not contain any data or where there was data and the provided predicate
+    /// returned [`true`].
     ///
     /// # Safety
     ///
@@ -424,8 +428,9 @@ impl<T: Copy + Debug> Container<T> {
     ///  * All existing [`ContainerHandle`] that belong to the [`OwnerId`] must never be removed with
     ///    [`Container::remove()`] otherwise we corrupt the state.
     ///
-    pub unsafe fn recover<F: FnMut(OwnerId, T) -> bool>(
+    pub unsafe fn recover<F: FnMut(T) -> bool>(
         &self,
+        dead_owner_id: OwnerId,
         mut predicate: F,
         mode: ReleaseMode,
     ) -> ReleaseState {
@@ -435,14 +440,21 @@ impl<T: Copy + Debug> Container<T> {
         // once it the predicate to store the generation count before the entry is released and
         // afterwards to set the entry to empty if no other thread has in the meantime acquired the new
         // index and repaired it
-        let current_generation_count = RefCell::new(0u64);
+        let current_element_generation_count = RefCell::new(0u64);
 
         let p = |owner_id, index| {
             let element_generation_counter =
                 unsafe { &*self.element_generation_counter_ptr.as_ptr().add(index) };
 
+            if dead_owner_id != owner_id {
+                return false;
+            }
+
             loop {
-                let gen_count = element_generation_counter.load(Ordering::Relaxed);
+                //////////////////////////////////////
+                // SYNC POINT with reading data values
+                //////////////////////////////////////
+                let gen_count = element_generation_counter.load(Ordering::Acquire);
 
                 if Self::contains_data(gen_count) {
                     let mut contents = MaybeUninit::<T>::uninit();
@@ -458,17 +470,17 @@ impl<T: Copy + Debug> Container<T> {
                         continue;
                     }
 
-                    *current_generation_count.borrow_mut() = gen_count;
+                    *current_element_generation_count.borrow_mut() = gen_count;
 
-                    return predicate(owner_id, unsafe { contents.assume_init_read() });
+                    return predicate(unsafe { contents.assume_init_read() });
                 } else {
-                    return false;
+                    return true;
                 }
             }
         };
 
         let on_success = |_owner_id, index| {
-            let v = *current_generation_count.borrow();
+            let v = *current_element_generation_count.borrow();
 
             // If setting the entry to empty failed then an other thread won the `Self::add()` race,
             // acquired the next index and already repaired it.
@@ -717,11 +729,12 @@ impl<T: Copy + Debug, const CAPACITY: usize> FixedSizeContainer<T, CAPACITY> {
     ///  * All existing [`ContainerHandle`] that belong to the [`OwnerId`] must never be removed with
     ///    [`Container::remove()`] otherwise we corrupt the state.
     ///
-    pub unsafe fn recover<F: FnMut(OwnerId, T) -> bool>(
+    pub unsafe fn recover<F: FnMut(T) -> bool>(
         &self,
+        dead_owner_id: OwnerId,
         predicate: F,
         mode: ReleaseMode,
     ) -> ReleaseState {
-        unsafe { self.container.recover(predicate, mode) }
+        unsafe { self.container.recover(dead_owner_id, predicate, mode) }
     }
 }
--- a/iceoryx2-bb/lock-free/src/mpmc/container.rs
+++ b/iceoryx2-bb/lock-free/src/mpmc/container.rs
@@ -318,25 +318,27 @@ impl<T: Copy + Debug> Container<T> {
 
             let element_generation_counter =
                 &*self.element_generation_counter_ptr.as_ptr().add(index as _);
-            let current_generation_count = element_generation_counter.load(Ordering::Relaxed);
+            let current_element_generation_count =
+                element_generation_counter.load(Ordering::Acquire);
 
-            // Another process died and the entry was released by `Self::recover()` but the
+            // Another process died and the entry was released by `Self::recover()` but the element
             // generation counter has not yet marked the element as free. Now we race with the
-            // recovery part.
+            // recovery part. This is the only part that is allowed to set the
+            // element_generation_counter from contains_data to free since the index set ensures that
+            // we are the only owner of that element.
             //
             // We increment the generation count by 2 to signal the next generation of an entry that
             // contains data. If this succeeds we won the race.
             //
             // If this fails the entry was released successfully and we need to increment it by 1.
-
             //////////////////////////////////////
             // SYNC POINT all atomic operations with reading data values
             //////////////////////////////////////
-            if Self::contains_data(current_generation_count) {
+            if Self::contains_data(current_element_generation_count) {
                 if element_generation_counter
                     .compare_exchange(
-                        current_generation_count,
-                        current_generation_count + 2,
+                        current_element_generation_count,
+                        current_element_generation_count + 2,
                         Ordering::Release,
                         Ordering::Relaxed,
                     )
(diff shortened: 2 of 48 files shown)


iceoryx2-bb/lock-free/src/mpmc/container.rs
```diff
@@ -374,6 +414,80 @@ impl<T: Copy + Debug> Container<T> {
...
+                    let mut contents = MaybeUninit::<T>::uninit();
+                    unsafe {
+                        core::ptr::copy_nonoverlapping(
+                            (*self.data_ptr.as_ptr().add(index)).get(),
+                            contents.as_mut_ptr().cast(),
+                            1,
+                        )
+                    };
+
+                    if gen_count != element_generation_counter.load(Ordering::Relaxed) {
```
FerdinandSpitzschnueffler: Shouldnt this be Release cas, copy_nonoverlapping must happen before (as-if rule?) element generation count reading

iceoryx2-cxx/include/iox2/config.hpp
```diff
@@ -125,6 +120,11 @@ class Global {
     auto root_path() && -> const char*;
     /// Defines the path under which all other directories or files will be created
     void set_root_path(const iox2::bb::Path& value) &&;
+    /// Defines the time of how long another process will wait until an entity creation
+    /// is finished. An entity could be a Node or a Service
+    auto creation_timeout() && -> iox2::bb::Duration;
+    /// Set the creation timeout
+    void set_creation_timeout(const iox2::bb::Duration& value) &&;
```
elBoberido, line 131: Can this be considered a breaking change?

iceoryx2-bb/lock-free/src/spmc/unrestricted_atomic.rs
```diff
@@ -210,21 +210,7 @@ impl UnrestrictedAtomicMgmt {
...
-            // prevent reordering of reading from `data` after checking for a change
-            // of the `write_cell` position which would result in a data race
-            /////////////////////////
-            let expected_write_cell = read_cell + 1;
-            let write_cell_result = self.write_cell.compare_exchange(
-                expected_write_cell,
-                expected_write_cell,
-                Ordering::Release,
-                Ordering::Acquire,
-            );
```
elBoberido, line 224: I would keep this code. AFAIK, the compiler is not allowed to optimize this away. Although the call will not change the value of the atomic, it has to load and return the actual value of the atomic, since that value is used further down.

iceoryx2-bb/lock-free/src/mpmc/container.rs
```diff
@@ -403,41 +515,36 @@ impl<T: Copy + Debug> Container<T> {
         previous_state.current_change_counter = current_change_counter;
 
         for i in 0..self.capacity {
+            let element_generation_counter =
+                unsafe { &*self.element_generation_counter_ptr.as_ptr().add(i) };
```
elBoberido: nitpicking: opportunity for small optimization. Since the relativ address of the counter does not change, the conversion to an absolute pointer needs to be done only once.

iceoryx2-bb/lock-free/src/mpmc/container.rs
```diff
@@ -403,41 +515,36 @@ impl<T: Copy + Debug> Container<T> {
...
                             1,
                         );
                     }
 
-                    // MUST HAPPEN AFTER all other operations
-                    if let Err(count) = (*self.active_index_ptr.as_ptr().add(i)).compare_exchange(
-                        current_index_count,
-                        current_index_count,
-                        Ordering::AcqRel,
-                        Ordering::Acquire,
```
elBoberido, line 437: Here, I would also keep this approach..

iceoryx2-bb/lock-free/src/mpmc/container.rs
```diff
@@ -376,6 +414,80 @@ impl<T: Copy + Debug> Container<T> {
...
+        mut predicate: F,
+        mode: ReleaseMode,
+    ) -> ReleaseState {
+        self.verify_init("recover()");
+
+        // this is an atomic since the current generation count needs to be borrowed twice as mutable,
+        // once it the predicate to store the generation count before the entry is released and
+        // afterwards to set the entry to empty if no other thread has in the meantime acquired the new
+        // index and repaired it
+        let current_element_generation_count = RefCell::new(0u64);
```
elBoberido: It seems the comment is outdated. This is a `RefCell`, not an `Atomic`. Have you considered to use just a `Cell`?

One would need to `impl <T: Copy> for Cell<T> { pub fn copy(&self)} -> T` for our `Cell` wrapper.

iceoryx2-bb/lock-free/src/mpmc/container.rs
elBoberido: It is the same in `update_state`

iceoryx2-bb/lock-free/src/mpmc/container.rs
```diff
@@ -305,25 +301,58 @@ impl<T: Copy + Debug> Container<T> {
...
+            let index = self.index_set.acquire(owner_id)?;
             core::ptr::copy_nonoverlapping(
                 &value,
                 (*self.data_ptr.as_ptr().add(index as _)).get().cast(),
                 1,
             );
 
+            let element_generation_counter =
+                &*self.element_generation_counter_ptr.as_ptr().add(index as _);
+            let current_generation_count = element_generation_counter.load(Ordering::Relaxed);
```
elBoberido: I need to check this, but it might be that we need `Ordering::Acquire` here.

iceoryx2-bb/lock-free/src/mpmc/container.rs
```diff
@@ -349,10 +378,19 @@ impl<T: Copy + Debug> Container<T> {
...
 
-        unsafe { &*self.active_index_ptr.as_ptr().add(handle.index as _) }
-            .fetch_add(1, Ordering::Relaxed);
+        unsafe {
+            &*self
+                .element_generation_counter_ptr
+                .as_ptr()
+                .add(handle.index as _)
+        }
+        .fetch_add(1, Ordering::Relaxed);
```
elBoberido: Same here. This might also need to be `Ordering::Acquire`

elBoberido (commented):
Looks good in general but I need a fresh mind for the lock-free algorithm

iceoryx2-bb/lock-free/src/mpmc/container.rs
```diff
@@ -305,25 +301,58 @@ impl<T: Copy + Debug> Container<T> {
...
+                        current_generation_count + 2,
+                        Ordering::Release,
+                        Ordering::Relaxed,
+                    )
+                    .is_err()
+                {
+                    element_generation_counter.fetch_add(1, Ordering::Release);
+                }
+            } else {
+                element_generation_counter.fetch_add(1, Ordering::Release);
```
elBoberido: Don't we need also do a CAS here? Isn't it possible that `contains_data` is false at the time of check and before the increment happens, someone else already incremented the counter, resulting in a wrong state?

iceoryx2-bb/lock-free/src/mpmc/container.rs
elfenpiff: done

iceoryx2-bb/concurrency/src/cell.rs
```diff
@@ -41,6 +41,13 @@ impl<T> Cell<T> {
         self.0.as_ptr()
     }
 
+    pub fn get(&self) -> T
+    where
+        T: Copy,
+    {
+        unsafe { *self.0.as_ptr() }
```
elBoberido: Why unsafe?
```suggestion
        self.0.get()
```

iceoryx2-bb/lock-free/src/spmc/unrestricted_atomic.rs
```diff
@@ -194,12 +194,12 @@ impl UnrestrictedAtomicMgmt {
...
-        // SYNC POINT - read
-        /////////////////////////
-        let mut read_cell = self.write_cell.load(Ordering::Acquire) - 1;
-
         loop {
+            /////////////////////////
+            // SYNC POINT - read
+            /////////////////////////
+            let read_cell = self.write_cell.load(Ordering::Acquire) - 1;
+
```
elBoberido, line 204: There are quire a few `- 1` and `+ 1`. This removes some of the semantics since it mixes read and write terms. What do you think about something like this

```suggestion

        /////////////////////////
        // SYNC POINT - read
        /////////////////////////
        let mut current_write_cell = self.write_cell.load(Ordering::Acquire);
        
        loop {
            let read_cell = current_write_cell - 1;

```

and further down

```rust
            if let Err(new_write_cell) = self.write_cell.compare_exchange(
                current_write_cell,
                current_write_cell,
                Ordering::Release,
                Ordering::Acquire,
            ) {
                current_write_cell = new_write_cell;
            } else {
                break;
            }
```

The `if read_cell + 1 == ...` can be removed.

elBoberido, line 204: There are quire a few `- 1` and `+ 1`. This removes some of the semantics since it mixes read and write terms. What do you think about something like this

```suggestion

        /////////////////////////
        // SYNC POINT - read
        /////////////////////////
        let mut current_write_cell = self.write_cell.load(Ordering::Acquire);
        
        loop {
            let read_cell = current_write_cell - 1;

```

and further down

```rust
            if let Err(new_write_cell) = self.write_cell.compare_exchange(
                current_write_cell,
                current_write_cell,
                Ordering::Release,
                Ordering::Acquire,
            ) {
                current_write_cell = new_write_cell;
            } else {
                break;
            }
```

The `if read_cell + 1 == ...` can be removed.

iceoryx2-cxx/include/iox2/config.hpp
elBoberido: Do you want to add this to the release notes?

iceoryx2-bb/lock-free/src/mpmc/container.rs
elfenpiff: No, this can never happen. The index set ensures that I am the only owner of that index. So no one is allowed to operate on that index.

But since remove and recover release first the index to the unique index set and then increment the element generation counter we have a race here. So it could be that we require the index, before the element generation counter was incremented and the element is still in the old contains data state.
But once it is released it can never turn back. I add some comments

iceoryx2-bb/concurrency/src/cell.rs
elfenpiff: Since get is not available in `core`

iceoryx2-bb/lock-free/src/spmc/unrestricted_atomic.rs
elfenpiff: I removed the compare_exchange and replaced it with the suggestion from @ComicSansMS `load(SeqCst)` the other suggestions I applied.

iceoryx2-bb/concurrency/src/cell.rs
elBoberido: `get` is available since Rust 1.0. With Rust 1.88 it was made `const`, but that does not matter for our use case.

iceoryx2-bb/lock-free/src/spmc/unrestricted_atomic.rs
```diff
@@ -210,21 +211,7 @@ impl UnrestrictedAtomicMgmt {
...
-            let write_cell_result = self.write_cell.compare_exchange(
-                expected_write_cell,
-                expected_write_cell,
-                Ordering::Release,
-                Ordering::Acquire,
-            );
-            if let Err(write_cell) = write_cell_result {
-                read_cell = write_cell - 1;
-            } else {
+            if current_write_cell == self.write_cell.load(Ordering::SeqCst) {
```
elBoberido: Food for thought. Would it make sense to have something like this?

```suggestion
            let old_write_cell = current_write_cell;
            let current_write_cell = self.write_cell.load(Ordering::SeqCst);
            if old_write_cell == current_write_cell {
```

and also move the initial load out of the loop?

With the changed algorithm it might also make sense to change the name of `current_write_cell` to something else, e.g. `write_cell_local`. But up to you if you want to change it.

iceoryx2/src/service/dynamic_config/mod.rs
```diff
@@ -202,13 +202,17 @@ impl DynamicConfig {
...
+        &self,
+        handle: ContainerHandle,
+    ) -> Result<DeregisterNodeState, ContainerRemoveError> {
+        match unsafe { self.nodes.remove(handle, ReleaseMode::LockIfLastIndex) } {
+            Ok(ReleaseState::Locked) => Ok(DeregisterNodeState::NoMoreOwners),
+            Ok(ReleaseState::Unlocked) => Ok(DeregisterNodeState::HasOwners),
+            Err(e) => {
+                fail!(from self, with e,
+                    "Unable to deregister the node since it was not registered.");
+            }
```
elBoberido: Is it ensured that this is the only case the error can occur and we don't end up leaving something behind?

iceoryx2-bb/lock-free/src/mpmc/container.rs
```diff
@@ -305,25 +309,60 @@ impl<T: Copy + Debug> Container<T> {
...
+            // Another process died and the entry was released by `Self::recover()` but the element
+            // generation counter has not yet marked the element as free. Now we race with the
+            // recovery part. This is the only part that is allowed to set the
+            // element_generation_counter from contains_data to free since the index set ensures that
+            // we are the only owner of that element.
+            //
+            // We increment the generation count by 2 to signal the next generation of an entry that
+            // contains data. If this succeeds we won the race.
+            //
+            // If this fails the entry was released successfully and we need to increment it by 1.
```
elBoberido: I don't understand the logic behind this.

iceoryx2-bb/lock-free/src/mpmc/container.rs
```diff
@@ -376,6 +436,80 @@ impl<T: Copy + Debug> Container<T> {
...
+                    let mut contents = MaybeUninit::<T>::uninit();
+                    unsafe {
+                        core::ptr::copy_nonoverlapping(
+                            (*self.data_ptr.as_ptr().add(index)).get(),
+                            contents.as_mut_ptr().cast(),
+                            1,
+                        )
+                    };
+
+                    if gen_count != element_generation_counter.load(Ordering::SeqCst) {
```
elBoberido: Similar to the unrestricted atomic, this could be slightly optimized. In this case it should not matter that much, since this will not be in the hot path.

iceoryx2-bb/lock-free/src/mpmc/container.rs
```diff
@@ -376,6 +436,80 @@ impl<T: Copy + Debug> Container<T> {
...
+
+                    return predicate(unsafe { contents.assume_init_read() });
+                } else {
+                    return true;
+                }
+            }
+        };
+
+        let on_success = |_owner_id, index| {
+            let v = current_element_generation_count.get();
```
elBoberido, line 513: It is possible that `on_success` is called without `current_element_generation_count.set(gen_count);` being set. Is this a problem?

It also makes me a little bit nervous that the count could get out of sync with the `count % 2` requirement.

iceoryx2-bb/lock-free/src/mpmc/container.rs
```diff
@@ -401,43 +535,38 @@ impl<T: Copy + Debug> Container<T> {
...
-                    // MUST HAPPEN AFTER all other operations
-                    if let Err(count) = (*self.active_index_ptr.as_ptr().add(i)).compare_exchange(
-                        current_index_count,
-                        current_index_count,
-                        Ordering::AcqRel,
-                        Ordering::Acquire,
-                    ) {
-                        current_index_count = count
-                    } else {
+                    if current_generation_count == element_generation_counter.load(Ordering::SeqCst)
```
elBoberido: Same here with the load. Potential for small optimization. But not necessary since not in the hot path.

iceoryx2/src/service/dynamic_config/mod.rs
elfenpiff: I made it more explicit

iceoryx2-bb/lock-free/src/mpmc/container.rs
elfenpiff: No, because it will be always set in the predicate, some lines above and the return value of the predicate defines if the element is removed or not and if this is successful, `on_success` is called.

So `predicate` is called always and when it returns `true` and the removal was successful, then `on_success` is called.

elBoberido: Yes, `on_success` is only called if `p` returns true, but there is a path where `p` return true without calling `current_element_generation_count.set(gen_count);`. But in that case, the `compare_exchange(v, v + 1, ...) ` will fail.

elBoberido (approved):

PR #1630, opened 2026-05-16
[#1629] Fix segfault when connection is removed

### LINKED ISSUE #1629: Segfault when accessing data of a removed connection
## Required information

**Operating system:**

n/a

**Rust version:**

n/a

**Cargo version:**

n/a

**iceoryx2 version:**

v0.7 and newer

**Detailed log output:**

n/a

Accessing a sample of a removed connection will lead to a segfault.

Accessing a sample should never lead to a segfault.

When a connection is removed while there are still active borrows, the shared memory is unmapped and the sample point to invalid memory.

This example reproduces the bug.

```rust
use iceoryx2::prelude::*;

    let node = NodeBuilder::new().create::<ipc::Service>()?;

    let service = node
        .service_builder(&"iox2_clean_drop_reproe".try_into()?)
        .publish_subscribe::<u64>()
        .open_or_create()?;

    let subscriber = service.subscriber_builder().create()?;
    let publisher = Some(service.publisher_builder().create()?);

    let sample = publisher.as_ref().unwrap().loan_uninit()?;
    let sample = sample.write_payload(2);
    sample.send()?;

    let sample = subscriber.receive()?.unwrap();

    coutln!("#### sample: {:?}", sample.payload());

    drop(publisher);

    // this removes the connection and triggers the bug
    let _ = subscriber.receive()?;
    // alternatively, this would also trigger the bug 'subscriber.update_connections();'

    coutln!("#### does it crash?");
    coutln!("#### sample: {:?}", sample.payload()); // SEGFAULT

    coutln!("###############################");
    coutln!("# survived — bug may be fixed #");
    coutln!("###############################");

    Ok(())
}
```

elBoberido:
## Notes for Reviewer

This PR fixes a segfault when a sample is accesses which corresponding connection was already closed.

The issue occurs only with pub-sub and not with request-response.

In order to have better debug information from CI failures, the `Expected` now also prints the error value on an access violation.

## PR Reviewer Reminders

* Public API is documented

## References

Closes #1629
Closes #1528

### DIFF
--- a/iceoryx2/src/port/details/receiver.rs
+++ b/iceoryx2/src/port/details/receiver.rs
@@ -264,28 +264,130 @@ impl<Service: service::Service> Receiver<Service> {
             let connection_storage = unsafe { &mut *self.connection_storage.get() };
 
             if let Some(connection) = connection_storage.get_mut(key) {
-                let mut keep_connection = false;
-                for id in 0..self.number_of_channels {
-                    if connection.receiver.has_data(ChannelId::new(id)) {
-                        keep_connection = true;
-                        break;
-                    }
-                }
+                let receiver = &connection.receiver;
+                let (connection_has_data, connection_has_borrows) =
+                    Self::all_receiver_channel_has_data_and_has_borrows(receiver);
 
+                let keep_connection = connection_has_data | connection_has_borrows;
                 if keep_connection {
-                    if unsafe { &mut *to_be_removed_connections.get() }
-                        .push(key)
-                        .is_err()
-                    {
-                        warn!(from self,
-                            "Expired connection buffer exceeded. A sender disconnected with undelivered data that will be discarded. Increase the expired connection buffer to mitigate the problem.");
-                        connection_storage.remove(key);
+                    let to_be_removed_connections =
+                        unsafe { &mut *to_be_removed_connections.get() };
+                    if to_be_removed_connections.push(key).is_err() {
+                        let msg_begin = "Expired connection buffer exceeded.";
+                        let msg_disconnet_with_data =
+                            "A sender disconnected with undelivered data that will be discarded.";
+                        let msg_disconnet_with_borrows = "A sender disconnected with data that is still borrowed. This will lead to segmentation faults when the data is accessed later on.";
+                        let msg_end =
+                            "Increase the expired connection buffer to mitigate the problem.";
+
+                        // push failed! let's see if there is something that can be removed
+                        let index_and_key_to_remove =
+                            Self::find_connection_without_data_and_borrows(
+                                connection_storage,
+                                to_be_removed_connections,
+                            );
+
+                        if let Some((index, key)) = index_and_key_to_remove {
+                            // we found a connection without data and borrows which can be removed from the container
+                            to_be_removed_connections.remove(index);
+                            connection_storage.remove(key);
+                        } else if connection_has_borrows {
+                            // we did not find a connection that can safely be removed to create some space for the connection with borrows;
+                            // removing a connection with borrows might lead to segfaults -> try to remove a connection without borrows
+                            let index_and_key_to_remove = Self::find_connection_without_data(
+                                connection_storage,
+                                to_be_removed_connections,
+                            );
+
+                            if let Some((index, key)) = index_and_key_to_remove {
+                                // we found a connection without borrows which can be removed from the container
+                                warn!(from self, "{} {} {}", msg_begin, msg_disconnet_with_data, msg_end);
+                                to_be_removed_connections.remove(index);
+                                connection_storage.remove(key);
+                            }
+                        }
+
+                        if to_be_removed_connections.push(key).is_err() {
+                            if connection_has_borrows {
+                                fatal_panic!(from self, "{} {} {}", msg_begin, msg_disconnet_with_borrows, msg_end);
+                            } else {
+                                warn!(from self, "{} {} {}", msg_begin, msg_disconnet_with_data, msg_end);
+                            }
+                            connection_storage.remove(key);
+                        }
                     }
                 } else {
                     connection_storage.remove(key);
                 }
             }
         }
+    }
+
+    fn find_connection_without_data_and_borrows(
+        connection_storage: &SlotMap<Connection<Service>>,
+        connections: &PolymorphicVec<'static, SlotMapKey, HeapAllocator>,
+    ) -> Option<(usize, SlotMapKey)> {
+        Self::find_connection_with_condition(
+            connection_storage,
+            connections,
+            |has_data, has_borrows| !(has_data | has_borrows),
+        )
+    }
+
+    fn find_connection_without_data(
+        connection_storage: &SlotMap<Connection<Service>>,
+        connections: &PolymorphicVec<'static, SlotMapKey, HeapAllocator>,
+    ) -> Option<(usize, SlotMapKey)> {
+        Self::find_connection_with_condition(
+            connection_storage,
+            connections,
+            |has_data, _has_borrows| !has_data,
+        )
+    }
+
+    fn find_connection_with_condition<C: Fn(bool, bool) -> bool>(
+        connection_storage: &SlotMap<Connection<Service>>,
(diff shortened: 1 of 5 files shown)


iceoryx2/src/port/details/receiver.rs
```diff
@@ -264,22 +264,57 @@ impl<Service: service::Service> Receiver<Service> {
...
-                        .is_err()
-                    {
-                        warn!(from self,
-                            "Expired connection buffer exceeded. A sender disconnected with undelivered data that will be discarded. Increase the expired connection buffer to mitigate the problem.");
-                        connection_storage.remove(key);
+                    let to_be_removed_connections =
+                        unsafe { &mut *to_be_removed_connections.get() };
+                    if to_be_removed_connections.push(key).is_err() {
+                        let msg_begin = "Expired connection buffer exceeded.";
+                        let msg_disconnet_with_data =
```
elfenpiff: Give me a `c`

iceoryx2/src/port/details/receiver.rs
```diff
@@ -264,22 +264,57 @@ impl<Service: service::Service> Receiver<Service> {
...
+                            "A sender disconnected with undelivered data that will be discarded.";
+                        let msg_disconnet_with_borrows = "A sender disconnected with data that is still borrowed. This will lead to segmentation faults when the data is accessed later on.";
```
elfenpiff, line 279: Give me a `c`

iceoryx2/src/port/details/receiver.rs
```diff
@@ -264,22 +264,57 @@ impl<Service: service::Service> Receiver<Service> {
...
+                            );
+
+                        if let Some((index, key)) = index_and_key_to_remove {
+                            // we found a connection without data and borrows which can be removed from the container
+                            to_be_removed_connections.remove(index);
+                            connection_storage.remove(key);
+                        } else if connection_has_borrows {
+                            // we did not find a connection that can safely be removed to create some space for the connection with borrows;
+                            // removing a connection with borrows might lead to segfaults -> try to remove a connection without borrows
+                            let index_and_key_to_remove = Self::find_connection_without_data(
```
elfenpiff: Shouldn't this function be `find_connection_without_borrows`? Otherwise, you remove a connection with borrows and run into the segfault you describe in your comment above.

iceoryx2/src/port/details/receiver.rs
```diff
@@ -264,22 +264,57 @@ impl<Service: service::Service> Receiver<Service> {
...
+                                // we found a connection without borrows which can be removed from the container
+                                warn!(from self, "{} {} {}", msg_begin, msg_disconnet_with_data, msg_end);
+                                to_be_removed_connections.remove(index);
+                                connection_storage.remove(key);
+                            }
+                        }
+
+                        if to_be_removed_connections.push(key).is_err() {
+                            if connection_has_borrows {
+                                fatal_panic!(from self, "{} {} {}", msg_begin, msg_disconnet_with_borrows, msg_end);
```
elfenpiff, line 312: We should aim for just losing data from another connection with data but without borrows when a connection is removed that still has borrowed data.

Here is a thought process of mine:

* Let the number of borrowed samples be `N`
* Let the number of to be removed connections be `M`

When `M` >= `N` this case should never be encountered. The following scenarios would be possible in the worst case `M` == `N`.
* The user has borrowed `M` samples from the to be removed connections and connection cannot be moved to the to be removed connection buffer. Connection is removed and all data is lost.
* The user has borrowed `M - X` (X > 0) samples from the to be removed connections. So a to be removed connection can be removed. The data is lost from that older connection.

So we need to ensure in the service builder or in the (yet non-existing) constructor of this thing this invariant. Or translated into iceoryx2 config terms: 

* Publish-Subscribe:
  * `defaults.publish-subscribe.subscriber-max-borrowed-samples` >= `defaults.publish-subscribe.subscriber-expired-connection-buffer`

* Request-Response:
  * Server
    * `defaults.request-response.server-max-loaned-responses-per-request` >= `defaults.request-response.server-expired-connection-buffer`
  * Client
    * `defaults.request-response.client-expired-connection-buffer` >= `defaults.request-response.max-borrowed-responses-per-pending-response`

With that, a `fatal_panic` or crash at runtime is turned into a construction error when creating the `Subscriber`, `Server` or `Client` port. With this, we can guarantee freedom from interference again, otherwise a wild creation/deletion of a publisher + send sample can cause a panic in an unrelated subscriber.

iceoryx2/src/port/details/receiver.rs
```diff
@@ -288,6 +323,73 @@ impl<Service: service::Service> Receiver<Service> {
...
+            if condition(has_data, has_borrows) {
+                index_and_key = Some((n, connection_key));
+                break;
+            }
+        }
+
+        index_and_key
+    }
+
+    fn all_receiver_channel_has_data_and_has_borrows(
```
elfenpiff: This function name implies that every receiver channel has data and borrows but what you are checking is actually if at least one receiver channel has data and borrows.

Could you rename this function to `at_least_one_receiver_channel_has_data_and_has_borrows`

Mmmmhhhh ... the name would still be a bit misleading, this would imply that the channel has both at the same time. What you mean is that at least one receiver channel has data, and at least one receiver channel (maybe another one) has borrows. Maybe you get inspired for an even better name.

iceoryx2/src/port/details/receiver.rs
elBoberido: Indeed. The comment was right but the compiler does not read comments and compiled my wrong code

iceoryx2/src/port/details/receiver.rs
elBoberido: Can this be done in a follow up or can you give me a hint where to add this check?

The `fatal_panic` would still be there but with `"This should never happen..."`

iceoryx2/src/port/details/receiver.rs
elBoberido: What do you think about `receiver_channels_have_data_or_borrows`?

elfenpiff: I like it

iceoryx2/src/port/details/receiver.rs
elfenpiff: Lets do this in an immediate follow up

elfenpiff: I created the issue #1631

elfenpiff (commented):
I would approve it as soon as there is a test case confirming the bug fix. 

I would add the minimal example you have written in rust here for publish-subscribe:
https://github.com/eclipse-iceoryx/iceoryx2/blob/main/iceoryx2/conformance-tests/src/service_publish_subscribe.rs

And then we need a test cases for request response:
The same as the pubsub one where the client=publisher, server=subscriber
https://github.com/eclipse-iceoryx/iceoryx2/blob/main/iceoryx2/conformance-tests/src/service_request_response.rs

The other direction from server to client is not necessary since as soon as the server removes the active request, the clients corresponding pending response closes the connection since it is interpreted as the server is finished sending responses.

iceoryx2/src/port/details/receiver.rs
```diff
@@ -375,41 +476,57 @@ impl<Service: service::Service> Receiver<Service> {
...
-                        Some(connection) => connection,
-                        None => {
-                            unsafe { clean_connections.push_unchecked((n, *connection_key)) };
+                let mut indices_to_skip = 0;
+                // loop through all 'to_be_removed_connections' break the for-loop if a connection needs to be removed;
+                // then continue the for-loop but skip the previously looped indices
+                loop {
+                    let mut index_and_key = None;
+                    for (n, connection_key) in to_be_removed_connections
+                        .iter()
```
elBoberido, line 487: @elfenpiff if you do not like these changes to get rid of the heap usage, I can drop the commit.

elfenpiff (approved):

iceoryx2/src/port/details/receiver.rs
elfenpiff: I like it

PR #1659, opened 2026-05-22
BASED ON #1648 [#1584] refactor open and create of services

elfenpiff:

## PR Reviewer Reminders

* Public API is documented

## References

Relates to #1584

### DIFF
--- a/iceoryx2-cal/src/static_storage/file.rs
+++ b/iceoryx2-cal/src/static_storage/file.rs
@@ -149,11 +149,26 @@ impl NamedConcept for Locked {
 impl StaticStorageLocked<Storage> for Locked {
     fn unlock(mut self, contents: &[u8]) -> Result<Storage, StaticStorageUnlockError> {
         let msg = "Failed to unlock storage";
-        let bytes_written = fail!(from self, when self.static_storage.file.write(contents),
-            map FileWriteError::InsufficientPermissions => StaticStorageUnlockError::InsufficientPermissions;
-                FileWriteError::NoSpaceLeft => StaticStorageUnlockError::NoSpaceLeft,
-            unmatched StaticStorageUnlockError::InternalError,
-            "{} due to a failure while writing the contents.", msg);
+
+        let bytes_written = match self.static_storage.file.write(contents) {
+            Ok(bytes_written) => bytes_written,
+            Err(FileWriteError::InsufficientPermissions) => {
+                fail!(from self, with StaticStorageUnlockError::InsufficientPermissions,
+                    "{} due to insufficient permissions.", msg);
+            }
+            Err(FileWriteError::Interrupt) => {
+                fail!(from self, with StaticStorageUnlockError::Interrupt,
+                    "{} since an interrupt signal was raised.", msg);
+            }
+            Err(FileWriteError::NoSpaceLeft) => {
+                fail!(from self, with StaticStorageUnlockError::NoSpaceLeft,
+                    "{} since there is not enough space left to write the contents.", msg);
+            }
+            Err(e) => {
+                fail!(from self, with StaticStorageUnlockError::InternalError,
+                    "{} due to an unknown failure while writing the contents. [{e:?}]", msg);
+            }
+        };
 
         if bytes_written != contents.len() as u64 {
             fail!(from self, with StaticStorageUnlockError::NoSpaceLeft,
@@ -161,10 +176,31 @@ impl StaticStorageLocked<Storage> for Locked {
                 msg, contents.len(), bytes_written);
         }
 
-        fail!(from self, when self.static_storage.file.set_permission(FINAL_PERMISSIONS),
-                map FileSetPermissionError::InsufficientPermissions => StaticStorageUnlockError::InsufficientPermissions,
-                unmatched StaticStorageUnlockError::InternalError,
-                "{} due to a failure while updating the permissions to {}.", msg, FINAL_PERMISSIONS);
+        match self.static_storage.file.sync_all() {
+            Ok(()) => (),
+            Err(FileSyncError::Interrupt) => {
+                fail!(from self, with StaticStorageUnlockError::Interrupt,
+                    "{} since an interrupt signal was raised while syncing the underlying file.", msg);
+            }
+            Err(e) => {
+                fail!(from self, with StaticStorageUnlockError::InternalError,
+                    "{} due to an unknown failure while syncing the underlying file. [{e:?}]", msg);
+            }
+        }
+
+        match self.static_storage.file.set_permission(FINAL_PERMISSIONS) {
+            Ok(_) => (),
+            Err(FileSetPermissionError::InsufficientPermissions) => {
+                fail!(from self, with StaticStorageUnlockError::InsufficientPermissions,
+                    "{} since the static storage could not be unlocked due to insufficient permissions.",
+                    msg);
+            }
+            Err(e) => {
+                fail!(from self, with StaticStorageUnlockError::InternalError,
+                    "{} since the static storage could not be unlocked due to an unknown failure. [{e:?}]",
+                    msg);
+            }
+        }
 
         self.static_storage.len = contents.len() as u64;
 
@@ -226,9 +262,17 @@ impl crate::named_concept::NamedConceptMgmt for Storage {
         let mut file = match FileBuilder::new(&file_path).open_existing(AccessMode::Read) {
             Ok(f) => f,
             Err(FileOpenError::FileDoesNotExist) => return Ok(false),
+            Err(FileOpenError::InsufficientPermissions) => {
+                fail!(from origin, with NamedConceptRemoveError::InsufficientPermissions,
+                    "{} since the file could not be opened for permission adjustment.", msg);
+            }
+            Err(FileOpenError::Interrupt) => {
+                fail!(from origin, with NamedConceptRemoveError::Interrupt,
+                    "{} since an interrupt signal was raised.", msg);
+            }
             Err(v) => {
                 fail!(from origin, with NamedConceptRemoveError::InternalError,
-                    "{} since the file could not be opened for permission adjustment ({:?}).", msg, v);
+                    "{} since the file could not be opened due to an unknown failure [{:?}].", msg, v);
             }
         };
 
@@ -237,9 +281,9 @@ impl crate::named_concept::NamedConceptMgmt for Storage {
         match File::remove(&file_path) {
             Ok(v) => Ok(v),
             Err(e) => {
-                if let Err(e) = set_permission_result {
+                if let Err(permission_result) = set_permission_result {
                     warn!(from origin,
-                          "Unable to adjust the files permission as preparation to remove the file ({e:?}).");
+                          "Unable to adjust the files permission as preparation to remove the file ({permission_result:?}).");
                 }
                 match e {
                     FileRemoveError::InsufficientPermissions
@@ -247,7 +291,7 @@ impl crate::named_concept::NamedConceptMgmt for Storage {
                         fail!(from origin, with NamedConceptRemoveError::InsufficientPermissions,
                                 "{} due to insufficient permissions.", msg);
                     }
-                    _ => {
+                    e => {
                         fail!(from origin, with NamedConceptRemoveError::InternalError,
                                 "{} due to unknown failure ({:?}).", msg, e);
                     }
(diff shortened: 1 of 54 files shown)


iceoryx2/src/service/builder/blackboard.rs
```diff
@@ -599,191 +654,141 @@ impl<
         self.create_impl(attributes)
     }
 
+    fn create_blackboard_resources(
```
elfenpiff, line 657: The code was not added or changed, just extracted from the `create` function before the refactoring.

iceoryx2/src/service/builder/blackboard.rs
```diff
@@ -955,132 +961,76 @@ impl<
         self.open_impl(verifier)
     }
 
+    fn open_blackboard_resources(
```
elfenpiff, line 964: The code was not added or changed, just extracted from the `open` function before the refactoring.

iceoryx2/src/service/builder/mod.rs
```diff
@@ -257,73 +330,436 @@ impl<ServiceType: service::Service> BuilderWithServiceType<ServiceType> {
         blackboard::Opener::new(self)
     }
 
+    fn open_or_create<
```
elfenpiff, line 333: This is the key part of the PR, the `open_or_create`, `open`, `create` and `is_service_available` function. Please review this part carefully.

iceoryx2/src/service/mod.rs
```diff
@@ -614,97 +636,148 @@ pub mod internal {
         trace!(from origin, "Send dead node signal on service {}.", service_name);
     }
 
-    fn remove_sender_connection_and_data_segment<S: Service>(
```
elfenpiff, line 617: All `remove_***` function are just moved to `stale_resource_cleanup.rs` where they belong.

iceoryx2/src/service/stale_resource_cleanup.rs
```diff
@@ -11,39 +11,262 @@
...
-pub(crate) enum RemoveStalePortResourcesError {
+pub enum RemoveStalePortResourcesError {
     InsufficientPermissions,
+    Interrupt,
     VersionMismatch,
     InternalError,
 }
 
-pub(crate) unsafe fn remove_data_segment_of_port<Service: service::Service>(
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
elfenpiff, line 62: All `remove_***` function are just moved from `service/mod.rs`.

internal/scripts/ci_build_and_run_end_to_end_tests.sh
```diff
@@ -22,6 +22,8 @@ BUILD_END_TO_END_TESTS=true
 RUN_END_TO_END_TESTS=true
 PYTHON_END_TO_END_TESTS=true
 
+export IOX2_LOG_LEVEL=info
+
```
elBoberido, line 26: Please remove this. When this is in, one cannot change the log level in the terminal anymore but has to edit the script

iceoryx2/src/config.rs
```diff
@@ -235,10 +237,13 @@ impl Default for Global {
     fn default() -> Self {
         Self {
             root_path: Path::new(ICEORYX2_ROOT_PATH).unwrap(),
+            #[cfg(feature = "std")]
             prefix: FileName::new(b"iox2_").unwrap(),
+            #[cfg(not(feature = "std"))]
+            prefix: FileName::new(b"iox2_no_std").unwrap(),
```
elBoberido, line 243: There needs to be an underscore at the end. It is also quite long and we already have issues with long names on some platforms. How about just `iox2nostd_`.

I'm also not sure this is the right approach. While this helps with our setup, user might just use their own prefix without thinking about `std` and `no_std` and then they run into issues we are not able to reproduce.

A better solution would be to have the serialization format encoded in the hash of the file. This makes it independent of any prefix choices.

iceoryx2/src/service/naming_scheme.rs
```diff
@@ -16,7 +16,28 @@ use iceoryx2_bb_container::semantic_string::SemanticString;
...
-use crate::identifiers::UniqueListenerId;
+use crate::{
+    identifiers::{UniqueListenerId, UniqueServiceId},
+    service::service_hash::ServiceHash,
+};
+
+pub(crate) fn static_config_name(service_hash: &ServiceHash) -> FileName {
+    let msg =
+        "The system does not support the required file name length for the dynamic config name.";
+    let origin = "dynamic_config_name()";
```
elBoberido, line 27: Wrong origin
```suggestion
    let origin = "static_config_name()";
```

elBoberido (commented):
First round. The easy part is done. Another 9 files need to be reviewed.

iceoryx2-cal/src/static_storage/file.rs
elBoberido, line 605: With the current setup, this is hammering the file system for the first 10000 iterations. We need to check where the `AdaptiveWait` is used and where it would be better to use a fixed tick rate of 1ms.

iceoryx2/src/node/mod.rs
elBoberido, line 635: Please don't do this catch all and map to `InternalError`. It hides new error and it is impossible to understand what is going on. In this case for example, the new `ServiceRemoveNodeError::Interrupt` is mapped to `NodeCleanupFailure::InternalError`. Not sure it this is intended or not since it is a catch all.

Btw. further down there is a mapping of `RemoveStalePortResourcesError::Interrupt` to `NodeCleanupFailure::Interrupt`.

iceoryx2/src/service/builder/event.rs
```diff
@@ -93,19 +98,63 @@ impl core::error::Error for EventOpenError {}
...
+            EventOpenError::HangsInCreation => ServiceOpenError::HangsInCreation,
+            EventOpenError::IncompatibleMessagingPattern => {
+                ServiceOpenError::IncompatibleMessagingPattern
+            }
+            EventOpenError::InsufficientPermissions => ServiceOpenError::InsufficientPermissions,
+            EventOpenError::IsMarkedForDestruction => ServiceOpenError::IsMarkedForDestruction,
+            EventOpenError::ServiceInCorruptedState => ServiceOpenError::ServiceInCorruptedState,
+            EventOpenError::UnableToCreateServiceTag => ServiceOpenError::UnableToCreateServiceTag,
+            EventOpenError::VersionMismatch => ServiceOpenError::VersionMismatch,
+            _ => ServiceOpenError::InternalFailure,
```
elBoberido, line 148: This is not future proof. Any new `EventOpenError` will end up as `InternalFailure`

iceoryx2/src/config.rs
elfenpiff: I agree, but thte service hash would be the wrong approach. The names require an internal prefix which is the hash or a number of the service variant. This would be something bigger, I will create an issue.

iceoryx2/src/service/builder/event.rs
```diff
@@ -114,32 +163,71 @@ pub enum EventCreateError {
...
+            EventCreateError::ServiceInCorruptedState => {
+                ServiceCreateError::ServiceInCorruptedState
+            }
+            EventCreateError::UnableToCreateServiceTag => {
+                ServiceCreateError::UnableToCreateServiceTag
+            }
+            EventCreateError::ServiceConfigCouldNotBeCreated => {
+                ServiceCreateError::ServiceConfigCouldNotBeCreated
+            }
+            _ => ServiceCreateError::InternalFailure,
```
elBoberido, line 218: Same here

iceoryx2/src/config.rs
elfenpiff: Here is the issue: https://github.com/eclipse-iceoryx/iceoryx2/issues/1678

iceoryx2/src/service/builder/publish_subscribe.rs
```diff
@@ -142,30 +208,70 @@ impl core::fmt::Display for PublishSubscribeCreateError {
...
+            PublishSubscribeCreateError::ServiceConfigCouldNotBeCreated => {
+                ServiceCreateError::ServiceConfigCouldNotBeCreated
+            }
+            PublishSubscribeCreateError::ServiceInCorruptedState => {
+                ServiceCreateError::ServiceInCorruptedState
+            }
+            PublishSubscribeCreateError::UnableToCreateServiceTag => {
+                ServiceCreateError::UnableToCreateServiceTag
+            }
+            _ => ServiceCreateError::InternalFailure,
```
elBoberido, line 255: And here

iceoryx2/src/service/builder/request_response.rs
```diff
@@ -146,23 +215,68 @@ impl core::fmt::Display for RequestResponseCreateError {
...
+            RequestResponseCreateError::ServiceConfigCouldNotBeCreated => {
+                ServiceCreateError::ServiceConfigCouldNotBeCreated
+            }
+            RequestResponseCreateError::ServiceInCorruptedState => {
+                ServiceCreateError::ServiceInCorruptedState
+            }
+            RequestResponseCreateError::UnableToCreateServiceTag => {
+                ServiceCreateError::UnableToCreateServiceTag
+            }
+            _ => ServiceCreateError::InternalFailure,
```
elBoberido, line 279: And here

iceoryx2/src/service/builder/mod.rs
```diff
@@ -118,6 +127,70 @@ pub struct Builder<S: Service> {
...
+    ServiceConfigCouldNotBeCreated,
+    Interrupt,
+}
+
+impl From<ServiceState> for ServiceCreateError {
+    fn from(value: ServiceState) -> Self {
+        match value {
+            ServiceState::IncompatibleMessagingPattern
+            | ServiceState::HangsInCreation
+            | ServiceState::IncompatiblePayload => ServiceCreateError::AlreadyExists,
```
elBoberido, line 147: I guess the `HangsInCreation` can only happen if there is an open/close ping pong and therefore, an `AlreadyExists` should be fine for `ServiceCreateError`

iceoryx2/src/service/builder/mod.rs
```diff
@@ -257,73 +330,436 @@ impl<ServiceType: service::Service> BuilderWithServiceType<ServiceType> {
...
+              "{msg} since the adaptive wait could not be created.");
+        let start = fail!(from self,
+              when Time::now(),
+              with internal_error_type,
+              "{msg} since the current time could not be acquired.");
+        let creation_timeout = self.shared_node.config().global.creation_timeout;
+        let attribute_specifier = AttributeSpecifier(attributes.required_attributes().clone());
+
+        let mut flux_counter = 0;
+        let mut last_errors = vec![];
```
elBoberido, line 362: Please do not allocate for this. Just use a `StaticVec`

iceoryx2/src/service/builder/mod.rs
```diff
@@ -257,73 +330,436 @@ impl<ServiceType: service::Service> BuilderWithServiceType<ServiceType> {
...
+                            | ServiceCreateError::ServiceInCorruptedState
+                            | ServiceCreateError::UnableToCreateServiceTag
+                            | ServiceCreateError::Interrupt
+                            | ServiceCreateError::ServiceConfigCouldNotBeCreated => {
+                                return Err(Into::<ErrorTypeOpenOrCreate>::into(e));
+                            }
+                        }
+                    }
+                }
+            }
```
elBoberido, line 410: Please add a few lines of comments in order to understand the idea behind this logic in a few months.

It is sufficient if it is at the beginning of the loop.

I know that comments can go out of sync but at least there is a chance to understand if the logic works as intended.

iceoryx2/src/service/builder/mod.rs
```diff
@@ -257,73 +330,436 @@ impl<ServiceType: service::Service> BuilderWithServiceType<ServiceType> {
...
+                        | ServiceOpenError::Interrupt
+                        | ServiceOpenError::VersionMismatch => {
+                            return Err(Into::<ErrorTypeOpenOrCreate>::into(e));
+                        }
+                    }
+                }
+            }
+
+            if try_create_service {
+                flux_counter += 1;
```
elBoberido, line 391: For example. Here it would help to understand that the `flux_counter` can will only be incremented when the `open_call` fails with a `DoesNotExist`. Try to keep the cognitive load low when this code is looked at. It is an additional guard to check whether the code above does what is intended here.

iceoryx2/src/service/builder/mod.rs
```diff
@@ -257,73 +330,436 @@ impl<ServiceType: service::Service> BuilderWithServiceType<ServiceType> {
...
+                        "{} since the elapsed time could not be acquired.", msg);
+                    if elapsed > creation_timeout {
+                        fail!(from origin, with ServiceOpenError::HangsInCreation.into(),
+                            "{} since the service hangs in creation", msg);
+                    } else {
+                        if let Err(e) = adaptive_wait.wait() {
+                            fail!(from origin,
+                            with ServiceOpenError::InternalFailure.into(),
+                            "{} since the adaptive wait failed. [{e:?}]", msg);
+                        }
```
elBoberido, line 477: The `wait` call can also return a `Interrupt` error via the `NanosleepError`

iceoryx2/src/config.rs
elBoberido: Agree, but please add the `_`, i.e. `iox2_no_std_`

elBoberido (approved):
Since there are already other PRs open for this issue, the findings can be applied in those since they are not critical.

iceoryx2/src/service/builder/mod.rs
elfenpiff: Create does not come with a timeout and the HangsInCreation is emitted by is_service_available after it waited for the timeout. But in the create case it simply means there is already something.

PR #1723, opened 2026-06-09
[#1722] Remove allocations in tunnel hot path

### LINKED ISSUE #1722: Remove superfluous allocations in tunnel discovery logic
## Brief Description

The solution to track and synchronize discovered services across both the local host and remote hosts is sub-optimal, utilizing `Vec`s to collect discovery changes before reconciling them. This first solution prioritized functionality, but now that the tests are in place, a second pass for optimization should be made.

orecham:
## Notes for Reviewer

* Move discovery logic into `DiscoveryState` component
* Move port and relay logic into `Bridge` component
* Remove all allocations in tunnel's discovery hot path
    * No more temporary `Vec`s to gather changes and diff - swapped for operations over `DiscoveryState` and `Bridge`
* Update README to describe both polling and reactive tunnel execution

The `tunnel.rs` file changed significantly. Some parts moved to `DiscoveryState` and some to `Bridge`, and the discovery algorithm completely changed. It would be easiest to review the latest state of the file rather than the diff.

## PR Reviewer Reminders

* Public API is documented

## References

Closes #1722

### DIFF
--- a/iceoryx2-services/tunnel/src/bridge.rs
+++ b/iceoryx2-services/tunnel/src/bridge.rs
@@ -1,0 +1,193 @@
+// Copyright (c) 2026 Contributors to the Eclipse Foundation
+//
+// See the NOTICE file(s) distributed with this work for additional
+// information regarding copyright ownership.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Apache Software License 2.0 which is available at
+// https://www.apache.org/licenses/LICENSE-2.0, or the MIT license
+// which is available at https://opensource.org/licenses/MIT.
+//
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use alloc::format;
+
+use iceoryx2::identifiers::UniqueNodeId;
+use iceoryx2::node::Node;
+use iceoryx2::service::Service;
+use iceoryx2::service::static_config::StaticConfig;
+use iceoryx2::service::static_config::messaging_pattern::MessagingPattern;
+use iceoryx2_log::{fail, info};
+use iceoryx2_services_tunnel_backend::traits::{
+    Backend, EventRelay, PublishSubscribeRelay, RelayBuilder, RelayFactory,
+};
+use iceoryx2_services_tunnel_backend::types::publish_subscribe::LoanFn;
+
+use crate::ports::event::EventPorts;
+use crate::ports::publish_subscribe::PublishSubscribePorts;
+use crate::tunnel::{DiscoveryError, PropagateError};
+
+/// A bidirectional bridge for a single service: the local iceoryx2 ports on one
+/// side and the backend relay on the other.
+#[derive(Debug)]
+pub(crate) enum Bridge<S: Service, B: Backend<S>> {
+    PublishSubscribe {
+        ports: PublishSubscribePorts<S>,
+        relay: B::PublishSubscribeRelay,
+    },
+    Event {
+        ports: EventPorts<S>,
+        relay: B::EventRelay,
+    },
+}
+
+impl<S: Service, B: Backend<S>> Bridge<S, B> {
+    /// Creates the ports and relay matching the messaging pattern of
+    /// `static_config`.
+    pub(crate) fn open(
+        node: &Node<S>,
+        backend: &B,
+        static_config: &StaticConfig,
+    ) -> Result<Self, DiscoveryError> {
+        let origin = "Bridge::open";
+
+        match static_config.messaging_pattern() {
+            MessagingPattern::PublishSubscribe(_) => {
+                let ports = fail!(
+                    from origin,
+                    when PublishSubscribePorts::new(static_config, node),
+                    with DiscoveryError::PublishSubscribePortCreation,
+                    "Failed to create publish-subscribe ports"
+                );
+                let relay = fail!(
+                    from origin,
+                    when backend.relay_builder().publish_subscribe(static_config).create(),
+                    with DiscoveryError::PublishSubscribeRelayCreation,
+                    "Failed to create publish-subscribe relay"
+                );
+                Ok(Bridge::PublishSubscribe { ports, relay })
+            }
+            MessagingPattern::Event(_) => {
+                let ports = fail!(
+                    from origin,
+                    when EventPorts::new(static_config, node),
+                    with DiscoveryError::EventPortsCreation,
+                    "Failed to create event ports"
+                );
+                let relay = fail!(
+                    from origin,
+                    when backend.relay_builder().event(static_config).create(),
+                    with DiscoveryError::EventRelayCreation,
+                    "Failed to create event relay"
+                );
+                Ok(Bridge::Event { ports, relay })
+            }
+            pattern => {
+                fail!(
+                    from origin,
+                    with DiscoveryError::UnsupportedMessagingPattern,
+                    "Cannot open bridge for unsupported messaging pattern: {}", pattern
+                );
+            }
+        }
+    }
+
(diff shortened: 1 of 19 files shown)


iceoryx2-services/tunnel/src/discovery/state.rs
```diff
@@ -0,0 +1,210 @@
...
+            }
+        }
+
+        let mut result = Ok(());
+        self.offered.retain(|_, offered_service| {
+            if offered_service.last_seen == epoch {
+                true
+            } else {
+                if result.is_ok() {
+                    result = on_removed(&offered_service.static_config);
```
FerdinandSpitzschnueffler, line 100: Why do we stop calling `on_removed` once it returned an error for an offered service?

iceoryx2-services/tunnel/src/bridge.rs
```diff
@@ -0,0 +1,193 @@
...
+        }),
+        with PropagateError::PayloadPropagation,
+        "Failed to receive publish-subscribe payload for propagation"
+    );
+    if propagated {
+        info!(
+            from origin,
+            "Propagated {}({})",
+            port.static_config.messaging_pattern(),
+            port.static_config.name()
```
FerdinandSpitzschnueffler, line 126: Could be stored in separate variables and reused here and in l. 143/144.

iceoryx2-services/tunnel/src/discovery/tracker.rs
```diff
@@ -36,68 +31,30 @@ impl core::error::Error for DiscoveryError {}
...
-            );
-        }
-
-        for event in events {
-            process_discovery(event);
-        }
-
+    /// Synchronize the tracker with the services currently in the system.
+    pub fn sync(&mut self) -> Result<(), DiscoveryError> {
+        let origin = "DiscoveryTracker::refresh";
```
FerdinandSpitzschnueffler: The function is still named `sync`. Did you want to rename it?

iceoryx2-services/tunnel/src/tunnel.rs
```diff
@@ -226,521 +122,309 @@ impl<S: Service, B: for<'a> Backend<S> + Debug> Tunnel<S, B> {
...
-    /// Tracker-mode local discovery: sync against the local registry, then
-    /// derive `locally_offered` transitions from the snapshot. A service is
-    /// considered locally offered when at least one non-tunnel, non-dead node
-    /// is offering it.
-    fn discover_over_tracker(&mut self) -> Result<(), DiscoveryError> {
+    /// Tracker-mode local discovery: refresh the local registry snapshot, then
+    /// bring the locally-offered set in line with it. A service is considered
+    /// locally offered when at least one non-tunnel, non-dead node offers it.
+    fn tracker_discovery(&mut self) -> Result<(), DiscoveryError> {
         let origin = format!("Tunnel({})::discover_over_tracker", self.node.id());
```
FerdinandSpitzschnueffler:
```suggestion
        let origin = format!("Tunnel({})::tracker_discovery", self.node.id());
```

iceoryx2-services/tunnel/src/tunnel.rs
```diff
@@ -226,521 +122,309 @@ impl<S: Service, B: for<'a> Backend<S> + Debug> Tunnel<S, B> {
...
-            match self.relays.event.get(service_hash) {
-                Some(relay) => {
-                    propagate_events::<S, B>(self.node.id(), port, relay)?;
-                }
-                None => {
-                    warn!(from origin, "No relay available for {:?}", service_hash);
-                    continue;
-                }
-            };
+    /// Updates the locally offerred services..
```
FerdinandSpitzschnueffler:
```suggestion
    /// Updates the locally offerred services.
```

iceoryx2-services/tunnel/src/tunnel.rs
```diff
@@ -226,521 +122,309 @@ impl<S: Service, B: for<'a> Backend<S> + Debug> Tunnel<S, B> {
...
+    pub fn propagate(&mut self) -> Result<(), PropagateError> {
+        debug_assert!(
+            self.bridges.len() == self.discovery_state.snapshot().iter().count()
+                && self
+                    .discovery_state
+                    .snapshot()
+                    .iter()
+                    .all(|(hash, _)| self.bridges.contains_key(hash)),
+            "bridges out of sync with discovery state"
         );
```
FerdinandSpitzschnueffler: What do you think about moving this into a separate `verify_sync` function? The same check is done in `tunneled_services()` (and maybe in other functions as well).

orecham: @FerdinandSpitzschnueffler I had this initially but changed to this to keep the fact that it is a `debug_assert` visible. I only check this in two locations currently - `propagation`, and `tunneled_services`.

I can move it though if you disagree with my reasoning.

iceoryx2-services/tunnel/src/discovery/tracker.rs
orecham: Renaming oversight :sweat:

iceoryx2-services/tunnel/src/bridge.rs
orecham: @FerdinandSpitzschnueffler Does this make a difference? Or just a style preference?

iceoryx2-services/tunnel/src/discovery/state.rs
orecham: @FerdinandSpitzschnueffler  I am not sure if this is a recoverable error. Thoughts?

FerdinandSpitzschnueffler: `on_removed` calls at some point `announce_added` from the backend implementation, right? I was wondering if there could be an implementation where it makes sense that this function fails for one service hash but could still succeed for others... 🤔 But that probably doesn't make sense. I actually asked to understand, not to propose a change.

iceoryx2-services/tunnel/src/bridge.rs
FerdinandSpitzschnueffler: It's just a style preference - up to you.

iceoryx2-services/tunnel/src/tunnel.rs
FerdinandSpitzschnueffler: I'd prefer to have it in a separate function. I think this makes it clearer what happens inside `propagate` and `tunneled_services`. Additionally, the check is relatively complex and if it is in a separate function, one only need to understand it once. The latter might be more of a review issue, though.

iceoryx2-services/tunnel/src/discovery/state.rs
orecham: @FerdinandSpitzschnueffler Yes these callbacks map to `announce_added` and `announce_removed`.

The best approach is not obvious to me either, I just went with this because I thought if one announcement (addition or removal) were to fail the "world view" would be wrong across hosts. Failing hard prevents surprising (maybe annoying) behaviour like "The tunnel is running but I am not receiving data"

Recovering does not seem trivial ... 
* Keep retrying and stall the application?
* Let it stay out sync but retry next iteration?
* Just log a warning and continue running?
* Something else?

It does make the tunnel a little fragile though depending on the Backend used, could spend some more time on it in a follow-up.

FerdinandSpitzschnueffler: I think we should stay with your current approach of failing at this point for now. If it turns out to be insufficient, we can still consider how to give for instance the Backend more control over the recovery...

FerdinandSpitzschnueffler (approved):

PR #1744, opened 2026-06-15
[#1742] Implement ros2 tunnel send path

### LINKED ISSUE #1742: Add ROS 2 Gateway
## Brief Description

Add a `Backend` implementation that propagates payloads between `iceoryx2` services and ROS 2 topics. First version to target both Humble/`rmw_fastdds` and Jazzy/`rmw_fastdds`, with pathways to extend support to other distros/rmws.

* [x] Discovery
    * [x] Static - specify ROS 2 topics to tunnel on initialization
    * [x] Dynamic - discover topics over ROS 2 graph
* [x] Matching
    * [x] Map between topic names and service names
    * [x] Ergonomic API on `iceoryx2` services to provide automatic mapping of ros 2 topic names to iceoryx2 service names
    * [x] Resolve QoS
* [x] Tunnel mode - propagate CDR-serialized payloads as-is
    * `iceoryx2` applications to use `cdr` crate and `rosidl_rust` message definitions to serialize/deserialize payloads into and out of `iceoryx2` payloads
* [x] Gateway mode - translate the CDR bytes to a "zero-copy deserialization" wire format, directly on write into the `iceoryx2` payload buffer
    * The zero-copy deserialization format enables usage of all message types, fixed, bounded, unbounded, however iceoryx apps must then work with the payloads via the chosen deserialization library
    * `iceoryx2` to provide an ergonomic zero-copy API for application to work with the serialized structs in the `iceoryx2` payloads
* [ ] Add documentation to the `iceoryx2-book`

orecham:
## Notes for Reviewer

Implements the send path for the ROS 2 tunnel backend.

Remaining capabilities and CI to be merged in subsequent pull requests. See the README in `integrations/ros2/tunnel-backend` for the development status.

## PR Reviewer Reminders

* Public API is documented

## References

Closes #1742


integrations/ros2/examples/demo_nodes_iceoryx2/src/bin/listener.rs
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) 2026 Contributors to the Eclipse Foundation
```
elfenpiff, line 1: Shouldn't the file be named `subscriber.rs` since it has a subscriber and receives data? A listener port is something different in iceoryx2. Another alternative would be `receiver.rs`.

What terminology do you follow here? ROS2 or iceoryx2?

integrations/ros2/examples/demo_nodes_iceoryx2/src/bin/talker.rs
```diff
@@ -0,0 +1,75 @@
+// Copyright (c) 2026 Contributors to the Eclipse Foundation
```
elfenpiff, line 1: Why not name the file `publisher.rs` or `sender.rs`?

integrations/ros2/examples/demo_nodes_iceoryx2/README.md
```diff
@@ -0,0 +1,51 @@
+# demo_nodes_iceoryx2
```
elfenpiff, line 1: I assume that it is not possible to move the ros2 examples folder into our main root example folder but would it be possible to add at least a link there? So I think about a links in the `$ROOT$/examples/README.md` file and an additional `ros2.md` file that links to this location?

integrations/ros2/examples/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+```bash
+cd <workspace>
+source /opt/ros/<distro>/setup.bash && source install/setup.bash
+colcon build --packages-select <example>
+```
+
+## Troubleshooting
+
+```console
```
elfenpiff, line 59: Can you please also add this entry to the `FAQ.md`? It is the first location I look when I have a problem. You can leave this entry here as well.

integrations/ros2/tunnel-backend/src/rcl/node.rs
```diff
@@ -0,0 +1,107 @@
...
+    rcl_get_zero_initialized_init_options, rcl_get_zero_initialized_node, rcl_init,
+    rcl_init_options_fini, rcl_init_options_init, rcl_node_fini, rcl_node_get_default_options,
+    rcl_node_init, rcl_node_t, rcl_shutdown, rcutils_get_default_allocator,
+};
+
+/// rcl is initialized without forwarding any command-line arguments.
+const NO_ARGS: core::ffi::c_int = 0;
+
+#[derive(Debug, Eq, PartialEq, Clone, Copy)]
+pub enum CreationError {
```
elfenpiff, line 32: For completeness, would you mind adding documentation here as well? And what does the `i32` contain?

integrations/ros2/tunnel-backend/src/rcl/node.rs
```diff
@@ -0,0 +1,107 @@
...
+/// the context.
+#[derive(Debug)]
+pub struct Node {
+    node: *mut rcl_node_t,
+    context: *mut rcl_context_t,
+}
+
+impl Node {
+    pub fn create(name: &str, namespace: &str) -> Result<Self, CreationError> {
+        let name = CString::new(name).map_err(|_| CreationError::InvalidName)?;
```
elfenpiff: When can this happen? I think only when the user adds a `\0` to the name?

Could you please use the `fail!` macro everywhere and provide a meaningful log-messsage for every error case with origin `ros2::tunnel_backend::rcl::Node::create()` as origin.

integrations/ros2/tunnel-backend/src/rcl/node.rs
```diff
@@ -0,0 +1,107 @@
...
+        let namespace = CString::new(namespace).map_err(|_| CreationError::InvalidName)?;
+
+        unsafe {
+            let mut init_options = rcl_get_zero_initialized_init_options();
+            let ret = rcl_init_options_init(&mut init_options, rcutils_get_default_allocator());
+            if ret != RCL_RET_OK as i32 {
```
elfenpiff: When something goes wrong, it would be great if we had more detailed error handling here. Especially, we must explicitly handle the signal interrupt error everywhere and forward this cleanly to the user.

Instead of forwarding the `i32`, could you try to handle this here more explicitly with detailed error messages and a more diverse error enum, instead of just forwarding the number, where the user has no clue where the number even originates or what it means.

This applies to every `rcl` call.

integrations/ros2/tunnel-backend/src/rcl/publisher.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+impl core::error::Error for PublishError {}
+
+/// Publishes pre-serialized messages on a ROS 2 topic.
+#[derive(Debug)]
+pub struct Publisher {
+    publisher: *mut r2r_rcl::rcl_publisher_t,
+    node: Rc<Node>,
+    /// Keeps the typesupport library loaded while the endpoint uses it.
```
elfenpiff: I do not understand this comment? The typesupport library is kept being loaded? Otherwise a dynamic linked library is removed from the binary?

integrations/ros2/tunnel-backend/src/rcl/publisher.rs
```diff
@@ -0,0 +1,119 @@
...
+            let options = rcl_publisher_get_default_options();
+            let ret = rcl_publisher_init(
+                publisher.as_mut(),
+                node.handle(),
+                type_support.handle(),
+                topic.as_ptr(),
+                &options,
+            );
+            if ret != RCL_RET_OK as i32 {
+                return Err(CreationError::PublisherInit(ret));
```
elfenpiff: Please refine the error handling here as well. Please do not return error enums directly, always use the `fail!` macro and try to perform also an error handling on the `i32` level for all `rcl` calls so that interrupt signals are forwarded and also the other errors are handled more explicitly. (As far as it makes sense.) See this as an example: https://github.com/eclipse-iceoryx/iceoryx2/blob/main/iceoryx2-bb/posix/src/process_state.rs#L1028

Please apply this to all `rcl` calls in this file.

integrations/ros2/tunnel-backend/src/rcl/node.rs
```diff
@@ -0,0 +1,107 @@
...
+/// The tunnel is a single node, so it can be coupled to
+/// the context.
+#[derive(Debug)]
+pub struct Node {
+    node: *mut rcl_node_t,
+    context: *mut rcl_context_t,
+}
+
+impl Node {
+    pub fn create(name: &str, namespace: &str) -> Result<Self, CreationError> {
```
elfenpiff: If you have more than one argument in `new` we use the builder pattern throughout the code base to make this more readable and less error prone. Especially, when both arguments have the same type.

Could you therefore please create a `NodeBuilder` where in `new` the name is set and there is a `namespace` explicit setter and `create` creates the node.
As far as I understood you, the namespace is an optional argument.

Since the `name` and the `namespace` name can have invalid states and we would be required to test their validity everytime we get a `str`, it would be best to create a strong type for them. So could you please introduce a `Ros2NodeName` and `Ros2Namespace` strong types that check the validity when they are created.

They should maybe be based on the semantic string, and implement ZeroCopySend so that we can send them around.

integrations/ros2/tunnel-backend/src/rcl/node.rs
```diff
@@ -0,0 +1,107 @@
...
+            let ret = rcl_init_options_init(&mut init_options, rcutils_get_default_allocator());
+            if ret != RCL_RET_OK as i32 {
+                return Err(CreationError::InitOptionsInit(ret));
+            }
+
+            let mut context = Box::new(rcl_get_zero_initialized_context());
+            let ret = rcl_init(NO_ARGS, core::ptr::null(), &init_options, context.as_mut());
+            let _ = rcl_init_options_fini(&mut init_options);
+            if ret != RCL_RET_OK as i32 {
+                return Err(CreationError::ContextInit(ret));
```
elfenpiff: Don't we need to add here `rcl_shutdown` and `rcl_context_fini` as well?

elfenpiff: Does it make sense to use our scope guard here: https://github.com/eclipse-iceoryx/iceoryx2/blob/main/iceoryx2-bb/elementary/src/scope_guard.rs

That would take care of those accidental resource leaks caused by errors.

integrations/ros2/tunnel-backend/src/rcl/publisher.rs
```diff
@@ -0,0 +1,119 @@
...
+#[derive(Debug)]
+pub struct Publisher {
+    publisher: *mut r2r_rcl::rcl_publisher_t,
+    node: Rc<Node>,
+    /// Keeps the typesupport library loaded while the endpoint uses it.
+    _type_support: TypeSupport,
+}
+
+impl Publisher {
+    pub fn create(
```
elfenpiff: Since you need a node to create the publisher, can we upgrade the `Ros2Node` from this PR to a publisher factory?

integrations/ros2/tunnel-backend/src/rcl/publisher.rs
```diff
@@ -0,0 +1,119 @@
...
+        node: &Rc<Node>,
+        topic: &str,
+        type_support: TypeSupport,
```
elfenpiff: I am wondering what this thing is. Could we add some documentation here and a minimal example in the docs?

integrations/ros2/tunnel-backend/src/rcl/publisher.rs
elfenpiff: It may be even worth to introduce a `PublisherBuilder`. The topic name is required as argument when `Node::create_publisher()` is called, and all other config options are then collected by the builder that then creates the publisher.

integrations/ros2/tunnel-backend/src/rcl/publisher.rs
```diff
@@ -0,0 +1,119 @@
...
+        write!(f, "PublishError::{self:?}")
+    }
+}
+
+impl core::error::Error for PublishError {}
+
+/// Publishes pre-serialized messages on a ROS 2 topic.
+#[derive(Debug)]
+pub struct Publisher {
+    publisher: *mut r2r_rcl::rcl_publisher_t,
```
elfenpiff: Can we put this in a `Box`? I think I know why you are using a raw pointer here instead of a box, since you would have a problem in `Publisher::send` then. But here I would prefer maybe a `RefCell<Box<>>`

integrations/ros2/tunnel-backend/src/rcl/node.rs
```diff
@@ -0,0 +1,107 @@
...
+}
+
+impl core::error::Error for CreationError {}
+
+/// An rcl node together with the context it belongs to.
+/// The tunnel is a single node, so it can be coupled to
+/// the context.
+#[derive(Debug)]
+pub struct Node {
+    node: *mut rcl_node_t,
```
elfenpiff: Please use here a `Box` or a `RefCell<Box<>>` instead of a raw pointer. This applies to all raw pointers.

integrations/ros2/tunnel-backend/src/relays/event.rs
```diff
@@ -0,0 +1,108 @@
...
+/// Builder for event [`Relay`]s.
+#[derive(Debug)]
+pub struct Builder<'config, S: Service> {
+    static_config: &'config StaticConfig,
+    wake: Option<Arc<WakeHandle<local_threadsafe::Service>>>,
+    _phantom: core::marker::PhantomData<S>,
+}
+
+impl<'config, S: Service> Builder<'config, S> {
+    pub fn new(
```
elfenpiff, line 83: If you have a builder, please do not use more than one argument per function call. Usually you have `new` and `create` where you could add the mandatory arguments. Especially the `wake` thing seems optional, so you could set it to `None` by default.

integrations/ros2/tunnel-backend/src/relays/event.rs
```diff
@@ -0,0 +1,108 @@
...
+    }
+}
+
+impl<S: Service> RelayBuilder for Builder<'_, S> {
+    type CreationError = CreationError;
+    type Relay = Relay<S>;
+
+    fn create(self) -> Result<Self::Relay, Self::CreationError> {
+        warn!(
+            "ROS 2 has no equivalent to iceoryx2 events; events of '{}' will not be tunneled",
```
elfenpiff, line 101: This is correct but it still makes sense to forward them as pubsub messages with an `u64` payload. But when a user uses pubsub and event with the same name, this might clash - maybe it should be configurable and the user can decide. But I would not disable it by default - fine for now.

integrations/ros2/tunnel-backend/src/relays/factory.rs
```diff
@@ -0,0 +1,84 @@
...
+    node: Rc<rcl::Node>,
+    type_support: Rc<TypeSupportRegistry>,
+    /// Wake handle to be signaled by relays when new data arrives.
+    /// `None` when the backend was constructed in polled mode.
+    wake: Option<Arc<WakeHandle<local_threadsafe::Service>>>,
+    _phantom: core::marker::PhantomData<S>,
+}
+
+impl<S: Service> Factory<S> {
+    pub fn new(
```
elfenpiff, line 34: Can we introduce here a FactoryBuilder abstraction as well and make the `Ros2Node` from this PR a Factory factory as well? Maybe we can rename this to RelayFactory as well.

integrations/ros2/tunnel-backend/src/relays/publish_subscribe.rs
```diff
@@ -0,0 +1,146 @@
...
+    _phantom: core::marker::PhantomData<S>,
+}
+
+impl<S: Service> PublishSubscribeRelay<S> for Relay<S> {
+    type SendError = SendError;
+    type ReceiveError = ReceiveError;
+
+    fn send(&self, sample: Sample<S>) -> Result<(), Self::SendError> {
+        self.publisher
+            .publish(payload::as_bytes(sample.payload()))
```
elfenpiff: Please do not return errors without printing a debug log message. Can you please use the `fail!` macro here and everywhere, where just the error is returned.

integrations/ros2/tunnel-backend/src/relays/publish_subscribe.rs
```diff
@@ -0,0 +1,146 @@
...
+pub struct Builder<'config, S: Service> {
+    node: Rc<rcl::Node>,
+    type_support: Rc<TypeSupportRegistry>,
+    static_config: &'config StaticConfig,
+    wake: Option<Arc<WakeHandle<local_threadsafe::Service>>>,
+    _phantom: core::marker::PhantomData<S>,
+}
+
+impl<'config, S: Service> Builder<'config, S> {
+    pub fn new(
```
elfenpiff, line 107: Please use at most one argument per builder function call and spread these 4 arguments over 4 builder functions.

integrations/ros2/tunnel-backend/src/payload.rs
```diff
@@ -0,0 +1,31 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Reinterpretation between iceoryx2's untyped payload marker and raw
+//! bytes.
+
+use iceoryx2::service::builder::CustomPayloadMarker;
+
+// The byte views rely on the marker being exactly one byte.
+const _: () = assert!(core::mem::size_of::<CustomPayloadMarker>() == 1);
```
elfenpiff, line 19: This is actually important for other parts of iceoryx2 as well. Could you add this to the definition of the custom payload marker so that this guarantee is available for everything in iceoryx2.

integrations/ros2/tunnel-backend/src/typesupport.rs
```diff
@@ -0,0 +1,159 @@
...
+
+impl TypeSupportRegistry {
+    pub fn load(&self, type_name: &str) -> Result<TypeSupport, LoadError> {
+        if let Some(entry) = self.entries.borrow().get(type_name) {
+            return Ok(TypeSupport {
+                entry: entry.clone(),
+            });
+        }
+
+        let entry = Rc::new(load_typesupport_library(type_name)?);
```
elfenpiff: Please use the fail macro everywhere instead of returning the error directly.

integrations/ros2/tunnel-backend/src/typesupport.rs
```diff
@@ -0,0 +1,159 @@
...
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(f, "LoadError::{self:?}")
+    }
+}
+
+impl core::error::Error for LoadError {}
+
+#[derive(Debug)]
+struct Entry {
+    handle: *const rosidl_message_type_support_t,
```
elfenpiff, line 45: Can we use `Box` here?

integrations/ros2/examples/demo_nodes_iceoryx2/src/bin/listener.rs
orecham: the names mirror the ROS 2 demos:

* https://github.com/ros2/demos/blob/rolling/demo_nodes_cpp/src/topics/talker.cpp
* https://github.com/ros2/demos/blob/rolling/demo_nodes_cpp/src/topics/listener.cpp

integrations/ros2/examples/demo_nodes_iceoryx2/README.md
orecham: Could this be postponed until the implementation is complete and the organization of things is stabilized? Since this is the decoupled `integrations` workspace I am not so sure about the idea though.

integrations/ros2/tunnel-backend/src/rcl/publisher.rs
orecham: The `Drop` implementation for the underlying `Library` struct closes the loaded typesupport library.

integrations/ros2/tunnel-backend/src/relays/event.rs
orecham: This is not a user API, it is called by the factory. The second argument is optional for reactive tunnels. It is set by defining the `ReactiveTunnel` trait.

Changing in this PR would be scope creep, could you make an issue if you think it is important? I do not immediately follow the necessity but could be convinced.

integrations/ros2/tunnel-backend/src/relays/event.rs
orecham: Can be added in a follow-up. If you decide it is important, can you add it to: https://github.com/eclipse-iceoryx/iceoryx2/issues/1742

integrations/ros2/tunnel-backend/src/relays/factory.rs
orecham: I do not follow. Do you mean the wrapper around the FFI functions? This trait is machinery for the tunnel specifically.

integrations/ros2/tunnel-backend/src/relays/publish_subscribe.rs
orecham: Same as https://github.com/eclipse-iceoryx/iceoryx2/pull/1744#discussion_r3423305833

integrations/ros2/tunnel-backend/src/payload.rs
orecham: We can do this in a follow-up - could you create an issue? Otherwise it could scope creep this PR.

integrations/ros2/tunnel-backend/src/rcl/publisher.rs
orecham: Agree the invariants could be better represented here, will think what fits best.

integrations/ros2/tunnel-backend/src/rcl/node.rs
orecham: There's nothing to clean up when `rcl_init` fails. The `context` is not created.

integrations/ros2/tunnel-backend/src/typesupport.rs
orecham: This is a pointer into the typesupport library managed by the `Library`. The `Library` is kept open by all consumers of the pointer via the `_type_support` members (e.g. in `Publisher`). 
Is something more correct than a raw pointer for this case?

integrations/ros2/tunnel-backend/src/rcl/node.rs
orecham: Codes have been wrapped in an enum.

integrations/ros2/examples/README.md
orecham: This is specific to the backend, not iceoryx in general. Not sure it fits in the root FAQ. There will eventually be an article on this in the book which I think is more discoverable anyway.

integrations/ros2/tunnel-backend/src/rcl/publisher.rs
orecham: Added a more detailed explanation to the struct

integrations/ros2/tunnel-backend/src/rcl/node.rs
orecham: I used `Box<UnsafeCell>` since it seems to be the most accurate.

integrations/ros2/tunnel-backend/src/relays/publish_subscribe.rs
orecham: Looked at this again, and I find it hard to see value in changing the machinery of the tunnel traits in this way.

None of these are optional. The `Option` field is managed by defining a separate `ReactiveTunnel` trait for tunnels to opt-in to reactive mode support and is thus always passed in.

integrations/ros2/tunnel-backend/src/payload.rs
elfenpiff: I will do it in my upcoming PR where I refactor the markers.

integrations/ros2/tunnel-backend/src/rcl/node.rs
```diff
@@ -0,0 +1,177 @@
...
+
+/// Builder for [`Node`].
+#[derive(Debug)]
+pub struct Builder<'a> {
+    name: &'a str,
+    namespace: &'a str,
+}
+
+impl<'a> Builder<'a> {
+    fn new(name: &'a str) -> Self {
```
elfenpiff: Please use here `&NodeName` as type instead of a `str`. When we use the strong types as early as possible, the rest of the code base becomes simpler instantly.

We no longer need the error enums: `InvalidName` and `InvalidNamespace`, we no longer need to test them and in this code we no longer need to add `fail` macros and check if the name is correct.

integrations/ros2/tunnel-backend/src/rcl/node.rs
```diff
@@ -0,0 +1,177 @@
...
+        Self {
+            name,
+            namespace: "",
+        }
+    }
+
+    /// Sets the node's namespace. Defaults to the root namespace.
+    pub fn namespace(mut self, namespace: &'a str) -> Self {
```
elfenpiff: Please use here `&Namespace` instead of `&str`

integrations/ros2/tunnel-backend/src/rcl/node.rs
```diff
@@ -0,0 +1,177 @@
...
+
+            Ok(Node { node, context })
+        }
+    }
+}
+
+impl Node {
+    /// Begins building a node with the given name. The namespace defaults to
+    /// the root namespace unless set via [`Builder::namespace`].
+    pub fn new(name: &str) -> Builder<'_> {
```
elfenpiff: Please use here `NodeName` instead of `str`

integrations/ros2/tunnel-backend/src/rcl/node.rs
```diff
@@ -0,0 +1,177 @@
...
+        Builder::new(name)
+    }
+
+    /// Build a publisher on this node for the given topic and typesupport.
+    pub fn publisher_builder(
+        self: Rc<Self>,
```
elfenpiff: Does this allow the user to call `node.publisher_builder()....` when the `node` is packed into an `Rc`.

This feels also weird from another standpoint since internals (that there is a shared state between nodes and ports) are now moved to the end user API.

Would you mind creating a separate `SharedNodeState` struct which is a member of the `Node` and all other structs that require the shared state and passing this around.
Then this method signature can be changed to `pub fn publisher_builder(&self,...`

integrations/ros2/tunnel-backend/src/rcl/publisher.rs
```diff
@@ -0,0 +1,161 @@
...
+/// Builder for [`Publisher`]. Created via [`Node::publisher_builder`].
+#[derive(Debug)]
+pub struct Builder<'a> {
+    node: Rc<Node>,
+    topic: &'a str,
+    type_support: TypeSupport,
+}
+
+impl<'a> Builder<'a> {
+    pub(crate) fn new(node: Rc<Node>, topic: &'a str, type_support: TypeSupport) -> Self {
```
elfenpiff: The topic should of type `&TopicName` , this would again simplify the code here and we can remove the `InvalidTopic` enum value.

integrations/ros2/tunnel-backend/src/rcl/node.rs
orecham: Yes that 100% makes sense.

integrations/ros2/tunnel-backend/src/rcl/node.rs
```diff
@@ -45,19 +45,36 @@ impl core::fmt::Display for CreationError {
...
+/// node, so it can be coupled to the context.
+struct NodeInner {
     node: Box<UnsafeCell<rcl_node_t>>,
     context: Box<UnsafeCell<rcl_context_t>>,
 }
 
+impl Drop for NodeInner {
+    fn drop(&mut self) {
+        unsafe {
+            let _ = rcl_node_fini(self.node.get());
```
elfenpiff: Could we still perform some error handling here and either print a `warn!` message or `fatal_panic!`. I looked up the documentation of those functions and when they fail it seems like we have screwed up our abstraction or the user did do some really weird stuff.

If we then fail here, we see it directly in future bug reports and know what to investigate.

integrations/ros2/tunnel-backend/src/rcl/node.rs
```diff
@@ -45,19 +45,36 @@ impl core::fmt::Display for CreationError {
...
+            let _ = rcl_shutdown(self.context.get());
+            let _ = rcl_context_fini(self.context.get());
+        }
+    }
+}
+
+/// A handle to an rcl node.
+///
+/// The node and its context stay alive until the last handle is dropped.
+#[derive(Clone)]
```
elfenpiff: I wouldn't implement `Clone` here since clone always gives the impression that you get a new fresh independent clone of the original and can do whatever you want with the new thing without changing the old thing. This is not the case here.

`Arc` and `Rc` are, semantically some exceptions but in this case it is an implementation detail.

integrations/ros2/tunnel-backend/src/rcl/node.rs
```diff
@@ -153,25 +172,15 @@ impl Node {
...
-        self: Rc<Self>,
-        topic: &str,
+    pub fn publisher_builder<'a>(
+        &self,
+        topic: &'a str,
         type_support: TypeSupport,
-    ) -> publisher::Builder<'_> {
-        publisher::Builder::new(self, topic, type_support)
+    ) -> publisher::Builder<'a> {
+        publisher::Builder::new(self.clone(), topic, type_support)
```
elfenpiff: The publisher builder should just use the `Rc<NodeInner>` as input argument and then this can be written as `self.inner.clone()`

integrations/ros2/tunnel-backend/src/relays/factory.rs
```diff
@@ -22,62 +21,62 @@ use crate::typesupport::TypeSupportRegistry;
 
 /// Factory for creating relay builders.
 #[derive(Debug)]
-pub struct Factory<S: Service> {
+pub struct Factory<'a, S: Service> {
```
elfenpiff, line 24: In iceoryx2 v0.1 I added the lifetime parameter to the shared state to the samples, port factories and all ports and it caused massive pains since the users were unable to store samples of the same type but different lifetimes in a `Vec` or any other container.

I just want to raise this here so that we do not run into the same issue. If we never see a use case where this type is stored in a container, then this is completely fine.

integrations/ros2/tunnel-backend/src/rcl/node.rs
elfenpiff: I thought some more about this and since ROS2 or iceoryx2 allow multiple `Node`s, implementing clone would be semantically wrong here.

Just think of this code snippet:

```rust
let node_1 = node::Builder::new("fuu");
let node_2 = node_1.clone();
let node_3 = node::Builder::new("bar");
```

You might think there are 3 nodes in the system, but actually there are just two nodes because of the shallow "clone".

Another alternative would be, to not use derive clone and implement a `Node::shallow_clone()` explicitly but I think we can remove the `#[derive(Clone)]` completely and use in the other abstractions only the `Rc<NodeInnner>` type.

integrations/ros2/tunnel-backend/src/rcl/publisher.rs
```diff
@@ -0,0 +1,152 @@
...
+        write!(f, "PublishError::{self:?}")
+    }
+}
+
+impl core::error::Error for PublishError {}
+
+/// Builder for [`Publisher`].
+#[derive(Debug)]
+pub struct Builder {
+    node: Node,
```
elfenpiff: Here and in the `Publisher` we could use `Rc<NodeInner>` which would also show other readers that the inner type is shared.

integrations/ros2/tunnel-backend/src/relays/publish_subscribe.rs
```diff
@@ -0,0 +1,173 @@
...
+
+impl<S: Service> RelayBuilder for Builder<'_, S> {
+    type CreationError = CreationError;
+    type Relay = Relay<S>;
+
+    // Endpoint creation simultaneously announces over the DDS SEDP.
+    fn create(self) -> Result<Self::Relay, Self::CreationError> {
+        let origin = "publish_subscribe::Relay::create";
+
+        let topic = fail!(from origin,
```
elfenpiff, line 131: Would you mind adding to more strong types, one for a ros2 service name and one for a ros2 type name. Those two structs could then implement `try_from` for the iox2 service name and the iox2 type name and in this could we could write

```rust
let topic: Ros2TopicName = fail!(from origin,
             when self.static_config.name().try_into(),
             with CreationError::InvalidServiceName,
             "Failed to map service name to a ROS2 topic");
```

When we have then those abstractions in place: Ros2TopicName and Ros2TypeName, then we can use them directly in the gateway config. This would have the nice side-effect that the wrong Ros2{Topic|Type}Name would directly lead to a parsing error of the config file and the user would see in which line they have to adjust something. And we do not have to do error handling later, since the config file parser and the strong types take care of everything.

integrations/ros2/tunnel-backend/src/backend.rs
```diff
@@ -0,0 +1,169 @@
...
+
+use crate::{
+    discovery::Discovery,
+    rcl,
+    relays::{Factory, event, publish_subscribe},
+    typesupport::TypeSupportRegistry,
+};
+
+/// The name of the ROS 2 node representing the tunnel.
+const NODE_NAME: &str = "iceoryx2_tunnel";
```
elfenpiff, line 28: Create in our Ros2 NodeName abstraction a function that is const and called new_unchecked and then you can use it here.
This makes the future usage of the  `NODE_NAME` much easier.

I did the same thing for our system types, see: https://github.com/eclipse-iceoryx/iceoryx2/blob/main/iceoryx2-bb/system-types/src/file_name.rs#L207

You can also apply this to all other strong types.

integrations/ros2/tunnel-backend/src/typesupport.rs
```diff
@@ -0,0 +1,193 @@
...
+/// compile time. ROS ships each package's typesupport as a compiled shared
+/// object (`lib<pkg>__rosidl_typesupport_c.so`) exporting a C getter function
+/// per type, so the only way to obtain the handle for a given name is to
+/// `dlopen` that library and resolve the getter symbol at runtime. Without
+/// the handle, rcl cannot create a publisher or subscription that DDS peers
+/// will match.
+///
+/// The handle points into the loaded library's memory, so the `Library` is
+/// kept alive here; it stays loaded as long as any clone is alive.
+#[derive(Debug, Clone)]
```
elfenpiff, line 68: Please do not implement `Clone` for a shallow clone. 

In my opinion, it is cleaner when everyone, who requires the type support to be alive uses the `Rc<TypeSupportInner>` struct. Otherwise, one might think that two distinct `TypeSupport` instances are alive whilst there is just one.

integrations/ros2/tunnel-backend/src/relays/publish_subscribe.rs
orecham: @elfenpiff `Typename` is already on the next branch, if we can ever get to it.

I would not add a ROS `ServiceName` until we needed it.

integrations/ros2/tunnel-backend/src/typesupport.rs
orecham: @elfenpiff I don't really follow your point. This is just a "handle to the type support".

integrations/ros2/tunnel-backend/src/backend.rs
orecham: @elfenpiff What is the benefit of doing that?

integrations/ros2/tunnel-backend/src/rcl/node.rs
orecham: @elfenpiff I removed the `Clone` derive.

integrations/ros2/tunnel-backend/src/backend.rs
elfenpiff: You wouldn't need to write this here anymore:

```rust
        let node_name = fail!(from origin,
            when rcl::NodeName::new(NODE_NAME),
            with CreationError::Node,
            "Invalid node name '{}'", NODE_NAME
        );
```

But to speed things up here see it as an optional suggestion.

integrations/ros2/tunnel-backend/src/typesupport.rs
elfenpiff: Since we have two different views here, could we agree that @elBoberido solves the tie?

integrations/ros2/tunnel-backend/src/relays/factory.rs
```diff
@@ -0,0 +1,84 @@
...
+use iceoryx2_services_tunnel_backend::{traits::RelayFactory, types::wake::WakeHandle};
+
+use crate::rcl;
+use crate::relays::{event, publish_subscribe};
+use crate::typesupport::TypeSupportRegistry;
+
+/// Factory for creating relay builders.
+#[derive(Debug)]
+pub struct Factory<'a, S: Service> {
+    node: Rc<rcl::Node>,
```
elfenpiff: Shouldn't this be a `Rc<NodeInner>` like in the Publisher and its Builder.

integrations/ros2/tunnel-backend/src/relays/factory.rs
```diff
@@ -0,0 +1,84 @@
...
+    type_registry: &'a TypeSupportRegistry,
+    /// Wake handle to be signaled by relays when new data arrives.
+    /// `None` when the backend was constructed in polled mode.
+    wake: Option<Arc<WakeHandle<local_threadsafe::Service>>>,
+    _phantom: core::marker::PhantomData<S>,
+}
+
+impl<'a, S: Service> Factory<'a, S> {
+    pub fn new(
+        node: Rc<rcl::Node>,
```
elfenpiff: Would you mind to use here a `Rc<NodeInner>` otherwise it is a bit like a `Rc<Rc<NodeInner>>` since the `Node` is just a `Rc<NodeInner>`

orecham: @elfenpiff I don't think I follow your motivations for these requests.

If we do use `NodeInner` here, then we can remove `Node` since nobody is using it, and we end back up with more or less the same code as before requesting the "shared state" struct.

I had thought the shared state would be an internal thing to the `rcl` module, which makes sense to me. Then having the `Node` be a cheap-to-clone handle which keeps a specific node instance alive in `rcl` and can be used to create other entities also makes sense. The `Rc` is only used here because of the request to not derive `Clone`.

BTW I saw this pattern in Zenoh: https://docs.rs/zenoh/latest/zenoh/session/index.html

integrations/ros2/tunnel-backend/src/typesupport.rs
orecham: This is just a minimal wrapper of `rcl` only for use by the tunnel. Is it worth to spend more time here? We can change it later.

integrations/ros2/tunnel-backend/src/backend.rs
orecham: https://github.com/eclipse-iceoryx/iceoryx2/issues/1761

elfenpiff (approved):

integrations/ros2/tunnel-backend/src/typesupport.rs
elBoberido: A little bit late to the party, but I would also prefer to have a `Rc<TypeSupportInner>`/`Rc<TypeSupport>` instead of hiding the `Rc` in a wrapper.

orecham: @elBoberido @elfenpiff I will revise it all later, let me get the functionality in, then I will do some additional iterations to make it pretty and conventional. I cannot do both at once :melting_face: .

PR #1758, opened 2026-06-22
[#1185] Set history per subscriber

### LINKED ISSUE #1185: Improvement of history and buffer size QoS settings

The current possibilities for history and buffer size QoS settings could be improved

- As of today you can set the history size for publishers only at the service level. This means that all publishers created for one service will have the same history size. It might be seen as a corner case, but the consequence is that it is not possible to have several publishers for a service with different history sizes.
- All subscribers created for a service will request the history size defined at the service level. There might be subscribers interested in historical data and others which aren't. Today, if history is enabled at the service level, all subscribers will get the available history
- If history is defined an the service level, the publishers will deliver the full history size to connecting subscribers, no matter what their buffer size is. It would be more efficient to only send messages up to the buffer size of the subscriber (or a requested history size when introducing this) as more samples would lead to an overflow of the subscriber buffer anyway. 
- The subscriber buffer overflow behavior can only be set on the service level. This means that all subscribers are forced to have the same overflow behavior.    

## Improvement Suggestion

- Be able to set the history individually per publisher
- Be able to set the requested history per subscriber
- Be able to define the overflow behavior per subscriber
- Only send historical messages based on subscriber's request

elBoberido:

## PR Reviewer Reminders

* Public API is documented

## References

Closes #1185

### DIFF
--- a/doc/release-notes/iceoryx2-unreleased.md
+++ b/doc/release-notes/iceoryx2-unreleased.md
@@ -15,6 +15,7 @@
 
 * [#820](https://github.com/eclipse-iceoryx/iceoryx2/issues/820) Allow restricting the tunnel to a configurable allowlist of services via `Config::services` and the `--service`/`-s` flag on `iox2 tunnel zenoh`
 * [#925](https://github.com/eclipse-iceoryx/iceoryx2/issues/925) Adjust event API and guarantee that events can be always delivered.
+* [#1185](https://github.com/eclipse-iceoryx/iceoryx2/issues/1185) Make history configurable per subscriber
 * [#1584](https://github.com/eclipse-iceoryx/iceoryx2/issues/1584) Introduce `Node::force_remove_service` to remove corrupted services manually.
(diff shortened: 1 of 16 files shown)


iceoryx2-cal/src/zero_copy_connection/common.rs
```diff
@@ -209,6 +209,7 @@ pub mod details {
         segment_details: RelocatableVec<SegmentDetails>,
         state: AtomicU8,
         max_borrowed_samples: usize,
+        history_request: usize,
```
elfenpiff: This shouldn't be part of the ZeroCopyConnection. This is comparable to a zero-copy unix domain socket. Could you move this whole history request part into the iceoryx2 crate into publish-subscribe.

For context, when adding this here, the logic would also be available for request response for the client to server and server to client (response) communication and it wouldn't make sense in this context.

This layer does not yet know things about service-oriented architecture, quality of service settings or messaging patterns. It is just a safe zero-copy unix domain socket and the settings are only settings required to enable such a safe zero-copy communication path.

elfenpiff (approved):

Issue #1761, opened 2026-06-24
ROS 2 Gateway: Post-implementation


orecham:
## Brief Description

This issue shall track tweaks / optimizations to be made to the implementation once end-to-end functionality is complete.

## Items

1. [x] Add `new_unchecked` to name validation structs
    * https://github.com/eclipse-iceoryx/iceoryx2/pull/1744#discussion_r3450351111
1. [x] Revise use of Handles and Rc
    * https://github.com/eclipse-iceoryx/iceoryx2/pull/1744#discussion_r3474211491 
1. [x] Decouple tunnel backend machinery from the `StaticConfig` of a service
    * https://github.com/eclipse-iceoryx/iceoryx2/pull/1768#discussion_r3482073971
1. [x] Consolidate `DiscoveryEvent` and `DiscoveryUpdate` (are both needed?)
1. [ ] ~~Organize tests in ROS 2 `tunnel-backend`~~
1. [ ] ~~Use `clap` for example applications with CLI~~
    * Examples requiring CLI replaced with actual `iox2` CLI
1. [x] Revise handling of discovered topics/services with no mapping
    * https://github.com/eclipse-iceoryx/iceoryx2/pull/1822#discussion_r3632448680 
1. [x] Consolidate allowlist handling if possible
    * https://github.com/eclipse-iceoryx/iceoryx2/pull/1850#discussion_r3657337563
    * Allow list provided in multiple locations for different reasons, needs to be consolidated where possible:
        * GatewayConfig - filters what the tunnel will bridge by iceoryx2 service name
        * BackendConfig - preloads typesupport and supplies the candidate set for ROS-side discovery
        * Mapping - derives each side's service/topic description from the other, unmappable inputs are dropped
1. [x] Rename `IntrospectionTranslator` to something more descriptive
    * https://github.com/eclipse-iceoryx/iceoryx2/pull/1850#discussion_r3689388442
1. [x] Move the `RosHeader` to a common location
1. [x] Link to ROS 2 examples from main examples directory
    * https://github.com/eclipse-iceoryx/iceoryx2/pull/1856#discussion_r3711386713
1. [x] Add release notes

PR #1768, opened 2026-06-25
[#1742] Implement ros2 tunnel recv path

orecham:
## Notes for Reviewer

Implements the recv path for the ROS 2 tunnel backend.

Remaining capabilities and CI to be merged in subsequent pull requests. See the README in integrations/ros2/tunnel-backend for the development status.

**NOTE:** The usage of "handle" structs and `Rc` needs to be revised - the big picture got muddled in the midst of addressing review findings. This will be done in a **follow-up** in order to keep this PR scoped to the receive path.

## PR Reviewer Reminders

* Public API is documented

## References

Relates #1742


iceoryx2/src/service/static_config/mod.rs
```diff
@@ -115,6 +115,32 @@ impl StaticConfig {
...
 
+    /// Creates the [`StaticConfig`] of a publish-subscribe
+    /// [`Service`](crate::service::Service) with the provided message type
+    /// details; all other values are taken from the defaults in `config`,
+    /// but this might need to be revised to be more precise.
+    ///
+    /// Intended for infrastructure that describes services it does not
+    /// create itself, e.g. tunnels. Not part of the stable public API.
+    #[doc(hidden)]
+    pub fn __internal_new_publish_subscribe_with_details<Hasher: Hash>(
```
orecham, line 126: Will be removed after tunnel backend machinery is refactored - immediately after this PR.

https://github.com/eclipse-iceoryx/iceoryx2/issues/1761

integrations/ros2/tunnel-backend/src/rcl/names.rs
```diff
@@ -92,7 +93,19 @@ impl NodeName {
             );
         }
 
-        Ok(Self(into_cstring(name)))
+        Ok(Self(owned(name)))
+    }
+
+    /// Wraps an already-valid node name without re-checking it.
+    ///
+    /// The caller must guarantee `name` is a well-formed ROS 2 node name.
```
elfenpiff, line 133: All unchecked methods should use the `unsafe` keyword since the user is able to break the contract and cause any kind of trouble.

The line 101 should be a bullet point in the `# Safety` section every unsafe function requires.

integrations/ros2/tunnel-backend/src/rcl/names.rs
```diff
@@ -92,7 +93,19 @@ impl NodeName {
...
+    pub fn new_unchecked(name: &str) -> Self {
+        Self(owned(name))
+    }
+
+    /// Wraps a static, already-valid node name at compile time.
```
elfenpiff, line 138: Same here, please make it unsafe and add a safety section. See this as an example: https://doc.rust-lang.org/beta/std/alloc/struct.Layout.html#method.from_size_align_unchecked

integrations/ros2/tunnel-backend/src/rcl/names.rs
```diff
@@ -135,7 +148,19 @@ impl NodeNamespace {
...
         }
 
-        Ok(Self(into_cstring(namespace)))
+        Ok(Self(owned(namespace)))
+    }
+
+    /// Wraps an already-valid namespace without re-checking it.
+    ///
+    /// The caller must guarantee `namespace` is a well-formed ROS 2 node namespace.
+    pub fn new_unchecked(namespace: &str) -> Self {
```
elfenpiff: Please add `unsafe`

integrations/ros2/tunnel-backend/src/rcl/names.rs
```diff
@@ -135,7 +148,19 @@ impl NodeNamespace {
...
+        Self(owned(namespace))
+    }
+
+    /// Wraps a static, already-valid namespace at compile time.
+    pub const fn new_static_unchecked(namespace: &'static CStr) -> Self {
```
elfenpiff: Please add `unsafe`

integrations/ros2/tunnel-backend/src/rcl/names.rs
```diff
@@ -179,12 +204,93 @@ impl TopicName {
...
 
-        Ok(Self(into_cstring(topic)))
+        Ok(Self(owned(topic)))
+    }
+
+    /// Wraps an already-valid topic name without re-checking it.
+    ///
+    /// The caller must guarantee `topic` is a well-formed ROS 2 topic name,
+    /// e.g. it came straight from an rcl graph query.
+    pub fn new_unchecked(topic: &str) -> Self {
```
elfenpiff: Please add `unsafe`

integrations/ros2/tunnel-backend/src/rcl/names.rs
```diff
@@ -179,12 +204,93 @@ impl TopicName {
...
+        Self(owned(topic))
+    }
+
+    /// Wraps a static, already-valid topic name at compile time.
+    pub const fn new_static_unchecked(topic: &'static CStr) -> Self {
```
elfenpiff: Please add `unsafe`

integrations/ros2/tunnel-backend/src/rcl/names.rs
```diff
@@ -179,12 +204,93 @@ impl TopicName {
...
+        }
+
+        Ok(Self(owned(type_name)))
+    }
+
+    /// Wraps an already-valid type name without re-checking it.
+    ///
+    /// The caller must guarantee `type_name` is a well-formed ROS 2 type name,
+    /// e.g. it came straight from an rcl graph query.
+    pub fn new_unchecked(type_name: &str) -> Self {
```
elfenpiff: Please add `unsafe`

integrations/ros2/tunnel-backend/src/rcl/names.rs
```diff
@@ -179,12 +204,93 @@ impl TopicName {
...
+        Self(owned(type_name))
+    }
+
+    /// Wraps a static, already-valid type name at compile time.
+    pub const fn new_static_unchecked(type_name: &'static CStr) -> Self {
```
elfenpiff: Please add `unsafe`

integrations/ros2/tunnel-backend/src/rcl/names.rs
```diff
@@ -92,7 +93,19 @@ impl NodeName {
...
+
+    /// Wraps an already-valid node name without re-checking it.
+    ///
+    /// The caller must guarantee `name` is a well-formed ROS 2 node name.
+    pub fn new_unchecked(name: &str) -> Self {
+        Self(owned(name))
+    }
+
+    /// Wraps a static, already-valid node name at compile time.
+    pub const fn new_static_unchecked(name: &'static CStr) -> Self {
```
elfenpiff: Why are all `new_static_unchecked` using a `CStr` and the `new_unchecked` uses `str`. Could we rename those methods to `from_c_str_unchecked`? Then it would be more idiomatic in the rust universe.

integrations/ros2/tunnel-backend/src/rcl/subscription.rs
```diff
@@ -0,0 +1,342 @@
...
+        }
+    }
+
+    pub fn create(self) -> Result<Subscription, CreationError> {
+        let origin = "Subscription::Builder::create";
+
+        unsafe {
+            let mut subscription = Box::new(rcl_get_zero_initialized_subscription());
+            let mut options = rcl_subscription_get_default_options();
+            // Prevent loopback.
```
elfenpiff, line 124: I understand why you did this but this (so that the gateway does not receive its own data) is actually something we explicitly want? What if you use a ROS2 Gateway/Tunnel to communicate with other local ROS2 nodes since you have not yet ported them to iceoryx2.

Is there the possibility to throw away the messages since we can identify the origin like in iceoryx2 with the publisher id in the sample header?

In the long-term this should be configurable by the user but in the first shot I would set it to always `false` and add this as a configurable optimization for later.

integrations/ros2/tunnel-backend/src/rcl/subscription.rs
```diff
@@ -0,0 +1,342 @@
...
+                &mut message,
+                &mut message_info,
+                core::ptr::null_mut(),
+            )
+        };
+        if ret == RCL_RET_SUBSCRIPTION_TAKE_FAILED as i32 {
+            return Ok(None);
+        }
+        if ret != RCL_RET_OK as i32 {
+            // A consumed loan closure means the buffer request was declined,
```
elfenpiff, line 244: This is a horrible way of handling errors ... Nothing you can do about it.

integrations/ros2/tunnel-backend/src/backend.rs
```diff
@@ -25,17 +25,59 @@ use crate::{
...
+    pub(crate) type_name: rcl::TypeName,
 }
 
+impl TopicConfig {
+    /// Creates a config entry, validating the ROS 2 topic name and message
+    /// type name.
+    ///
+    /// * The topic must be a valid ROS 2 topic name
+    /// * The type name must have the form `package/msg/Message`
+    pub fn new(topic: &str, type_name: &str) -> Result<Self, TopicConfigError> {
```
elfenpiff: Don't we have now strong types for topic and type name? Couldn't we use them here and with this we could get rid of returning a result and have error handling in here.

integrations/ros2/tunnel-backend/src/rcl/names.rs
orecham: Was not aware of the convention, thanks!

integrations/ros2/tunnel-backend/src/rcl/subscription.rs
orecham: > What if you use a ROS2 Gateway/Tunnel to communicate with other local ROS2 nodes since you have not yet ported them to iceoryx2.

The RMW documentation is a little bit vague, but my interpretation is that the scenario you describe will work fine:

```
  /// If true then the middleware should not deliver data from local publishers.
  /**
   * This setting is most often used when data should only be received from
   * remote nodes, especially to avoid "double delivery" when both intra- and
   * inter- process communication is taking place.
   *
   * The definition of local means that in the same context.
   */
  bool ignore_local_publications;
```

Link: https://github.com/ros2/rmw/blob/rolling/rmw/include/rmw/types.h#L172

I take "in the same context" to mean that which is represented by the handle  `rmw_context_s`. This means only the gateway will not receive it's own publications. Other ROS applications (same host, or even same process) should have their own context and thus their publications should still be received.

integrations/ros2/tunnel-backend/src/backend.rs
orecham: Yes I deliberately did not use them here, my thinking was:

* The configuration will be eventually coming from a file, so they will be `&str`
* The validated types are internal to the `rcl` module of the implementation - other applications should not be constructing them

elfenpiff: No, the strong types can also implement `Serialize` and `Deserialize` and then we can use strong types even in the config. We do this already today.

It has also a usability advantage since the deserializer points the user directly to the line in the config where the string violates the contract.

If you don't mind could you please use them here.

elfenpiff (approved):

PR #1822, opened 2026-07-14
[#1742] Add mapping implementations for ROS 2 tunnel

orecham:
## Notes for Reviewer

Adds the following `Mapping` implementations:
* `PrefixMapping` - dynamic mapping of services with the prefix `ros2://topics/` to ROS 2 topics
* `StaticMapping` - static mapping of services to topics explicitly specified by configuration

## PR Reviewer Reminders

* Public API is documented

## References

Relates #1742


integrations/ros2/tunnel-backend/examples/tunnel_static_mapping.rs
```diff
@@ -0,0 +1,63 @@
...
+const POLL_INTERVAL: Duration = Duration::from_millis(100);
+const DEFAULT_MAPPING_FILE: &str = "examples/mapping.toml";
+
+fn main() -> Result<(), Box<dyn core::error::Error>> {
+    set_log_level_from_env_or(LogLevel::Info);
+
+    let path = std::env::args()
+        .nth(1)
+        .unwrap_or_else(|| DEFAULT_MAPPING_FILE.to_string());
+
```
elBoberido, line 38: Food for thought. We already use `clap` in other examples. What do you think about also using it here too?

integrations/ros2/tunnel-backend/src/rcl/qos.rs
```diff
@@ -0,0 +1,132 @@
...
+    rmw_qos_profile_t, rmw_qos_reliability_policy_e, rmw_time_s,
+};
+
+use crate::qos::{Durability, History, Liveliness, QosProfile, Reliability};
+
+/// Matches `RMW_DURATION_INFINITE`. Represents unbounded.
+const INFINITE: rmw_time_s = rmw_time_s {
+    sec: 9223372036,
+    nsec: 854775807,
+};
```
elBoberido, line 30: Can you add a note that this weird numbers originate from i64::MAX which is `9223372036854775807` and the nsecs are the lower 9 decimal digits and the secs the remaining decimal digits, making infinite representable as i64::MAX in nanoseconds?

integrations/ros2/tunnel-backend/src/rcl/qos.rs
```diff
@@ -0,0 +1,132 @@
...
+    rmw_time_s {
+        sec: duration.as_secs(),
+        nsec: duration.subsec_nanos() as u64,
+    }
+}
+
+fn duration(time: &rmw_time_s) -> Option<Duration> {
+    let unset = time.sec == 0 && time.nsec == 0;
+    let infinite = (time.sec, time.nsec) >= (INFINITE.sec, INFINITE.nsec);
+    (!unset && !infinite).then(|| Duration::from_secs(time.sec) + Duration::from_nanos(time.nsec))
```
elBoberido, line 135: I don't see this often, but I like the style

integrations/ros2/tunnel-backend/src/qos.rs
```diff
@@ -0,0 +1,211 @@
...
+    /// Matches the ROS 2 default profile (`rmw_qos_profile_default`).
+    fn default() -> Self {
+        Self {
+            history: History::KeepLast(10),
+            reliability: Reliability::Reliable,
+            durability: Durability::Volatile,
+            deadline: None,
+            lifespan: None,
+            liveliness: Liveliness::SystemDefault,
+            liveliness_lease_duration: None,
```
elBoberido, line 72: Is this something we could get from some `r2r_rcl` constants or have a test that these defaults do match the ROS 2 default profile?

Ignore this, if the default ROS 2 profile is guaranteed to never change.

integrations/ros2/tunnel-backend/src/testing.rs
elBoberido, line 1: Does this need to be part of the production crate?

integrations/ros2/tunnel-backend/src/testing.rs
```diff
@@ -0,0 +1,99 @@
...
+            .expect("failed to query the ROS 2 graph")
+            .into_iter()
+            .filter(|(name, _)| name.as_str() == topic)
+            .flat_map(|(_, types)| types.into_iter().map(|name| name.as_str().to_string()))
+            .collect()
+    }
+}
+
+/// Polls `condition` until it holds or `timeout` elapses.
+pub fn wait_until(timeout: Duration, mut condition: impl FnMut() -> bool) -> bool {
```
elBoberido: Would it make sense to use the `AdaptiveWait::wait_while` instead?

integrations/ros2/tunnel-backend/src/discovery.rs
```diff
@@ -61,36 +53,118 @@ impl core::error::Error for AnnouncementError {}
...
+        &self,
+        topic: &TopicName,
+        type_name: &TypeName,
+        process_discovery: &mut F,
+    ) -> Result<(), DiscoveryError> {
+        let origin = "Discovery::discover_added";
+
+        let topic_description = self.topic_description(topic, type_name)?;
+        let Some(service_description) = self.mapping.local::<S>(&topic_description) else {
+            return Ok(());
```
elBoberido, line 140: I don't really understand why `Ok` is returned. Although this is a private function, it would help to have a small description to understand what the function is supposed to do or at least a comment why `Ok` is returned here, since the behavior is a bit unexpected.

integrations/ros2/tunnel-backend/src/mapping/static_mapping/config.rs
```diff
@@ -0,0 +1,156 @@
...
+    ) -> Result<PortSettings<T>, D::Error> {
+        T::deserialize(deserializer).map(PortSettings::Value)
+    }
+
+    pub fn local_defaults<T>() -> PortSettings<T> {
+        PortSettings::LocalDefaults
+    }
+
+    pub fn is_local_defaults<T>(settings: &PortSettings<T>) -> bool {
+        matches!(settings, PortSettings::LocalDefaults)
```
elBoberido, line 154: Wondering why not just
```suggestion
        settings == PortSettings::LocalDefaults
```

integrations/ros2/tunnel-backend/src/mapping/static_mapping/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+            .map(|entry| TopicConfig {
+                topic: entry.ros2.topic.clone(),
+                type_name: entry.ros2.type_name.clone(),
+            })
+            .collect()
+    }
+
+    /// The entry mapping the service specified by `description`, if present.
+    fn entry(&self, description: &ServiceDescription) -> Option<&Entry> {
+        if !matches!(description.pattern, PatternDescription::PublishSubscribe(_)) {
```
elBoberido, line 114: Similar here, why not just
```suggestion
        if description.pattern != PatternDescription::PublishSubscribe(_) {
```

integrations/ros2/tunnel-backend/src/mapping/prefix_mapping.rs
```diff
@@ -0,0 +1,148 @@
...
+            }),
+        ))
+    }
+}
+
+/// Convert a conventional iceoryx2 service name into a ROS 2 topic name.
+fn topic(service_name: &str) -> Option<&str> {
+    let topic = service_name.strip_prefix(TOPIC_PREFIX)?;
+    if !topic.starts_with('/') || topic.len() == 1 {
+        return None;
```
elBoberido: Can you elaborate a bit on this? Do we have to enforce that the iceoryx2 service names have to start with a `/`?

elBoberido (commented):
Reviewed everything but the tests. They will follow soon. Just that you already get some feedback.

integrations/ros2/tunnel-backend/tests/prefix_mapping_tests.rs
```diff
@@ -0,0 +1,298 @@
...
+fn ignores_unprefixed_names_and_invalid_type_names() {
+    let sut = PrefixMapping;
+
+    for (name, type_name) in [
+        ("My/Funk/ServiceName", "std_msgs/msg/String"),
+        ("ros2://topics", "std_msgs/msg/String"),
+        ("ros2://topics/", "std_msgs/msg/String"),
+        ("iox2://something", "std_msgs/msg/String"),
+        ("/chatter", "std_msgs/msg/String"),
+        ("ros2://topics/chatter", "NotARosTypeName"),
```
elBoberido, line 80: Should this also be tested?
```
("ros2://some/thing", "std_msgs/msg/String"),
```

integrations/ros2/tunnel-backend/tests/prefix_mapping_tests.rs
```diff
@@ -0,0 +1,298 @@
...
+    let service_description = service_description_with_default_settings(
+        "ros2://topics/Camera/FrontRight",
+        "sensor_msgs/msg/Image",
+    );
+
+    let topic_description = sut
+        .remote(&service_description)
+        .expect("service maps to a topic");
+
+    assert_eq!(topic_description.topic.as_str(), "/Camera/FrontRight");
```
elBoberido: Using `assert_that` would give some additional information when a test fails. It looks like you used `assert` or `assert_eq` everywhere.

integrations/ros2/tunnel-backend/tests/static_mapping_tests.rs
```diff
@@ -0,0 +1,215 @@
...
+    }
+}
+
+#[test]
+fn example_config_is_valid() {
+    let config: Config = toml::from_str(include_str!("../examples/mapping.toml")).unwrap();
+
+    assert!(!config.entries.is_empty());
+    assert!(StaticMapping::new(config).is_ok());
+}
```
elBoberido, line 64: Food for thought. I think I initially misunderstood to idea behind this test. What do you think about a slightly different test name like `validate_config_used_for_examples`?

integrations/ros2/tunnel-backend/src/mapping/prefix_mapping.rs
elBoberido: Okay, forget about it. I think I misunderstood this. It is only related to ROS 2 topic names, which need to start with a slash, right?

elBoberido (commented):
Looks good. Just some minor thing in general.

integrations/ros2/tunnel-backend/examples/tunnel_static_mapping.rs
orecham: There's no `clap` in this workspace yet. Were you asking for consistency or e.g. you think the automatic `--help` output is useful?

integrations/ros2/tunnel-backend/src/rcl/qos.rs
orecham: This is news to me (I did not look further into it), the value was taken from the RMW (which I point to in the existing comment): https://github.com/ros2/rmw/blob/rolling/rmw/include/rmw/time.h#L54

I can add the extra information if it would be useful. What was the reason you had in mind?

integrations/ros2/tunnel-backend/src/qos.rs
orecham: It will not change for the released versions being supported currently. It could change in future versions but I would put that as an unlikely breaking change...

Added a test.

integrations/ros2/tunnel-backend/src/testing.rs
orecham: The issue is that the utilities depend on rcl wrappers that are not publicly exported. This follows the same convention as e.g. `iceoryx2-cal` and `iceoryx2` crates.

integrations/ros2/tunnel-backend/src/testing.rs
orecham: Not sure, do you think so? What's your reasoning? Reduced duplication or something functional?
I don't think adaptive wait helps functionality, maybe even makes it worse. 

I replaced the function with this generic trait to reduce duplication and unlock possibility of conformance tests later: https://github.com/eclipse-iceoryx/iceoryx2/blob/main/iceoryx2-services/tunnel-backend/src/traits/testing.rs#L26

integrations/ros2/tunnel-backend/src/mapping/static_mapping/config.rs
orecham: What you suggest doesn't compile without adding more strict trait bounds. The `matches!` checks the variant with less constraints.

integrations/ros2/tunnel-backend/src/mapping/static_mapping/mod.rs
orecham: Does not compile. You can't do this with the generic `PublishSubscribe(_)`.

integrations/ros2/tunnel-backend/src/mapping/prefix_mapping.rs
orecham: Yes this is attempting to convert an iceoryx2 service in the form `ros2://topics/my_namespace/my_topic` into a valid ROS 2 topic.

That being said, I notice now that this implementation is stale (long PR delays) and there is actually a `TopicName` struct that can be used for more comprehensive validation. Changed.

integrations/ros2/tunnel-backend/examples/tunnel_static_mapping.rs
elBoberido: Just for consistency. Since this has a default value, one does not need to know the parameter to just run the example. No need to action.

integrations/ros2/tunnel-backend/src/rcl/qos.rs
elBoberido: I guess just my OCD. I was wondering if these numbers are indeed infinity in RMW or maybe some copy & paste error. Having the explanation right there would help everybody who is asking this question in the future.

integrations/ros2/tunnel-backend/src/discovery.rs
elBoberido: The comments still do not shed some light on `Ok(())` equals skip and that returning `Ok(())` is not a programming error and an `Err` should be returned instead?

The reason I'm asking is that, depending on the viewpoint, the current behavior is a silent error if e.g. someone has a typo in a topic name. It is okay if this is the intention but I cannot derive this conclusion from the current code and documentation.

elBoberido (dismissed):
Looks good. Just some documentation that could be added but that can also be done in a follow up.

integrations/ros2/tunnel-backend/examples/tunnel_static_mapping.rs
orecham: Added to tracking issue: https://github.com/eclipse-iceoryx/iceoryx2/issues/1761

The examples get an overhaul in 1-2 PRs from this one, so I will add it in then.

integrations/ros2/tunnel-backend/src/discovery.rs
orecham: Hmm, it depends on the mapping whether or not this is an error. For instance, for the prefix mapping it deliberately wants to skip any service names not following its convention, which is not an error.

We could add a log to the `StaticMapping`, but that could annoying when deliberately excluding some topics since they would all noisily log that they are not mapped.

elBoberido: The question is whether the return value should then be a `Ok(())` or a `Ok(Foo)` with `Foo` either being a `bool` that tells that the discovery found the topic or not. Or an enum like

```rust
enum DiscoveryResult {
    Found,
    NotFound,
}
```

But maybe I'm overthinking this.

elBoberido (approved):

integrations/ros2/tunnel-backend/src/discovery.rs
orecham: I added it to https://github.com/eclipse-iceoryx/iceoryx2/issues/1761

Will tackle it in a small follow-up.

PR #1830, opened 2026-07-22
[#1797] Reclaim disconnected client connections

### LINKED ISSUE #1797: Request-response server with enable_fire_and_forget_requests(false) leaks a connection_storage slot per client disconnect -> fatal_panic! "Internal connection storage capacity exceeded"
> Analysis, reproduction and suggested fix written by **Claude Code**.
> (FAQ checked — the *"ExceedsMaxSupported"* entry is a different, retriable
> error; nothing there covers this `fatal_panic!`.)

## Required information

* **OS:** Ubuntu 24.04, Linux 6.17.0-35-generic, x86_64
* **Rust / Cargo:** 1.96.1
* **iceoryx2 version:** `main` (workspace `0.9.999`)

**Observed behaviour**

A `Server` on a request-response service built with
`enable_fire_and_forget_requests(false)` never reclaims a client's
`connection_storage` slot on disconnect. After
`server_expired_connection_buffer + max_clients` client turnovers the next
`connection_storage.insert()` returns `None` and the server aborts:

```text
thread 'main' panicked at iceoryx2/src/port/details/receiver.rs:247:
This should never happen! Internal connection storage capacity exceeded.
```

It is a deterministic, uptime-cumulative crash of the **server** for any
long-running service whose clients come and go (default buffer `128`, so it
crashes after ~`128 + max_clients` turnovers).

**Root cause**

Every `connection_storage.remove()` in the receiver is inside
`if let Some(to_be_removed_connections)` (`receiver.rs:257`), and the server sets
that field to `None` when fire-and-forget is off (`server.rs:391-401`). So
`prepare_connection_removal` becomes a no-op and stale connections are never
reclaimed, while the SlotMap capacity stays `server_expired_connection_buffer +
max_clients` (`server.rs:374`).

**Minimal reproduction** (single process; the buffer is shrunk only so it
overflows fast — the default `128` crashes the same way, later):

```rust
use iceoryx2::prelude::*;

    let mut config = Config::default();
    config.defaults.request_response.server_expired_connection_buffer = 4;

    let node = NodeBuilder::new().config(&config).create::<ipc::Service>()?;
    let service = node
        .service_builder(&"repro/f23".try_into()?)
        .request_response::<u64, u64>()
        .enable_fire_and_forget_requests(false) // trigger; `true` never crashes
        .max_clients(2)
        .open_or_create()?;
    let server = service.server_builder().create()?;

    for i in 0..60u64 {
        let client = service.client_builder().create()?;
        let _pending = client.send_copy(i)?;
        while let Some(active) = server.receive()? { drop(active); }
        drop(client); // next client reuses the freed client-list index
    }
    Ok(())
}
```

With buffer `4` + `max_clients 2` (capacity `6`) it panics on **iteration 6**;
flipping the flag to `true` runs all 60 iterations cleanly.

**Expected behaviour**

Stale connections are reclaimed from `connection_storage` on disconnect
regardless of `enable_fire_and_forget_requests`, so the server does not
accumulate dead connections and crash.

**Suggested fix**

Allocate `to_be_removed_connections` for the server unconditionally (drop the
`else { None }` at `server.rs:391-401`); the existing borrow-aware removal logic
then runs and the flag only affects the receive-side drop (`server.rs:799`).
Removal must stay borrow-safe — a disconnected client may still have a borrowed
`ActiveRequest`, and removing it unmaps live shared memory — which is exactly
what the buffer's defer path already handles.

## Relationship to #989

#989 (*Request Response Advanced Connection Handling*) touches the same subsystem
but does not cover this leak or the `enable_fire_and_forget_requests == false`
gating; implementing it would not fix this.

xiao-yang25:
## Notes for Reviewer

When fire-and-forget requests are disabled, the server does not allocate
`Receiver::to_be_removed_connections`. Since disconnected connections are
reclaimed through this buffer, every disconnected client permanently consumes
a connection storage slot.

After enough client reconnects, the connection storage reaches its capacity
and the server terminates with `Internal connection storage capacity exceeded`.

This change:

* Allocates the server's deferred connection-removal buffer regardless of
  whether fire-and-forget requests are enabled.
* Reuses the existing borrow-aware removal path, ensuring connections with
  outstanding `ActiveRequest`s remain deferred until they can be removed safely.
* Adds a conformance test that reconnects more clients than the combined
  `max_clients` and expired connection buffer capacity.
* Updates the unreleased bugfix notes.

* [ ] Consider switching the PR to a draft
* [x] Branch follows the naming format
* [x] Commit messages follow the project guideline
    * [x] Commit messages have the issue ID
* [x] Tests follow the best practice for testing
* [x] Changelog updated in the unreleased section

## PR Reviewer Reminders

* Commits are properly organized and messages follow the guideline
* Unit tests have been written for the corrected behavior

## References

Closes #1797

### DIFF
--- a/iceoryx2/src/port/server.rs
+++ b/iceoryx2/src/port/server.rs
@@ -385,17 +385,13 @@ impl<
             enable_safe_overflow: static_config.enable_safe_overflow_for_requests,
             buffer_size: static_config.max_active_requests_per_client,
             tagger: CyclicTagger::new(),
-            to_be_removed_connections: if static_config.enable_fire_and_forget_requests {
-                Some(UnsafeCell::new(
-                    PolymorphicVec::new(
-                        HeapAllocator::global(),
-                        number_of_to_be_removed_connections,
-                    )
-                    .expect("Heap allocator provides memory."),
-                ))
-            } else {
-                None
-            },
+            to_be_removed_connections: Some(UnsafeCell::new(
+                PolymorphicVec::new(
+                    HeapAllocator::global(),
+                    number_of_to_be_removed_connections,
+                )
+                .expect("Heap allocator provides memory."),
+            )),
             degradation_handler: server_factory.request_degradation_handler,
             number_of_channels: 1,
             connection_storage: UnsafeCell::new(SlotMap::new(number_of_connections)),
--- a/iceoryx2/src/port/server.rs
+++ b/iceoryx2/src/port/server.rs
@@ -385,13 +385,10 @@ impl<
             enable_safe_overflow: static_config.enable_safe_overflow_for_requests,
             buffer_size: static_config.max_active_requests_per_client,
             tagger: CyclicTagger::new(),
-            to_be_removed_connections: Some(UnsafeCell::new(
-                PolymorphicVec::new(
-                    HeapAllocator::global(),
-                    number_of_to_be_removed_connections,
-                )
-                .expect("Heap allocator provides memory."),
-            )),
+            to_be_removed_connections: UnsafeCell::new(
+                PolymorphicVec::new(HeapAllocator::global(), number_of_to_be_removed_connections)
+                    .expect("Heap allocator provides memory."),
+            ),
             degradation_handler: server_factory.request_degradation_handler,
             number_of_channels: 1,
             connection_storage: UnsafeCell::new(SlotMap::new(number_of_connections)),
--- a/doc/release-notes/iceoryx2-unreleased.md
+++ b/doc/release-notes/iceoryx2-unreleased.md
@@ -45,6 +45,7 @@
 * [#1777](https://github.com/eclipse-iceoryx/iceoryx2/issues/1777) Fix service root folder creation named concept of iceoryx2-cal fixing execution on Windows platform.
 * [#1786](https://github.com/eclipse-iceoryx/iceoryx2/issues/1786) Disable transport_compression feature in Zenoh.
 * [#1792](https://github.com/eclipse-iceoryx/iceoryx2/issues/1792) Set key eq comparison function in language bindings for blackboard opener.
+* [#1797](https://github.com/eclipse-iceoryx/iceoryx2/issues/1797) Reclaim disconnected request-response client connections when fire-and-forget requests are disabled.
 * [#1800](https://github.com/eclipse-iceoryx/iceoryx2/issues/1800) iceoryx2-cxx: CleanupState is defined in global namespace
 * [#1807](https://github.com/eclipse-iceoryx/iceoryx2/issues/1807) Fix generated C FFI strings for `UPPER_SNAKE_CASE` enum variants.
 * [#1810](https://github.com/eclipse-iceoryx/iceoryx2/issues/1810) Make mgmt segment globally accessible
--- a/iceoryx2/conformance-tests/src/service_request_response.rs
+++ b/iceoryx2/conformance-tests/src/service_request_response.rs
@@ -1728,4 +1728,39 @@ pub mod service_request_response {
         assert_that!(response, is_some);
         assert_that!(*response.unwrap(), eq 1);
     }
+
+    #[conformance_test]
+    pub fn reconnecting_clients_do_not_exhaust_server_connection_storage_when_fire_and_forget_is_disabled<
+        S: Service,
+    >() {
+        const MAX_CLIENTS: usize = 2;
+        const EXPIRED_CONNECTION_BUFFER: usize = 4;
+        const ITERATIONS: usize = 3 * (MAX_CLIENTS + EXPIRED_CONNECTION_BUFFER);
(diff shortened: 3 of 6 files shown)


iceoryx2/src/port/server.rs
```diff
@@ -385,17 +385,13 @@ impl<
...
-                    PolymorphicVec::new(
-                        HeapAllocator::global(),
-                        number_of_to_be_removed_connections,
-                    )
-                    .expect("Heap allocator provides memory."),
-                ))
-            } else {
-                None
-            },
+            to_be_removed_connections: Some(UnsafeCell::new(
```
elBoberido: Looks good to me.

@elfenpiff I guess you overlooked my question in the issue. 

> Maybe @elfenpiff can shed some light why the `to_be_removed_connections` are currently only used when fire and forget is active. Is is also the only place where this option is used. A client should not be able to crash the server, so we either need to use the `to_be_removed_connections` buffer all the time, or abort with an error code. But with the latter, one could also create a DOS, so that's also not desirable.

With the changes from this PR, the `to_be_removed_connections` flag would be obsolete. Can we just remove it?

xiao-yang25: Yes. I checked all `Receiver` construction sites, and with this change the
client, subscriber, and server always allocate `to_be_removed_connections`.

I will make the field non-optional and remove the redundant `if let Some(...)`
branches. The connection removal buffer itself will remain, since it is still
required for borrow-aware cleanup. Only the obsolete `Option` wrapper will be
removed.

`enable_fire_and_forget_requests` will remain because it still controls whether
requests from disconnected clients can be returned by the server receive APIs.

xiao-yang25: Implemented in 5fe6a3e96. `Receiver::to_be_removed_connections` is now
non-optional, and the redundant optional branches have been removed.

elBoberido: @xiao-yang25 right, I overlooked the usage in the server. So the flag must remain. The changes look okay but it would be good if @elfenpiff could also take a look at it.

elBoberido: @xiao-yang25 I guess you also need to run `cargo fmt --all` to please the CI.

xiao-yang25: Understood. The flag remains unchanged. I applied the rustfmt changes reported
by CI in 38165cce1.

elfenpiff: @xiao-yang25 @elBoberido 

Sorry for the late response from my side.

> Maybe @elfenpiff can shed some light why the to_be_removed_connections are currently only used when fire and forget is active.

The `to_be_removed_connections` are a buffer for pubsub and request response to receive messages from removed publishers, clients or servers. So the old connection is "cached" in the `to_be_removed_connections` until the last element was received and then the connection is removed.

When the client is in fire-and-forget mode, the message shall always reach the server (this is what fire-and-forget means in this context). My thinking was that when the client is not in fire-and-forget mode and therefore requires a response, it does not make sense for the server to receive a request from the `to_be_removed_connections` cache and create a response when the client is already out of scope. In this scenario, the request shall be just lost.

Now to your solution @xiao-yang25 . I think your approach is fine; now the server would also receive requests from clients that expect a response but are no longer available. But with `active_connection.is_connected() == false,` the server would be able to see that the client is no longer available and would not waste CPU cycles responding to something that no longer exists.

elBoberido (approved):

elBoberido:
@xiao-yang25 thanks for your contribution

xiao-yang25:
@elBoberido @elfenpiff thanks.I’m very happy to contribute to the project. I currently work on communication middleware at an embodied AI company. We previously used Cyclone DDS and iceoryx bundled with ROS 2 Humble.

However, iceoryx relies on RouDi and its centralized architecture. If a node fails to send heartbeats to RouDi for reasons unrelated to an actual node failure, RouDi may remove the node even though it is still accessing shared memory, which can cause the node to crash.

For this reason, I’m currently evaluating iceoryx2 for practical use on NVIDIA Jetson Orin platforms. I hope I’ll have more opportunities to contribute to iceoryx2 in the future.

elBoberido:
@xiao-yang25 cool. Are you allowed to share on which product iceoryx2 would run?

xiao-yang25:
@elBoberido It’s still in the evaluation and testing stage, and we haven’t decided yet which product it might be used in.= =

elBoberido:
@xiao-yang25 once you have made the decision and you are allowed to talk about, would be great to have a post in the [Showcase](https://community.iceoryx.io/c/showcase/6) section of our community forum.

xiao-yang25:
@elBoberido Sure! Once we’ve made a decision and I’m allowed to share it, I’d be happy to post there—assuming the startup is still around by then. You never know what might happen in startup land, haha.

Issue #1763, opened 2026-06-24
When dead Server/Client is cleaned up the connection remains connected


elfenpiff:
**Observed result or behaviour:**

When one side of `PendingResponse` <-> `ActiveRequest` connection crashes and is being cleaned up, the connection remains connected (`is_connected()` returns `true`).

**Expected result or behaviour:**
When a dead server/client is cleaned up, the connection is set to disconnected.

Issue #1680, opened 2026-05-25
`WaitSet::wait_and_process_with_timeout()` spins at 100% CPU after first event when 2+ listeners attached


gelanchez:
## Required information

**Operating system:**
Kubuntu 24.04 LTS

**Rust version:**
rustc 1.95.0 (59807616e 2026-04-14)

**Cargo version:**
cargo 1.95.0 (f2d3ce0bd 2026-03-21)

**iceoryx2 version:**
v0.9.0 (Python bindings via pip install iceoryx2==0.9.0)

**Observed result or behaviour:**
WaitSet.wait_and_process_with_timeout() returns immediately on every call after the first event is received on any attached listener, regardless of whether new events are pending. This causes 100% CPU usage on one core instead of blocking for the specified timeout.

**Expected result or behaviour:**
`wait_and_process_with_timeout(Duration.from_millis(100))` should block for up to 100 ms when no new events are pending and return approximately 10 times per second. After firing a single event and processing it, subsequent calls should block normally.

**Conditions where it occurred / Performed steps:**
1. Create a `WaitSet` with 2 or more listeners attached via attach_notification().
2. Fire exactly one event on any one of the attached listeners.
3. Call `wait_and_process_with_timeout()` in a loop, it never blocks again.

The bug occurs with any number of attachments ≥ 2. `timed_wait_one()` on a single listener is unaffected.

elfenpiff:
@gelanchez Thanks for the bug report. Currently, I am refactoring the event messaging pattern, and in this context, I will look into your bug report.

gelanchez:
Thanks @elfenpiff. I'll test the new code once it's ready. Meanwhile I have a workaround.

elrom-oai:
@elfenpiff re: the event refactor. Are there more details on this somewhere? Are there any plans to build an event mechanism such that e.g. request-response needn't poll?

Thanks!
-e

elfenpiff:
@gelanchez The event refactoring was merged and should now be much more efficient, and we can also guarantee that events are never lost. Could you please check out the latest main and verify that your issue is gone? If not, then I would start debugging it in the next days.

The API has changed slightly:

```rust
// old: no more `**_wait_one()`
while let Ok(Some(event_id)) = listener.timed_wait_one(CYCLE_TIME) {
    coutln!("event was triggered with id: {event_id:?}");
}

// old: renamed to `**_wait()`
listener.timed_wait_all(|event_id| {
[...]

PR #1857, opened 2026-07-28
[#1680] Add waitset test that blocks after a consumed event with multiple listeners

Nawazz-ish:
## Notes for Reviewer

@elBoberido this is the test you asked for in #1680.

The existing `wait_and_process_once_with_timeout_blocks_at_each_invocation` attaches a single listener and never fires an event, so it doesn't cover the reported scenario. This one attaches two listeners, fires one event, consumes it in the callback, and then checks that the following calls still block for the full timeout.

The bug itself is already gone, so this passes on current main. To make sure the test wasn't passing for the wrong reason, I removed the `try_wait` drain to recreate the old behaviour and confirmed all four service variants fail with:

```
assertion failed: expr: callback_called == false;  value: true == false
```

which is the reported symptom. Restored it afterwards and it's green again.

No changelog entry, since test-only changes don't appear to get one (e.g. `31d03b36e` for #1763). Happy to add one if you'd rather have it.

Checked with `cargo fmt` and `cargo test --workspace` on aarch64 macOS, 76 waitset tests pass.

Left this as `Relates to` rather than `Closes` since @gelanchez hasn't confirmed on Linux yet, so you may want to keep the issue open until then.

* [x] Branch follows the naming format (`iox2-1680-waitset-blocks-after-event-with-multiple-listeners`)
    * [x] Commit messages have the issue ID (`[#1680]`)
* [x] Tests follow best practice for testing guidelines
* [~] Changelog updated in the unreleased section including API breaking changes - not applicable, test-only change, see notes

## References

Relates to #1680

### DIFF
--- a/iceoryx2/conformance-tests/src/waitset.rs
+++ b/iceoryx2/conformance-tests/src/waitset.rs
@@ -308,6 +308,61 @@ pub mod waitset {
     }
 
     #[conformance_test]
+    pub fn wait_and_process_once_with_timeout_blocks_after_consumed_event_with_multiple_listeners<
+        S: Service,
+    >()
+    where
+        <S::Event as Event<RelocatableCountingBitSet>>::Listener: SynchronousMultiplexing,
+    {
+        let test = Test::<S>::new_with_custom_watchdog(Watchdog::new_with_timeout(TIMEOUT * 40));
+        let node = test.create_node();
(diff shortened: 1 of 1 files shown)


elBoberido (approved):

elBoberido:
@Nawazz-ish thanks for your contribution :)

Issue #1788, opened 2026-07-02
Node::list() panics on any non-u128 entry in the node-monitoring directory (unwrap on foreign/malformed names)


lakshay2010sharma:
## Summary

`Node::<Service>::list()` (and anything built on it, e.g. dead-node cleanup sweeps) panics the calling process if the node-monitoring directory contains **any** entry whose name does not parse as a bare `u128`:

```rust
// iceoryx2 0.9.1, src/node/mod.rs (Node::list, ~line 1128)
let node_id = core::str::from_utf8(node_name.as_bytes()).unwrap();
let node_id = UniqueNodeId(node_id.parse::<u128>().unwrap().into());
```

Both `.unwrap()`s are reachable from ordinary runtime conditions the listing process does not control, so a listing/monitoring/cleanup process can be killed by state it merely *observes*.

## Reproduction A — prefix-overlapping configs (how we hit it)

Two `Config`s share a `root_path`, and one `global.prefix` is a strict string-prefix of the other (e.g. `app_x` and `app_x_planning`). Namespace matching is `starts_with`-based, so the shorter-prefix process's `Node::list()` also picks up the longer namespace's monitor files; after stripping its own prefix, the residue (`_planning_<id>` etc.) is not a `u128` → `ParseIntError` panic during node creation/cleanup. The process dies at startup with no actionable error.

We now mint fixed-length hashed prefixes on our side so no two prefixes can be in a prefix relation — that works around it — but the failure mode is easy to hit for any user composing multiple namespaces under one `root_path`.

## Reproduction B — any stray file

Any non-conforming filename in the monitoring directory (a leftover editor temp file, a user-created file, a partially-written entry) panics every process that calls `Node::list()` for that config, including `iox2`-style tooling and dead-node cleanup paths.

## Expected behavior

Unparseable monitor-directory entries should be skipped (ideally with a warning through the failure/log machinery, like the existing `NodeState::new` error arm right below), not `unwrap()`ed. A malformed *foreign* entry shouldn't be able to kill an unrelated observer process.

## Environment

- iceoryx2 `0.9.1` (crates.io), Rust stable
- Observed on Linux (aarch64 and x86_64); the code path is platform-independent

Happy to provide more detail. Thanks for iceoryx2 — the zero-copy core has been excellent to build on.

dkroenke:
@lakshay2010sharma Thank you for creating the issue.
[...]

PR #1858, opened 2026-07-28
[#1788] Skip non-UTF-8 entries in Node::list() instead of panicking

Nawazz-ish:
## Notes for Reviewer

Small hardening change. `Node::list()` had one remaining `unwrap` at `node/mod.rs:1170`:

```rust
let node_id = core::str::from_utf8(node_name.as_bytes()).unwrap();
```

The `parse::<u128>()` immediately below already skips entries that aren't nodes, so this just makes the UTF-8 check behave the same way.

Worth being upfront: I could not get this to panic on either platform tested. On Linux the directory-entry layer rejects the name before it reaches this line (per @dkroenke in #1788), and on macOS APFS refuses to create a non-UTF-8 filename at all, failing with `Illegal byte sequence (os error 92)`. So this is defence-in-depth rather than a fix for a live crash. Happy to drop it if you'd rather not carry the change.

No test added, for the same reason: I can't construct the triggering condition on macOS, so a test would have to be Linux-only or mock the directory listing. Let me know if you want one and how you'd prefer it scoped.

Checked with `cargo fmt`, `cargo clippy` and `cargo test --workspace` on aarch64 macOS. 376 node tests pass. The one clippy warning is in `iceoryx2-bb-posix` and is untouched by this change.

* [x] Branch follows the naming format (`iox2-1788-harden-node-list-against-non-utf8-names`)
    * [x] Commit messages have the issue ID (`[#1788]`)
* [ ] Tests follow best practice for testing guidelines - no test added, see notes above

## References

Relates to #1788

### DIFF
--- a/iceoryx2/src/node/mod.rs
+++ b/iceoryx2/src/node/mod.rs
@@ -1167,10 +1167,12 @@ impl<Service: service::Service> Node<Service> {
         match Self::list_all_nodes(&monitoring_config) {
             Ok(node_list) => {
                 for node_name in node_list {
-                    let node_id = core::str::from_utf8(node_name.as_bytes()).unwrap();
+                    // not bad, just found a file that is not a node
+                    let Ok(node_id) = core::str::from_utf8(node_name.as_bytes()) else {
+                        continue;
+                    };
                     let node_id = match node_id.parse::<u128>() {
                         Ok(v) => UniqueNodeId(v.into()),
-                        // not bad, just found a file that is not a node
                         Err(_) => continue,
                     };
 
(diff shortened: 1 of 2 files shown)


iceoryx2/src/node/mod.rs
```diff
@@ -1167,7 +1167,10 @@ impl<Service: service::Service> Node<Service> {
         match Self::list_all_nodes(&monitoring_config) {
             Ok(node_list) => {
                 for node_name in node_list {
-                    let node_id = core::str::from_utf8(node_name.as_bytes()).unwrap();
+                    // not bad, just found a file that is not a node
```
dkroenke, line 1170: This comment is redundant to the one below, we could keep this one and remove the one below.

Nawazz-ish: Good catch, removed the lower one. Also rebased onto main while I was at it.

dkroenke (approved):
Thanks for your contribution!

Issue #1664, opened 2026-05-23
Improve console logger


elfenpiff:
## (Code) Example Of Cumbersome API

For complex multi-process scenarios it would be great to configure the console logger to show PID, executable name, and a shorter origin.

Nawazz-ish:
I'd like to pick this up, if nobody else is already on it.

I ran into the "shorter origin" part of this by accident while testing something else on main (`db6d89ed3`). A single warning about an unsupported config value printed roughly 4 KB, because the origin is the full `Debug` of the service builder - the whole node config, every default, file paths, mutex state:

```
[W] "Builder { base: BuilderWithServiceType { service_config: StaticConfig { iceoryx2_version: PackageVersion { ... }, service_hash: ... , shared_node: SharedNode { ... config: Config { global: Global { root_path: ... } ... } } ... }"
    (~4000 characters)
    | Setting the subscribers max borrowed samples to 0 is not supported. Adjust it to 1, the smallest supported value.
```

It comes from `let origin = format!("{self:?}");` in `iceoryx2/src/service/builder/publish_subscribe.rs:565`. The same pattern is used in 29 places, mostly under `service/builder/`, `service/port_factory/`, `service/static_config/mod.rs`, and a handful of spots in `iceoryx2-cal`.

Two questions on scope before I start:

1. Origin shortening - should it live in the logger (truncate/elide on output, which covers all 29 call sites at once), or at the call sites (replace the `format!("{self:?}")` bindings with something meaningful)? The logger side looks like the much smaller change and matches the issue title, so that's what I'd propose, with the call-site cleanup as a separate issue.
2. PID and executable name - configurable on the console logger, or always on? "configure the console logger" in the description suggests configurable, but I'd rather confirm than add knobs nobody wants.

Happy to go either way.

Nawazz-ish:
Followed up on my own question about where the origin shortening belongs, and I think truncating in the logger is the wrong answer.

Measured the actual string: the service builder origin is 6493 characters, and the identifying parts sit a long way in. `service_name` starts at char 280, the name itself at 352, `messaging_pattern` at 676. Truncated to 256 you get:

```
Builder { base: BuilderWithServiceType { service_config: StaticConfig { iceoryx2_version: PackageVersion { major: 0, minor: 9, patch: 999 }, service_hash: ServiceHash(RestrictedFileName { value: StaticString<64> { len: 40, data: \"edad20eea12e3b98 ...
```

[...]

Issue #1450, opened 2026-03-17
Make logging allocation free


elBoberido:
## Brief feature description

The `format!` macro is used for logging and this formats into a `String`, which allocates. Furthermore, #1434 introduced a new allocation before writing to stdout, in order to prevent tearing. 

## Detailed information

With Rust 1.93, https://doc.rust-lang.org/stable/core/fmt/fn.from_fn.html was introduced. This can be used to circumvent the allocations from the `format!` macro. Since we cannot immediately switch to Rust 1.93, we might be able to crate a similar abstraction in `iceoryx2-bb`.

For the tearing, an option would be to use the `write!` macro with a fixed size buffer, before passing it to stdout. This could either be a `StaticString` which would need to implement the `Write` trait or alternatively just a Rust array, which already implements this trait.

PR #1859, opened 2026-07-29
[#1664] Use originating function name as log origin instead of Debug dump

Nawazz-ish:
## Notes for Reviewer

Follow up to the discussion in #1664, covering the "shorter origin" part only. PID and executable name are not in here, for the reasons in the issue comment.

29 call sites built their log origin with `format!("{self:?}")`, which dumps the whole builder. For a service builder warning that comes to 6493 characters, and the identifying parts sit far enough in that truncating in the logger doesn't help. `service_name` starts at char 280, the name itself at 352.

This replaces them with the name of the originating function, which is what the other 133 origin sites in the tree already do (`let origin = "Config::global_config()";` and so on). Before:

```
0 [W] "Builder { base: BuilderWithServiceType { service_config: StaticConfig { iceoryx2_version: PackageVersion { ... 6493 chars ... }"
| Setting the subscribers max borrowed samples to 0 is not supported. Adjust it to 1, the smallest supported value.
```

After:

```
0 [W] "publish_subscribe::Builder::adjust_configuration_to_meaningful_values()"
| Setting the subscribers max borrowed samples to 0 is not supported. Adjust it to 1, the smallest supported value.
```

Two things worth mentioning:

Most of these sites used `format!("{self:?}")` because `self` gets moved or mutably borrowed further down, so a `Debug` snapshot was the easy way to keep something to log. A `&'static str` sidesteps that, and drops an allocation from each of these paths, which is a small step towards #1450.

Removing the `format!` calls left 17 `use alloc::format;` imports unused, so those are gone too. I checked each file has no remaining `format!` usage rather than relying on the warning alone, in case any were feature gated.

No behaviour change beyond the log text. `cargo fmt`, `cargo clippy --workspace --all-targets` and `cargo test --workspace` are clean on aarch64 macOS. The one clippy warning in `iceoryx2-bb-posix` is present on main too.

Naming is the part I would most like input on. I used `Type::method()`, module qualified where the type name alone would be ambiguous, e.g. `posix_shared_memory::Connector::open_sender()`. Happy to change the scheme, it is a single line per site.

* [x] Branch follows the naming format (`iox2-1664-replace-debug-dump-origins`)
    * [x] Commit messages have the issue ID (`[#1664]`)
* [ ] Tests follow best practice for testing guidelines - no new tests, log output only, see notes

## References

Relates to #1664

### DIFF
--- a/iceoryx2-bb/linux/src/epoll.rs
+++ b/iceoryx2-bb/linux/src/epoll.rs
@@ -322,7 +322,7 @@ impl EpollBuilder {
             });
         }
 
-        let origin = format!("{self:?}");
+        let origin = "EpollBuilder::create()";
         let signal_fd = match SignalFdBuilder::new(self.signal_set)
             .set_close_on_exec(self.has_close_on_exec_flag)
             .create_non_blocking()
--- a/doc/release-notes/iceoryx2-unreleased.md
+++ b/doc/release-notes/iceoryx2-unreleased.md
@@ -60,6 +60,7 @@
 
 * [#996](https://github.com/eclipse-iceoryx/iceoryx2/issues/996) Move BumpAllocator from iceoryx2-bb-memory into iceoryx2-bb-elementary
 * [#1613](https://github.com/eclipse-iceoryx/iceoryx2/issues/1613) Remove `NonNullCompat` after moving to Rust 1.89
+* [#1664](https://github.com/eclipse-iceoryx/iceoryx2/issues/1664) Use the originating function name as log origin instead of a `Debug` dump of the whole builder
 * [#1776](https://github.com/eclipse-iceoryx/iceoryx2/issues/1776) Rename AtomicCopy::__for_each_field() to for_each_field()
 * [#1845](https://github.com/eclipse-iceoryx/iceoryx2/issues/1845) Reduce imports for usage of the `semantic_string` macro
 * [#1853](https://github.com/eclipse-iceoryx/iceoryx2/issues/1853) Improve error message in static asserts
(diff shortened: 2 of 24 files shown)


iceoryx2-bb/linux/src/epoll.rs
```diff
@@ -322,7 +322,7 @@ impl EpollBuilder {
             });
         }
 
-        let origin = format!("{self:?}");
```
elBoberido, line 325: This is also something we could discuss at the developer meetup next week. The idea was less about having the function name, but the internal state of the object when the log incident happened. We have ideas on how to make this more efficient and also limit the size of the log message but we also want to retain the possibility to get enough information for debugging.

Nawazz-ish: Ah, that makes sense. I read the `Debug` dump as accidental verbosity rather than something you deliberately want. Keeping the object state while bounding the cost is a different problem from the one this solves, and as it stands the PR throws away the part you actually care about.

Converting it to a draft so it isn't sitting in the review queue. Happy to rework it once you've settled on the direction.

elBoberido: If you have time to join the developer meetup next Tuesday, we could discuss this topic.

Nawazz-ish: Yes, I will be there Tuesday. Happy to go through it then.

Nawazz-ish:
Closing this one. #1877 covers the part of this that was worth keeping - dropping the eager
`format!` on the service creation path - and does it without replacing the origins wholesale,
which was the part you objected to here.

The measurement that came out of this is that the origin is not really too long so much as the
wrong shape: ~79% of the 6493 characters is the `shared_node` subtree, which is the node
`Config` and is identical for every service on that node. I have left a note on #1664 with the
numbers and a prototype, so the size question can continue there rather than on a stale PR.

Thanks for the steer, it was the right call.

Issue #1152, opened 2025-10-27
Use `unistd` for console logger


orecham:
## Brief Description

Use `unistd` from the POSIX abstraction instead of the standard library to enable the logger to be used on `no_std` platforms.

Nawazz-ish:
Had a look at this one. I think the `unistd` part is already done: `iceoryx2-pal-print/src/writer.rs` implements `Write for Stderr` with `posix::write(STDERR_FILENO, ...)` under `not(feature = "std")`, so the console logger's output already goes through the POSIX abstraction rather than `std::io`.

The goal in the description isn't quite met though. The console logger doesn't compile on `no_std` for anything other than linux and nto:

```
$ cargo build -p iceoryx2-bb-loggers --no-default-features --features console
error[E0433]: failed to resolve: use of undeclared type `Duration`
  --> iceoryx2-bb/loggers/src/console.rs:73:9
```

`duration_since_epoch()` has `use core::time::Duration;` inside the linux/nto branch, but the fallback branch below it uses `Duration` with nothing in scope. Adding the same import to that branch is enough to fix it, and `std` builds are unaffected.

Worth noting the no_std CI job runs on ubuntu-latest, so it always takes the linux branch and never sees this. That's probably why it went unnoticed.

For what it's worth `build.rs` rejects `buffer`, `file`, `log` and `tracing` on `no_std` but allows `console`, so console looks like the one that is meant to work there.

Happy to send the fix. Should it go under this issue, or would you rather have a separate one for it?

elBoberido:
@Nawazz-ish yes, please create a PR. It can be done with this issue. I will update the title to reflect the changes better.

elBoberido:
@orecham I think this issue should not be closed. Would you agree to limit the scope to no_std POSIX platforms? Bare metal would be excluded and also everything that does not have a POSIX layer.

PR #1861, opened 2026-07-29
[#1152] Fix no_std build of the console logger on non-linux platforms

Nawazz-ish:
## Notes for Reviewer

As discussed in #1152. The `unistd` part already looked done in `iceoryx2-pal-print/src/writer.rs`, so this only fixes the remaining `no_std` build break.

`duration_since_epoch()` in `console.rs` keeps `use core::time::Duration;` inside the linux/nto branch, but the fallback branch below it also uses `Duration`. So on any other target:

```
$ cargo build -p iceoryx2-bb-loggers --no-default-features --features console
error[E0433]: failed to resolve: use of undeclared type `Duration`
  --> iceoryx2-bb/loggers/src/console.rs:73:9
```

This adds the same import to that branch. Verified it fails on main and passes with the change, on aarch64 macOS. `std` builds are untouched, `cargo fmt` clean.

One thing you may want to pick up separately: the `no_std` CI job runs on ubuntu-latest, so it always takes the linux branch and can't see breakage in the fallback path. Adding a `no_std` build on a non-linux target, or a bare metal one, would catch this class of problem. Happy to open a separate issue for that if it's useful.

* [x] Branch follows the naming format (`iox2-1152-fix-no-std-console-logger-build`)
    * [x] Commit messages have the issue ID (`[#1152]`)
* [ ] Tests follow best practice for testing guidelines - build fix, no test added, see notes

## References

Relates to #1152

### DIFF
--- a/doc/release-notes/iceoryx2-unreleased.md
+++ b/doc/release-notes/iceoryx2-unreleased.md
@@ -35,6 +35,7 @@
 
 * [#156](https://github.com/eclipse-iceoryx/iceoryx2/issues/156) Remove `fchmod`/`shm_open` macOS workarounds; route permissions through a trampoline state file.
 * [#588](https://github.com/eclipse-iceoryx/iceoryx2/issues/588) Replace deprecated `serde_yaml` dependency with `yaml_serde`.
+* [#1152](https://github.com/eclipse-iceoryx/iceoryx2/issues/1152) Fix `no_std` build of the console logger on platforms other than linux and nto.
 * [#1548](https://github.com/eclipse-iceoryx/iceoryx2/issues/1548) Fix Payload data lifetime tracking in python ffi by anchoring views to their owning Sample.
 * [#1673](https://github.com/eclipse-iceoryx/iceoryx2/issues/1673) Thread-stack-size is the same as process-stack-size on all platforms.
 * [#1695](https://github.com/eclipse-iceoryx/iceoryx2/issues/1695) Remove port_tag when stale resources of port are removed.
--- a/iceoryx2-bb/loggers/src/console.rs
+++ b/iceoryx2-bb/loggers/src/console.rs
@@ -70,6 +70,8 @@ fn duration_since_epoch() -> core::time::Duration {
         not(any(target_os = "linux", target_os = "nto",))
     ))]
     {
+        use core::time::Duration;
+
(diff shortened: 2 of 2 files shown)


elBoberido (approved):
Nice. Does the output also work?

Nawazz-ish:
Good question, and no, not on targets other than linux and nto.

`Write for Stderr` in `iceoryx2-pal-print` only has a real implementation under `not(feature = "std")` for linux and nto, using `posix::write` with `STDERR_FILENO`. Every other target gets:

```rust
fn write_str(&mut self, _s: &str) -> fmt::Result {
    Ok(())
}
```

so the message is silently dropped and `is_terminal()` returns `false`. This PR only gets it compiling again.

I did check it builds for a real bare metal target rather than just macOS:

```
cargo build -p iceoryx2-bb-loggers --no-default-features --features console --target aarch64-unknown-none
```

2 errors before the change, 0 after. `thumbv7em-none-eabihf` still fails, but for an unrelated reason: `iceoryx2-pal-concurrency-sync` needs 64 bit atomics that the target doesn't have.

For output to actually work on bare metal something has to implement that writer, and it isn't obvious what stderr should mean there - semihosting, a UART, defmt, or something the user injects. Happy to open a separate issue if you want it tracked.

orecham:
The printing is stubbed on purpose on bare metal targets because it is target-specific.

Nawazz-ish:
Understood, that makes sense. I will leave it alone then.

elfenpiff:
@Nawazz-ish thank you for your contribution!

elBoberido:
@orecham would it make sense to have the no_std behavior on Linux and nto for every platform and let the platform abstraction drop the message if only a stub is implemented?

orecham:
> @orecham would it make sense to have the no_std behavior on Linux and nto for every platform and let the platform abstraction drop the message if only a stub is implemented?

@elBoberido This topic will need some discussions. The question is if we want to force users deploying to bare metal / rtos to implement the entire platform abstraction to get iceoryx2 running. I lean towards no, but enabling an easier pathway is not trivial.

elBoberido:
@orecham I also lean towards no. So let me phrase it differently. If we had a way to determine if the platform has POSIX support, then the no_std behavior should be the same like on Linux and nto with no_std. Would you agree?

PR #1874, opened 2026-08-02
[#1868] auto create platform config files

### LINKED ISSUE #1868: Windows: hardcoded `C:\Temp` platform paths break iceoryx2 on stock installs
## Required information

**Operating system:** Windows 11 Pro (10.0.26200)
**Rust version:** rustc 1.97.1
**iceoryx2 version:** 0.9.3 (also present on `main`)

## Summary

`iceoryx2-pal-configuration` hardcodes the Windows platform paths:

```rust
pub const TEMP_DIRECTORY: &[u8] = b"C:\\Temp\\";
pub const SHARED_MEMORY_DIRECTORY: &[u8] = b"C:\\Temp\\iceoryx2\\shm\\";
pub const ICEORYX2_ROOT_PATH: &[u8] = b"C:\\Temp\\iceoryx2\\";
```

Node creation fails out of the box without a clear error on Windows because `C:\Temp` does not exist on a stock install and iceoryx2 does not create it.

## Observed result

With the default config:

```
| Unable to create file since the path points to a directory that does not exist.
| Failed to create dynamic_storage::file::DynamicStorage since the underlying file could not be created.
| Unable to create node since the global management segment could not be opened. [DynamicStorageCreateError(InternalError)]
```

(visible only at debug log level; the API returns `NodeCreationFailure::InternalError`).

With a custom `Config` whose `root_path` exists, the shm state files still resolve to `TEMP_DIRECTORY`; the path-not-found from `CreateFileA(CREATE_NEW)` is misread as a stale previous instance and the process dies (exit 253):

```
| Old dynamic config from previous instance discovered - trying to remove it.
< Win32 API error > ...\src\windows\mman.rs:368 DeleteFileA(shm_file_path(name, SHM_STATE_SUFFIX).as_ptr())
 [ 3 ] The system cannot find the path specified.
```

## Expected result

iceoryx2 works on a fresh Windows install.

## Suggestion

The best solution for my use case would be to resolve the location at runtime through an API/config. This lets me create a fully self-contained iceoryx2 embedding that can't clobber another. The lower hanging fruit would be to use a standard Windows temp directory.

elfenpiff:

## PR Reviewer Reminders

* Public API is documented

## References

Relates to #1868

### DIFF
--- a/iceoryx2-pal/posix/src/windows/mman.rs
+++ b/iceoryx2-pal/posix/src/windows/mman.rs
@@ -15,7 +15,8 @@
 #![allow(unused_variables)]
 
 use crate::posix::{
-    Errno, MemZeroedStruct, constants::*, settings::*, to_dir_search_string, types::*,
+    Errno, MemZeroedStruct, constants::*, create_system_directories, settings::*,
+    to_dir_search_string, types::*,
 };
 use crate::win32call;
 
@@ -201,6 +202,7 @@ pub unsafe fn shm_list() -> Vec<[i8; 256]> {
 }
 
 pub unsafe fn shm_open(name: *const c_char, oflag: int, mode: mode_t) -> int {
+    create_system_directories();
     unsafe {
         let name = remove_leading_path_separator(name.cast());
         let handle: HANDLE = 0;
--- a/iceoryx2-pal/posix/src/windows/mod.rs
+++ b/iceoryx2-pal/posix/src/windows/mod.rs
@@ -39,6 +39,14 @@ pub mod win32_handle_translator;
 pub mod win32_security_attributes;
 mod win32_udp_port_to_uds_name;
 
+use iceoryx2_pal_concurrency_sync::atomic::AtomicU8;
+use iceoryx2_pal_concurrency_sync::atomic::Ordering;
+use iceoryx2_pal_configuration::GLOBAL_CONFIG_PATH;
+use iceoryx2_pal_configuration::ICEORYX2_ROOT_PATH;
+use iceoryx2_pal_configuration::SHARED_MEMORY_DIRECTORY;
+use iceoryx2_pal_configuration::TEMP_DIRECTORY;
+use iceoryx2_pal_configuration::TEST_DIRECTORY;
+
 pub use crate::os::posix::constants::*;
 pub use crate::os::posix::dirent::*;
 pub use crate::os::posix::errno::*;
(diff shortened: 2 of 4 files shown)


iceoryx2-pal/posix/src/windows/mod.rs
```diff
@@ -60,3 +68,45 @@ pub use crate::os::posix::support::*;
...
+    }
+}
+
+pub fn create_system_directories() {
+    match SYSTEM_DIRECTORIES_CREATION_STATE.compare_exchange(
+        MISSING,
+        IN_CREATION,
+        Ordering::SeqCst,
+        Ordering::SeqCst,
+    ) {
```
elBoberido, line 96: This does not fully work. One could have two nodes with disjoint domains and then the issue would still persist. Why not do this on node creation?

iceoryx2/conformance-tests/src/service.rs
```diff
@@ -1455,4 +1455,27 @@ pub mod service {
         let service_2 = test.open(&node_2, &service_name, &AttributeVerifier::new());
         assert_that!(service_2, is_err);
     }
+
+    #[conformance_test]
+    pub fn creating_service_with_different_root_path_works<
```
dkroenke, line 1460: There should be a test that checks if the files are generated within the correct directory.

iceoryx2-pal/posix/src/windows/mman.rs
```diff
@@ -201,6 +202,7 @@ pub unsafe fn shm_list() -> Vec<[i8; 256]> {
 }
 
 pub unsafe fn shm_open(name: *const c_char, oflag: int, mode: mode_t) -> int {
+    create_system_directories();
```
dkroenke, line 205: Do we want to have this behavior for Windows only?

elfenpiff: Yes, because it is a Windows-only problem. The other platforms do not require creating a `C:\Temp\iceoryx2\shm` directory to store shared memory files since they can use `/dev/shm,` for instance - and this directory is handled by the OS.

iceoryx2/conformance-tests/src/service.rs
elfenpiff: Yes, those tests are already inside the concept abstraction layer. The iceoryx2 layer does not know anything about files and depending on the platform, there are different files inside the root-path - depending on the concept implementation details.

iceoryx2-pal/posix/src/windows/mod.rs
elfenpiff: Because this has nothing to do with node creation. When a shared memory is created on the windows platform with `shm_open`, then we need to ensure that the system directory to host the shared memory files is available.

So this problem affects all layers of iceoryx2 (iceoryx2-bb-posix, iceoryx2-cal, iceoryx2) and the only way to fix this, is to handle it in the platform abstraction layer.

elfenpiff: @elBoberido 

> This does not fully work. One could have two nodes with disjoint domains and then the issue would still persist.

I do not understand. Where the bug originates, there are no domains. The issue is that the system directory where the shared memory files are stored (under Windows) is not present when the file is being created.

Could you please elabore this.

elBoberido: Maybe I misunderstood something, but I assumed this has to do with the root path and one could for example do this

```rust
    let mut config = Config::global_config().clone();
    config
        .global
        .set_root_path(&"C:\\fuu".try_into()?);
    let node1 = NodeBuilder::new()
        .config(&config)
        .create::<ipc::Service>()?;

    config
        .global
        .set_root_path(&"C:\\bar".try_into()?);
    let node2 = NodeBuilder::new()
        .config(&config)
        .create::<ipc::Service>()?;
```
With the current implementation, wouldn't this result in creating the paths only for "C:\\fuu" and not for "C:\\bar"?

The use case for such a scenario could be a bridge for two disjoint domains.

elfenpiff: No, it has nothing to do with the root path. By accident, the root path had a partially similar location, and therefore, in certain scenarios, it was created by coincidence in the same directory as the system-shared memory files required.
(low level platform shm dir: `C:\Temp\iceoryx2\shm`, high level root root path `C:\Temp\iceoryx2\`)

I added the tests on high-level iocorex2 to ensure that this is fixed, but it needs to be on the lowest layer.

elBoberido (approved):

PR #1877, opened 2026-08-03
[#1664] Make service builder log origins lazy

Nawazz-ish:
## Notes for Reviewer

Follow up to #1664 and the discussion on #1859. This is only the "stop doing pointless work"
half - it does not touch the size question, which is what I understood you have ideas about
and what we said we would discuss at the meetup.

The logging path is already lazy. `__internal_print_log_msg` checks the level before touching
the logger and takes `core::fmt::Arguments`, and `warn!(from x, ...)` passes `format_args!`,
so nothing is formatted when a message is filtered out. But these call sites build a `String`
at function entry, before any of the `if`s:

```rust
fn adjust_configuration_to_meaningful_values(&mut self) {
    let origin = format!("{self:?}");   // always runs
    ...
    if settings.subscriber_max_borrowed_samples == 0 {
        warn!(from origin, "...");      // may never fire
    }
```

For a service builder that `Debug` is 6493 characters, and it is paid on the success path with
logging disabled. `warn!(from self, ...)` is already what 224 other call sites do, so this just
makes two of the exceptions match.

Two sites only, both on the service creation path. In `BuilderWithServiceType::open` the
receiver is `&self` so it is a straight swap. In `adjust_configuration_to_meaningful_values`
the `&mut` borrow through `settings` spans all five `warn!` calls, so each check now reads
immutably via the existing `publish_subscribe()` accessor, logs, then mutates.

No behaviour change other than the log text, which actually gets slightly cleaner: the origin
was being formatted twice (call site builds a `String`, the macro then applies `{:?}` to it),
which added 58 backslash escapes. That goes to 0.

Measured, 2000 publish-subscribe create-and-destroy cycles, log level `Fatal` so nothing is
ever emitted, 7 runs, median:

| | before | after | saving |
| --- | --- | --- | --- |
| aarch64 macOS | 825.6 us | 759.9 us | 66 us |
| x86_64 Linux (m5.metal) | 5.887 ms | 5.748 ms | 139 us |

Worth being explicit that this is **not** a throughput or latency win - none of these sites is
on the send or receive path. It is startup and service creation work, and the percentage it
represents varies a lot with the storage under the temp directory (8.0% on the Mac's local SSD,
2.4% on the Linux box's EBS volume), so the absolute number is the meaningful one.

I left the other 27 sites alone deliberately. 12 more could convert the same way but showed no
measurable benefit, and the remaining 15 consume `self`, so `from self` does not compile there
at all. Those are better handled by whatever comes out of the size discussion.

Checked with `cargo fmt`, `cargo clippy` and `cargo test --workspace` on aarch64 macOS - 534
publish-subscribe tests pass. The one clippy warning in `iceoryx2-bb-posix` is on main too.

* [x] Branch follows the naming format (`iox2-1664-make-log-origins-lazy`)
    * [x] Commit messages have the issue ID (`[#1664]`)
* [ ] Tests follow best practice for testing guidelines - no new tests, existing coverage
      exercises these paths and the change has no behavioural effect

## References

Relates to #1664


iceoryx2/src/service/builder/publish_subscribe.rs
```diff
@@ -575,37 +573,58 @@ impl<
...
-        if settings.subscriber_max_borrowed_samples == 0 {
-            warn!(from origin,
+        if self
+            .base
+            .service_config
+            .publish_subscribe()
+            .subscriber_max_borrowed_samples
+            == 0
+        {
+            warn!(from self,
```
elBoberido: This makes the code more difficult to read. I guess in this case it would be fine to just have the name of the function as origin, e.g.

```rust
let origin = "publish_subscribe::Builder::adjust_configuration_to_meaningful_values";
```

Nawazz-ish: Agreed, and done. The restructuring was the price of keeping `self` borrowable at the `warn!`,
and it was not worth paying here - this function only ever warns about config values it is
about to correct, so the builder state adds nothing.

Using your suggestion verbatim. The eager `format!` is gone either way, so the saving is
unchanged: ~77 us per service create-and-destroy cycle against current main (medians of 9 runs
each, 880.7 us vs 804.1 us, aarch64 macOS).

`BuilderWithServiceType::open` still uses `from self`, since there the receiver is `&self` and
it was a one-line swap with no restructuring.

That leaves the diff at 3 files, +10/-13.

elBoberido (approved):

PR #1889, opened 2026-08-05
[#1458] Fix deadlock in SignalHandler::call_and_fetch

### LINKED ISSUE #1458: `signal_test::signal_call_and_fetch_with_registered_handler_works()` deadlocks in single-threaded execution
## Brief Description

Detected when migrating the test suite to use the custom test framework.

The custom test runner runs each test on the `main` thread. Conversely, `libtest` runs each test on a worker thread, leaving the main thread free.  
See: https://github.com/rust-lang/rust/blob/main/library/test/src/lib.rs#L634

Leads:

* `SignalHandler::call_and_fetch` calls `SignalHandler::instance()`: https://github.com/eclipse-iceoryx/iceoryx2/blob/main/iceoryx2-bb/posix/src/signal.rs#L403
* `handler` calls `SignalHandler::instance()`: https://github.com/eclipse-iceoryx/iceoryx2/blob/main/iceoryx2-bb/posix/src/signal.rs#L329
* `SignalHandler::instance()` calls `MTX.lock()`: https://github.com/eclipse-iceoryx/iceoryx2/blob/main/iceoryx2-bb/posix/src/signal.rs#L569

zacheryasc:
Fixes the single-thread deadlock tracked in #1458.

**Root cause.** `SignalHandler::call_and_fetch` held the global `SignalHandler` mutex across the user-provided callable. When the callable raised a signal that had a registered callback, the C signal handler ran synchronously on the same (only) thread and tried to re-acquire that same non-reentrant mutex to dispatch the callback, deadlocking. The custom test runner runs each test on the main thread (unlike `libtest`'s worker thread), which is what exposed it.

**Fix.** `call()` now runs outside the locked scope; the `LAST_SIGNAL` reset stays under the lock. It also removes a latent deadlock for the production caller in `shared_memory.rs` (`call_and_fetch(memset_call)`).

**Test.** Un-ignores `call_and_fetch_with_registered_handler_works`, now passes in ~0.1s (previously hung until the watchdog aborted). 

## Notes for Reviewer

TL;DR Fixes a signal handler mutex deadlock.

## PR Reviewer Reminders

* Public API is documented

## References

Closes #1458

### DIFF
--- a/iceoryx2-bb/posix/src/signal.rs
+++ b/iceoryx2-bb/posix/src/signal.rs
@@ -402,9 +402,11 @@ impl SignalHandler {
         {
             let _sighandle = Self::instance();
             LAST_SIGNAL.store(posix::MAX_SIGNAL_VALUE, Ordering::Relaxed);
-
-            call();
         }
+
+        // `call()` must run outside the lock: it may raise a signal whose
+        // handler re-locks this mutex (iox2 #1458).
+        call();
 
         match LAST_SIGNAL.load(Ordering::Relaxed) {
             posix::MAX_SIGNAL_VALUE => None,
--- a/doc/release-notes/iceoryx2-unreleased.md
+++ b/doc/release-notes/iceoryx2-unreleased.md
@@ -36,6 +36,7 @@
 * [#156](https://github.com/eclipse-iceoryx/iceoryx2/issues/156) Remove `fchmod`/`shm_open` macOS workarounds; route permissions through a trampoline state file.
 * [#588](https://github.com/eclipse-iceoryx/iceoryx2/issues/588) Replace deprecated `serde_yaml` dependency with `yaml_serde`.
 * [#1152](https://github.com/eclipse-iceoryx/iceoryx2/issues/1152) Fix `no_std` build of the console logger on platforms other than linux and nto.
+* [#1458](https://github.com/eclipse-iceoryx/iceoryx2/issues/1458) Fix deadlock in `SignalHandler::call_and_fetch` when the callable raises a signal with a registered callback; the global signal-handler mutex is no longer held across the user-provided callable.
 * [#1548](https://github.com/eclipse-iceoryx/iceoryx2/issues/1548) Fix Payload data lifetime tracking in python ffi by anchoring views to their owning Sample.
 * [#1673](https://github.com/eclipse-iceoryx/iceoryx2/issues/1673) Thread-stack-size is the same as process-stack-size on all platforms.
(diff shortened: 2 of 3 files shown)


iceoryx2-bb/posix/src/signal.rs
```diff
@@ -400,10 +402,12 @@ impl SignalHandler {
         {
             let _sighandle = Self::instance();
             LAST_SIGNAL.store(posix::MAX_SIGNAL_VALUE, Ordering::Relaxed);
-
-            call();
         }
 
+        // `call()` must run outside the lock: it may raise a signal whose
+        // handler re-locks this mutex (iox2 #1458).
+        call();
```
elBoberido, line 409: @zacheryasc unfortunately, this is not as simple as it looks like. The whole abstraction needs to be refactored in order to make this work as intended. The current implementation is a bit racy and it is not guaranteed that the `LAST_SIGNAL.load(...)` call from the next line will have the value set by a signal that was triggered by code from `call`.

In addition, there are two other test which are disabled due to #1458 and still deadlocks even with this change.

zacheryasc: I believe I see where I misunderstood. The issue is, we are waiting for response after our call on thread A, and meanwhile thread B runs its own POSIX call, writes it to the global variable, and thread A reads a totally unrelated response, or possibly no response.

One solution is handling the signal synchronously with `pthread_sigmask` and `sigtimedwait`.

This is if the signals we are handling here are not meant to be process-wide, i.e. `kill` or somesuch, which is what is stated in the [`call_and_fetch` doc comment](https://github.com/eclipse-iceoryx/iceoryx2/blob/9b872e54a/iceoryx2-bb/posix/src/signal.rs#L397-L400) and its only production caller, the [`shared_memory` memset path](https://github.com/eclipse-iceoryx/iceoryx2/blob/9b872e54a/iceoryx2-bb/posix/src/shared_memory.rs#L384-L399), but we might want a guard of sorts to make sure. Perhaps a test.

zacheryasc: If this direction seems reasonable, or at least not obviously the wrong approach, to you, then give the word and I will draft an implementation.

zacheryasc:
After thinking and exploring the code more deeply, I have realized this PR is the wrong scope. It narrowly targets one failure case, and insufficiently at that, of a broader class of issues. I have opened https://github.com/eclipse-iceoryx/iceoryx2/issues/1898 to track the general problem that this is a specific case of.

PR #1900, opened 2026-08-07
[#1898] Add test coverage for signal-isolation behavior

### LINKED ISSUE #1898: Signal handling lacks ownership and isolation across consumers, causing races and contention
## Brief Description

iceoryx2 provides several means of signal consumption but no model of ownership.

The signal subsystem is built around a single global singleton. `SignalHandler`
holds the disposition for every signal behind one mutex and exposes one
process-global latch (`LAST_SIGNAL`). The delivery handler acquires that mutex
to dispatch user callbacks, and at construction the singleton claims all
non-fatal signals.

The means of consumption are:

- **Capture per call** — `call_and_fetch` runs a callable and returns the
  signal it raised.
- **Latch** — `termination_requested` / `last_signal` report the most recent
  signal.
- **File-descriptor reception** — `SignalFd` / epoll deliver signals as
  pollable events.
- **Callback** — `register` runs a user callback when a signal arrives.

POSIX permits only one disposition per signal, process-wide. Routing all of the
above through one shared disposition and one shared latch gives no consumer an
isolated view of the signal it asked for; each assumes sole ownership and they
break one another. This produces four classes of defect:

- **Contention** — consumers collide over the shared disposition. The handler's
  mutex re-enters itself when a captured signal also has a registered callback,
  deadlocking (`call_and_fetch_with_registered_handler_works`, #1458).
- **Races** — a consumer observes a signal it did not cause. `call_and_fetch`
  reads the process-global latch, so under concurrent signal activity it can
  return another thread's signal, or miss its own.
- **Starvation** — a consumer never receives its signal because another
  consumes it first. `SignalFd` / epoll never receive: the global handler
  swallows the signal before it reaches the fd, and nothing blocks the signal
  for the signalfd.
- **Async-signal-unsafety** — the handler does work that is not
  async-signal-safe: it acquires a mutex and runs arbitrary user callbacks.

## Direction

Each consumer needs an isolated relationship with the signals it consumes.
Three properties are required: a signal's disposition has an explicit owner;
capture is scoped to the operation that requested it; and no lock is held on
the signal delivery path.

## Acceptance

- [ ] `call_and_fetch_with_registered_handler_works` no longer deadlocks *(#1458)*
- [ ] `call_and_fetch` returns the signal from *its own* call, not another
      thread's
- [ ] `SignalFd` / epoll receive signals sent to the process *(un-ignore the two
      `#[ignore] // TODO: #1458` tests)*

## Related

- #1458 — the deadlock; an instance of the contention mode and the anchor for
  this class.
- #1556 — propagating `InterruptedBySignal` as a distinct error; complementary,
  operating on the layer above signal delivery.

zacheryasc:
The purpose is to complete test coverage that asserts the behavior #1898 prescribes, so a future fix PR can be validated against it. This PR does **not** fix #1898. 

Coverage is spread across three test files:
* [x] `iceoryx2-bb/linux/tests-common/src/epoll_tests.rs` — a signal subscribed via `EpollBuilder::handle_signal` is delivered as `EpollEvent::Signal` despite the global handler claiming the disposition
* [x] `iceoryx2-bb/linux/tests-common/src/signal_fd_tests.rs` — signalfd delivery
* [x] `iceoryx2-bb/posix/tests-common/src/signal_tests.rs` — `call_and_fetch` per-thread isolation, and the #1458 contention case

## Notes for Reviewer

* This is coverage prep for #1898, not the fix. The ignored tests fail today by
  design; a fix PR un-ignores them as proof.
* With the ignore removed, the epoll test was confirmed to fail deterministically
  in ~0.2 s (no hang, no watchdog trip).

## PR Reviewer Reminders

* Public API is documented

## References

Relates to #1898 #1458

### DIFF
--- a/iceoryx2-bb/linux/tests-common/src/epoll_tests.rs
+++ b/iceoryx2-bb/linux/tests-common/src/epoll_tests.rs
@@ -312,55 +312,41 @@ pub fn blocking_wait_wakes_up_by_trigger() {
     }).unwrap();
 }
 
-// TODO: #1458
+// The global `SignalHandler` claims every non-fatal signal's disposition; an epoll
+// subscribed to that same signal must still receive it as `EpollEvent::Signal`.
+// TODO: #1898
+// currently failing as the epoll's signalfd is starved by the competing `call_and_fetch(..)`
 #[ignore]
 #[test]
 pub fn signals_can_be_received() {
-    let _watchdog = Watchdog::new_with_timeout(core::time::Duration::from_secs(1000));
-    let (socket_1, _socket_2) = StreamingSocket::create_pair().unwrap();
+    let _watchdog = Watchdog::new();
     let sut = EpollBuilder::new()
         .handle_signal(FetchableSignal::UserDefined1)
         .create()
         .unwrap();
-    let _guard = sut
-        .add(socket_1.file_descriptor())
-        .event_type(EventType::ReadyToRead)
-        .attach()
-        .unwrap();
 
-    let callback_was_called = AtomicBool::new(false);
-    let handle = BarrierHandle::new();
-    let barrier = BarrierBuilder::new(2).create(&handle).unwrap();
-    thread_scope(|s| {
-        s.thread_builder().spawn(|| {
-            barrier.wait();
-            assert_that!(sut.blocking_wait(|event| {
-                if let EpollEvent::Signal(signal) = event {
-                    assert_that!(signal.signal(), eq FetchableSignal::UserDefined1);
-                    assert_that!(signal.origin_uid(), eq User::from_self().unwrap().uid());
-                    assert_that!(signal.origin_pid(), eq Process::from_self().id());
-                } else {
-                    assert_that!(true, eq false);
-                }
-                callback_was_called.store(true, Ordering::Relaxed);
-            }).unwrap(), eq 1);
-        })?;
+    // Activate the global handler; its default disposition is what starves the fd.
+    let _ = SignalHandler::call_and_fetch(|| {});
 
-        barrier.wait();
-        nanosleep(TIMEOUT).unwrap();
-        assert_that!(callback_was_called.load(Ordering::Relaxed), eq false);
+    SignalHandler::call_and_fetch(|| {
+        Process::from_self()
+            .send_signal(FetchableSignal::UserDefined1.into())
+            .unwrap();
+    });
 
-        while !callback_was_called.load(Ordering::Relaxed) {
-            SignalHandler::call_and_fetch(|| {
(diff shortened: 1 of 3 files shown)


zacheryasc:
Reasoning for the change in the epoll test behavior:
- The previous watchdog-based guard would hang for over 10 minutes, and possibly flake if a signal slipped through somewhere, without actual guarantees on proper delivery.
- The rewrite deterministically succeeds or fails with a send-receive check that asserts proper `epoll` delivery.

zacheryasc:
Reasoning for the change in the signalfd test behavior:
- The previous delivery test used an unbounded retry loop that would spin forever when the fd was starved, never actually failing. The rewrite deterministically succeeds or fails within a bounded ~200 ms poll.
- Two regression guards are added to pin the isolation contract any #1898 fix must preserve: dropping a SignalFd must not leave its signal monopolized away from the global handler, and owning one signal must not suppress delivery of an unrelated signal.

zacheryasc:
Final coverage implemented. The reasoning for the change in signal test behavior:
- directly and deterministically show we do not contend for the same signal
- and we do not consume unrelated signals
- and we do not deadlock

zacheryasc:
@elBoberido This should provide coverage for the general problem exposed from our previous conversation. This PR and the associated issue is ready for review.

zacheryasc:
@elBoberido ping. This adds deterministic test coverage for race behavior and more when dealing with posix signals. It makes some extant tests more clear and straightforward, and checks for more possible violations of correct behavior.

elBoberido:
@zacheryasc sorry, this totally slipped my attention. Maybe @elfenpiff can take a look at the PR. He is deeper into this part of the code. I once had a brief look at the code in `signal.rs` and came to the conclusion that one could remove quite some code and potentially also get rid of `LAST_SIGNAL`, which opens the door to not have to use the mutex and run into deadlocks in a single threaded signal execution.

zacheryasc:
> @zacheryasc sorry, this totally slipped my attention. Maybe @elfenpiff can take a look at the PR. He is deeper into this part of the code. I once had a brief look at the code in `signal.rs` and came to the conclusion that one could remove quite some code and potentially also get rid of `LAST_SIGNAL`, which opens the door to not have to use the mutex and run into deadlocks in a single threaded signal execution.

The issue, speaking broadly, is that posix signals are just very uncomfortable for concurrency, and twist your arm into building a singleton-router-ish signal handler. So we might solve single thread in such a way, but mutlithread is a different beast. I didn't make any architectural decisions here, though, except implied concurrency via the test construction. This is just setting guardrails to make correctness more explicit.

PR #1936, opened 2026-08-22
[#1880] Avoid panic when cleanup finds live node

### LINKED ISSUE #1880: `fatal_panic!` in dead-node cleanup kills unrelated processes during concurrent `NodeBuilder::create()` (`MonitoringCreateCleanerError::InstanceStillAlive`)
### Required information

**Operating system:** `Microsoft Windows [Version 10.0.26200.8875]` (x86_64)

**Rust version:**

```
cargo 1.91.1 (ea2d97820 2025-10-10)
rustc 1.91.1 (ed61e7d7e 2025-11-07)
```

**iceoryx2 version:** `0.9.3` (from crates.io). **Also reproduced on `main`** —
commit `790493c068a99132cb979e19ada5d619b5bd618c` (2026-08-04), with the same program below,
by changing only the dependency line to:
`iceoryx2 = { git = "https://github.com/eclipse-iceoryx/iceoryx2", rev = "790493c0..." }`

**Observed result / behaviour:**
A fraction of processes that concurrently call `NodeBuilder::create()` are terminated by a
library-internal `fatal_panic!`. The terminated process is not the one that broke anything — it
is killed because of a transient state of *another* process that is starting up at the same
time. With `panic = "abort"` this ends the whole process; the caller gets no chance to react.

**Expected result / behaviour:**
`NodeBuilder::create()` either succeeds or returns an `Err`. A benign race between "this node
looks dead" and "this node is alive again" should not terminate the calling process.

**Conditions where it occurred / did not occur:**

- Occurs on Windows with 16 processes creating a `Node` simultaneously (see reproduction).
- **No dead nodes are required** — a freshly cleaned system reproduces it.
- Does not occur at concurrency 1.
- Did not reproduce on Linux (see *Scope*).

### Minimal reproduction

`Cargo.toml`:

```toml
[package]
name = "iox2_node_race_repro"
version = "0.1.0"
edition = "2021"

[dependencies]
iceoryx2 = "0.9.3"

[profile.release]
panic = "abort"
```

`src/main.rs`:

```rust
use std::process::{Command, Stdio};
use std::time::{Duration, SystemTime, UNIX_EPOCH};

use iceoryx2::prelude::*;

fn now_millis() -> u128 {
    SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_millis()
}

/// The whole child: create one Node, report, exit.
fn child(start_at_millis: u128) -> ! {
    while now_millis() < start_at_millis {
        std::hint::spin_loop();
    }
    // Drop the Node normally. `std::process::exit()` skips destructors, which would
    // leave this process' own node behind and quietly seed dead nodes.
    let code = match NodeBuilder::new().create::<ipc::Service>() {
        Ok(node) => { println!("OK"); drop(node); 0 }
        Err(e)   => { println!("ERR {e:?}"); 20 }   // recoverable, not the bug
    };
    std::process::exit(code);
}

fn main() {
    let args: Vec<String> = std::env::args().collect();
    if args.len() >= 3 && args[1] == "child" {
        child(args[2].parse().unwrap());
    }

    let concurrency: usize = args.get(1).and_then(|s| s.parse().ok()).unwrap_or(16);
    let rounds: usize = args.get(2).and_then(|s| s.parse().ok()).unwrap_or(40);

    let exe = std::env::current_exe().unwrap();
    let (mut total, mut ok, mut recoverable, mut killed) = (0u32, 0u32, 0u32, 0u32);
    let mut codes = std::collections::BTreeMap::new();

    for _ in 0..rounds {
        let start_at = now_millis() + 1500;   // all children spin until this instant
        let kids: Vec<_> = (0..concurrency)
            // stdout is silenced, but stderr is deliberately inherited so that the
            // library's panic message is visible when you run this.
            .map(|_| Command::new(&exe).arg("child").arg(start_at.to_string())
                     .stdout(Stdio::null()).spawn().unwrap())
            .collect();
        for mut k in kids {
            total += 1;
            match k.wait().unwrap().code() {
                Some(0)  => ok += 1,
                Some(20) => recoverable += 1,
                other    => { killed += 1; *codes.entry(format!("{other:?}")).or_insert(0u32) += 1; }
            }
        }
        std::thread::sleep(Duration::from_millis(50));
    }

    println!("total={total} ok={ok} returned_err={recoverable} killed={killed}");
    println!("exit codes of killed processes: {codes:?}");
}
```

```sh
cargo run --release          # 16 concurrent processes, 40 rounds
```

### Observed

Killed children print the following (output of the program above at default log level; paths
and the `NodeDetails` dump are shortened here for readability):

```
| This should never happen! Unable to acquire monitor cleaner since the Node is still alive.

thread 'main' (22376) panicked at
C:\Users\...\.cargo\registry\src\index.crates.io-.../iceoryx2-0.9.3\src\node\mod.rs:773:17:
From: DeadNodeView(AliveNodeView { id: UniqueNodeId(...), details: Some(NodeDetails { ... }) })
::: This should never happen! Unable to acquire monitor cleaner since the Node is still alive.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Summary lines across several runs on one machine:

```
total=640 ok=338 returned_err=0 killed=302     # 47.2 %
total=640 ok=387 returned_err=0 killed=253     # 39.5 %
total=640 ok=395 returned_err=0 killed=245     # 38.3 %
total=192 ok=121 returned_err=0 killed=71      # 37.0 %
total=640 ok=414 returned_err=0 killed=226     # 35.3 %
exit codes of killed processes: {"Some(-1073740791)": 226}
```

- **35 %-47 %** of processes are terminated across the five runs above, all on one machine.
  Under additional system load we once measured 29 %. Please treat it as a range; the rate is
  clearly load-sensitive, but the failure occurred in every run.
- Exit code `-1073740791` = `0xC0000409`, the Rust `abort` fast-fail path on Windows.
- **`returned_err = 0` in every run** — on this path the library never yields a recoverable
  error. Process termination is the only outcome the caller can observe.
- With `panic = "unwind"` the processes still die (106-112 of 320 in two runs), only with exit
  code `101` instead. Changing the panic strategy does not make this survivable.
- **On `main` @ `790493c` the same program gives `total=640 ok=466 returned_err=0 killed=174`
  (27.2 %)**, same exit code, same panic — so this is not something that has already been fixed
  on `main`. There the panic reports `...\iceoryx2\src\node\mod.rs:753:17` (the same code,
  at a different line than in 0.9.3).

**Trace-level log.** The program above runs at the default log level. Adding
`set_log_level(LogLevel::Trace);` as the first line of `child()` yields a full trace. The tail of
a process that is about to be killed looks like this (paths shortened):

```
62 [T] File { path: Some(FilePath { ... "C:\Temp\iceoryx2\nodes\iox2_1026961663275891200546741401191120.node_monitor_context" }), ... }
| closed

thread 'main' (23088) panicked at ...\iceoryx2-0.9.3\src\node\mod.rs:773:17:
From: DeadNodeView(AliveNodeView { id: UniqueNodeId(UniqueSystemId { value: 1026882442427510961434985072973984, pid: 22688, ... }) ... })
::: This should never happen! Unable to acquire monitor cleaner since the Node is still alive.
```

A full trace is ~8k lines for a 16x4 run and is mostly per-file open/close records; happy to
attach it if that is useful.

### Root cause (as far as we could read it, line numbers for 0.9.3)

```
node/mod.rs   cleanup_dead_nodes / remove_stale_resources
  1. a node is classified as Dead
  2. acquire_cleaner_lock() re-reads the state:
       iceoryx2-bb/posix/src/process_state.rs:1111   state == Alive        -> Err(ProcessIsStillAlive)
       iceoryx2-bb/posix/src/process_state.rs:1183   write lock still held -> Err(ProcessIsStillAlive)
  3. iceoryx2-cal/src/monitoring/file_lock.rs:330 -> MonitoringCreateCleanerError::InstanceStillAlive
  4. iceoryx2/src/node/mod.rs:773                 -> fatal_panic!("This should never happen! ...")
```

Steps 1 and 2 are not atomic, so a node that is starting up concurrently can legitimately be
`Dead` at step 1 and `Alive` at step 2.

Two things suggest this outcome is expected to be benign everywhere except here:

- The lower layer returns it as an ordinary recoverable error (`fail!` -> `Err`), and
  `ProcessState::Starting` exists in the state machine (`process_state.rs:174`).

- **The CAL layer's own documentation example treats it as benign**
  (`iceoryx2-cal/src/monitoring/mod.rs:56-59`):

  ```rust
  Err(MonitoringCreateCleanerError::InstanceStillAlive) => {
      // whoopsie, the monitored instance is not dead
      return;
  }
  ```

### Impact

The panicking process did nothing wrong; it merely started at the same time as another process.
For a library that is statically linked into a larger application, this terminates that whole
application. The caller cannot intercept it.

This looks like a concrete, quantified instance of #763 ("one process (causer) breaks the
contract, and this is detected by another process (detector) ... can cause a violation of
freedom-of-interference").

### Suggested fix

Treating `InstanceStillAlive` at `node/mod.rs:773` as a recoverable outcome rather than an
impossible one would remove the process kill without weakening cleanup: the node is alive, so
there is nothing to clean up.

The caller already tolerates this — `Node::blocking_cleanup_dead_nodes` (`node/mod.rs:1210`)
matches on the result and only increments `failed_cleanups` before continuing, so returning
`Err` here is a drop-in change.

One detail we are not sure about and would leave to you: of the neighbouring arms,
`DoesNotExist` semantics ("nothing to do here, move on") seem closer than
`AlreadyOwnedByAnotherInstance`, because the latter would make `blocking_cleanup_dead_nodes`
keep retrying against a node that is simply alive until the timeout expires.

### Workarounds we found

- Setting `global.node.cleanup-dead-nodes-on-creation`, `-on-destruction` and
  `global.service.cleanup-dead-nodes-on-open` to `false` in the processes that must not die
  makes this path unreachable for them. In a C++/FFI harness of ours (not the program above)
  this took a 16-way concurrent run from 116/240 killed down to 0/240.
- That has a cost: stale resources then accumulate. We also saw re-opening a service name from
  a crashed process stop making progress, but we have not localised it and it is a separate
  matter - we will file that on its own once we can reproduce it minimally.
- Concentrating cleanup in a single dedicated process reduces exposure but does not remove it —
  that process can still be killed by the same panic.

### Scope and limitations

- Reproduced on **Windows** with iceoryx2 0.9.3 from crates.io **and with `main` @ `790493c`**,
  using the same program in both cases (only the dependency line differs).
- The same program on **Linux** (WSL2, kernel 6.6, Rust 1.97.1) did **not** reproduce: 0 kills
  in 960 processes. We did check that the cleanup path is exercised there — a seeded dead node
  was reclaimed during the concurrent run — so this is not simply "cleanup never ran". Still,
  it is one kernel under WSL2, so we would not conclude more than "we could not reproduce it
  on Linux".
- All numbers come from a **single machine**, one Windows build, one MSVC/Rust version.
- Line numbers above are for 0.9.3. On `main` @ `790493c` the same code sits at
  `node/mod.rs:753`.

xiao-yang25:
## Notes for Reviewer

During concurrent node creation on Windows, a starting node can temporarily be classified as dead. When stale-resource cleanup rechecks the node state, the monitoring layer correctly returns `MonitoringCreateCleanerError::InstanceStillAlive`.

This result was previously handled with `fatal_panic!`, terminating the unrelated process performing the cleanup.

This PR:

- maps `InstanceStillAlive` to `NodeCleanupFailure::ResourcesAlreadyCleanedUp`;
- prevents cleanup from continuing against a live node;
- avoids adding a new public error variant or changing the existing ABI;
- updates the Rust, C, and C++ error documentation;
- adds a deterministic conformance test covering IPC, local, and their thread-safe variants;
- adds the fix to the unreleased changelog.

Windows reproducer before the fix:

```text
total=64 ok=36 returned_err=0 killed=28
```

After the fix:

```text
total=64 ok=64 returned_err=0 killed=0
```

The same reproducer was also exercised on macOS/aarch64, Linux/aarch64 (Jetson Orin), and Linux/x86_64. Across 5,760 concurrent node creations, no processes were killed.

### Validation

- `cargo fmt --all -- --check`
- Clippy with warnings denied
- Conformance test: 4 passed, 0 failed
- Windows concurrent reproducer: 64 passed, 0 killed

- [x] Add sensible notes for the reviewer
- [x] PR title is short, expressive and meaningful
- [x] Relevant issues are linked in the References section
- [x] Branch follows the naming format
- [x] Commit message contains the issue ID
- [x] Tests cover the new behavior
- [x] Changelog updated

## References

Closes #1880

### DIFF
--- a/doc/release-notes/iceoryx2-unreleased.md
+++ b/doc/release-notes/iceoryx2-unreleased.md
@@ -59,6 +59,7 @@
 * [#1868](https://github.com/eclipse-iceoryx/iceoryx2/issues/1868) Automatically create windows platform directories
 * [#1872](https://github.com/eclipse-iceoryx/iceoryx2/issues/1872) Register FileDescriptor pyclass with the python module
 * [#1878](https://github.com/eclipse-iceoryx/iceoryx2/issues/1878) Fix race in `pthread_create` on macOS and Windows
+* [#1880](https://github.com/eclipse-iceoryx/iceoryx2/issues/1880) Avoid terminating processes when a node is found alive during stale resource cleanup.
 * [#1893](https://github.com/eclipse-iceoryx/iceoryx2/issues/1893) Fix C language `ipc` and `local` mapping for payload types
 * [#1906](https://github.com/eclipse-iceoryx/iceoryx2/issues/1906) Do not set the exec bit for created resources
 * [#1918](https://github.com/eclipse-iceoryx/iceoryx2/issues/1917) Fix wrong `CLOCK_MONOTONIC` constant on macOS (1 instead of 6) which broke `Time::now_with_clock(ClockType::Monotonic)`
--- a/iceoryx2-cxx/include/iox2/node_failure_enums.hpp
+++ b/iceoryx2-cxx/include/iox2/node_failure_enums.hpp
@@ -63,7 +63,7 @@ enum class NodeCleanupFailure : uint8_t {
     InsufficientPermissions,
     /// Trying to cleanup resources from a [`Node`] node which was using a different iceoryx2 version.
     VersionMismatch,
-    /// Another instance has successfully cleaned up all resources.
+    /// There are no stale resources to remove.
     ResourcesAlreadyCleanedUp,
     /// Another instance has acquired the ownership of all resources and is currently cleaning up.
     AnotherInstanceIsCleaningUpTheNode,
--- a/iceoryx2-ffi/c/src/api/node.rs
+++ b/iceoryx2-ffi/c/src/api/node.rs
@@ -121,7 +121,7 @@ pub enum iox2_node_cleanup_failure_e {
     INSUFFICIENT_PERMISSIONS,
     /// Trying to cleanup resources from a dead node which was using a different iceoryx2 version.
     VERSION_MISMATCH,
-    /// Another instance has successfully cleaned up all resources.
+    /// There are no stale resources to remove.
     RESOURCES_ALREADY_CLEANED_UP,
     /// Another instance has acquired the ownership of all resources and is currently cleaning up.
     ANOTHER_INSTANCE_IS_CLEANING_UP_THE_NODE,
--- a/iceoryx2/conformance-tests/src/node.rs
+++ b/iceoryx2/conformance-tests/src/node.rs
@@ -23,7 +23,7 @@ pub mod node {
 
     use iceoryx2::config::Config;
     use iceoryx2::node::{
-        NodeCleanupFailure, NodeCreationFailure, NodeListFailure, NodeState, NodeView,
+        DeadNodeView, NodeCleanupFailure, NodeCreationFailure, NodeListFailure, NodeState, NodeView,
     };
     use iceoryx2::prelude::*;
     use iceoryx2::service::Service;
@@ -387,6 +387,48 @@ pub mod node {
         } else {
             test_fail!("Process internal nodes shall be always detected as alive.");
         }
+    }
+
+    #[conformance_test]
+    pub fn removing_stale_resources_of_live_node_does_not_remove_node<S: Service>() {
+        let test = Test::<S>::new();
+        let node = NodeBuilder::new()
+            .config(test.config())
+            .create::<S>()
+            .unwrap();
+
+        let mut node_details = None;
+        Node::<S>::list(node.config(), |node_state| {
+            if let NodeState::Alive(node_view) = node_state
+                && node_view.id() == node.id()
+            {
+                node_details = node_view.details().clone();
+            }
+            CallbackProgression::Continue
+        })
+        .unwrap();
+
+        let result = DeadNodeView::<S>::__internal_try_remove_stale_resources(
+            *node.id(),
+            node_details.unwrap(),
+        );
+        assert_that!(
+            result.err(),
+            eq Some(NodeCleanupFailure::ResourcesAlreadyCleanedUp)
+        );
+
+        let mut node_is_alive = false;
+        Node::<S>::list(node.config(), |node_state| {
+            if let NodeState::Alive(node_view) = node_state
+                && node_view.id() == node.id()
+            {
+                node_is_alive = true;
+            }
+            CallbackProgression::Continue
+        })
+        .unwrap();
+
+        assert_that!(node_is_alive, eq true);
     }
 
     #[conformance_test]
--- a/iceoryx2/src/node/mod.rs
+++ b/iceoryx2/src/node/mod.rs
@@ -275,7 +275,7 @@ pub enum NodeCleanupFailure {
     InsufficientPermissions,
     /// Trying to cleanup resources from a dead [`Node`] which was using a different iceoryx2 version.
     VersionMismatch,
-    /// Another instance has successfully cleaned up all resources.
+    /// There are no stale resources to remove.
     ResourcesAlreadyCleanedUp,
     /// Another instance has acquired the ownership of all resources and is currently cleaning up.
     AnotherInstanceIsCleaningUpTheNode,
@@ -750,8 +750,8 @@ impl<Service: service::Service> DeadNodeView<Service> {
                     "{} due to an internal error while acquiring monitoring cleaner.", msg);
             }
             Err(MonitoringCreateCleanerError::InstanceStillAlive) => {
-                fatal_panic!(from self,
-                        "This should never happen! {} since the Node is still alive.", msg);
+                fail!(from self, with NodeCleanupFailure::ResourcesAlreadyCleanedUp,
+                    "{} since the Node is still alive and has no stale resources.", msg);
             }
         }
     }


xiao-yang25:
The remaining CI failures appear to be unrelated to this PR:

- `nightly-check` fails because the latest Nightly workflow run is currently failing.
- The `unstable` job fails while compiling the third-party `zenoh 1.9.0` dependency with `rustc 1.100.0-nightly`. The Zenoh jobs using stable and Rust 1.89 pass on all tested platforms.

The regular `main` build does not run this `unstable` job because it is explicitly skipped for `refs/heads/main`. Since this PR does not modify the Zenoh integration or its dependencies, I have not changed the PR to work around the failure.

elBoberido:
@xiao-yang25 yes, the failures seem unrelated. There is a flaky test on Windows which we need to have a closer look at some time. I re-triggered the nightly, which suffered from the same flaky test on Windows.

The cleanup mechanism is not easy to get right and @elfenpiff is the final authority to tell whether this bugfix does not violate invariants. From my point of view, it should be okay and I guess we just overlooked the different behavior on Windows.

elBoberido:
@xiao-yang25 the `unstable` build is not mandatory for the CI, so it is not a blocker. It seems to be an issue in Zenoh and it also occurs when the Zenoh dependency is updated to 1.10.

The Rust team is currently experimenting with a new trait resolver on the nightly Rust compiler and that might cause the issue. If it persists, we will temporarily disable the Zenoh tunnle build on the unstable builds.

xiao-yang25:
@elBoberido Thanks for the clarification and for re-triggering the nightly. I will leave the unrelated CI failures as they are and wait for @elfenpiff's review regarding the cleanup invariants.

elBoberido:
It seems the issue with the nighty compiler fixed itself over the weekend. We now just need to wait for @elfenpiff :)

xiao-yang25:
Looks like nightly just needed the weekend off 😄 I’ll wait for @elfenpiff’s verdict.

elfenpiff:
I am a bit hesitant to approve the PR. The reason I introduced a `fatal_panic` here was that when the node is detected as dead and, in the cleanup, as alive, we have a race in our implementation in `iceoryx2-bb/posix/src/process_state.rs:982` in `state()`.

It should, in theory, cleanly distinguish between in set up, alive, dead, and cleanup. Somehow, it categorized in set up falsly as dead, then the cleanup branch is triggered and the next state check identifies the node as alive.

The question is now: is this some weird Windows sync issue (and Windows has a lot of weirdness inside its stack), and we need to add something to the Windows platform layer, or is it a bug in Iceoryx2 in `process_state.rs:982,` and we need to fix it there? I will investigate this ...

elfenpiff:
So I studied the implementation and the windows documentation. The problem seems to be that the initialization realized via file permissions can be insufficiently synchronized in Windows. In the documentation of `FindNextFileW` is a remark for the `GetFileInformationByHandle` function (who knows why, it is in the docs of an unrelated function)

```
Note:  In rare cases or on a heavily loaded system, file attribute information on NTFS file systems may not be current at the time this function is called. To be assured of getting the current NTFS file system file attributes, call the [GetFileInformationByHandle](https://learn.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-getfileinformationbyhandle) function.
```

But we cannot use this function here, since we do not have a handle - it is a foreign process and we need to acquire the file information by path, which could lead to a misclassification in the `ProcessState::state()` function. So despite that the file permissions are set correctly, the other process sees a different version, misclassifies the process state and the cleanup path is used.

It seems that the second call to the resource cleanup established the synchronization and therefore caused a correct `fatal_panic,` but the bigger issue would be: what would happen if the misclassification still happens since the process state is still misclassified because of the sync issue? Then the process would remove the node by force from the service and with all of its files. This could actually corrupt the system.

I will talk tomorrow to @elBoberido after a good night's sleep, and let's see what we can come up with. If we get this wrong, and my suspicion is correct, such a fix could hide the bigger, more dangerous problem.

xiao-yang25:
Thanks for the detailed investigation. Your concern makes sense: my change only handles the case where the second check observes the node as alive. If the state is misclassified repeatedly, cleanup could still proceed against a live node, so the change could indeed hide the more serious underlying issue.

One observation from the Windows PAL implementation: `fstat` appears to already have an open file handle, but permission retrieval is still performed by path through `GetFileSecurityA`. Microsoft documents [`[GetSecurityInfo](https://learn.microsoft.com/en-us/windows/win32/api/aclapi/nf-aclapi-getsecurityinfo)`](https://learn.microsoft.com/en-us/windows/win32/api/aclapi/nf-aclapi-getsecurityinfo) as retrieving an object’s security descriptor by handle. Would it be worth evaluating here, or is there an access or lifetime constraint I’m overlooking?

I’m happy to help test a revised approach on Windows.

xiao-yang25:
I ran several local experiments to distinguish delayed DACL visibility from a broader lifecycle race.

### Observations

In a controlled test where the initialization files remained unchanged for two seconds, the monitor observed `Starting` 5,793 times and never reported `Dead`.

I also ran an isolated cross-process DACL test with 10,000 permission transitions and 20,000 reads. After explicit synchronization, it observed no stale values, path/handle mismatches, or API errors. This does not rule out every narrow race, but I could not reproduce a simple delayed DACL update.

A separate stress test repeatedly created and removed a real `ProcessGuard` 3,000 times using the same path. This reproduced one transient `Dead` state. During the test, the Windows PAL frequently reported failures from `GetFileSecurityA` and occasionally from `GetFinalPathNameByHandleA`.

The current Windows `fstat()` implementation converts an existing handle back into a path before reading its DACL. If the file disappears or is recreated during that sequence, the permission lookup can fail. However, `fstat()` still returns success with no permission bits set, allowing `ProcessState::state()` to continue with incomplete information.

This appears to be a concrete path-based TOCTOU issue combined with a swallowed permission-read error.

### Prototype

I prototyped the following changes:

- Read permissions directly from the existing handle using [`[GetSecurityInfo](https://learn.microsoft.com/en-us/windows/win32/api/aclapi/nf-aclapi-getsecurityinfo)`](https://learn.microsoft.com/en-us/windows/win32/api/aclapi/nf-aclapi-getsecurityinfo).
- Return an error from `fstat()` when the DACL cannot be retrieved.
- Use [`[SetSecurityInfo](https://learn.microsoft.com/en-us/windows/win32/api/aclapi/nf-aclapi-setsecurityinfo)`](https://learn.microsoft.com/en-us/windows/win32/api/aclapi/nf-aclapi-setsecurityinfo) for handle-based permission updates where possible.
- Request `WRITE_DAC` for exclusively created files and protect generated DACLs from inherited directory ACEs.

A deterministic regression confirmed that the previous `fchmod()` also depended on the path: updating permissions through an open descriptor failed after the path was removed. The handle-based implementation succeeds.

With the initial read-side fix, the original 3,000-iteration stress test completed without observing `Dead`.

### Remaining observation

I then ran a stronger 5,000-iteration lifecycle test. It still observed one transient `Dead` state.

One possible explanation is a separate lifecycle race: the monitor may still hold an unlocked handle from the previous lifecycle while a new lifecycle is already starting under the same path. A later state check would then see the new lifecycle as `Alive`.

This suggests there may be two distinct effects:

1. A confirmed path-based permission race in the Windows PAL.
2. A narrower transition involving handles from different lifecycles.

The safety invariant seems to be that cleanup must not continue from incomplete or inconsistent state information. Permission-query failures should be propagated, and a later `Alive` result must always stop cleanup.

With the candidate changes, the complete Windows `iceoryx2-bb-posix` test suite passes: 327 passed, 4 ignored, and 0 failed.

The lifecycle test intentionally reuses the same path very aggressively, so it may not accurately represent the original failure. I do not want to overinterpret that result. Since this touches core process-state and cleanup logic, I think these findings require further careful discussion and validation before we draw a firm conclusion. The current tests may still miss relevant assumptions or Windows-specific behavior.

PR #1957, opened 2026-08-31
[#1910] unique id generator concept

### LINKED ISSUE #1910: UniqueSystemId concept
To allow exchangeable `UniqueSystemId` implementations, a concept shall be added to iceoryx2-cal.

FerdinandSpitzschnueffler:
## Notes for Reviewer

This PR adds a `UniqueIdGenerator` concept to allow different implementations for generating unique IDs.

## PR Reviewer Reminders

* Public API is documented

## References

Closes #1910


iceoryx2-cxx/include/iox2/unique_node_id.hpp
```diff
@@ -76,6 +80,20 @@ class UniqueNodeId {
     iox2_unique_node_id_h m_handle = nullptr;
 };
 
+template <ServiceType T>
+auto UniqueNodeId::pid() const -> int32_t {
```
elfenpiff, line 84: Please add the `inline` keyword.

iceoryx2-cxx/include/iox2/unique_node_id.hpp
```diff
@@ -76,6 +80,20 @@ class UniqueNodeId {
...
+    return iox2_unique_node_id_pid(&m_handle, iox2::bb::into<iox2_service_type_e>(T));
+}
+
+template <ServiceType T>
+auto UniqueNodeId::creation_time() const -> timespec {
```
elfenpiff, line 89: Please add the `inline` keyword.

iceoryx2/src/node/mod.rs
```diff
@@ -1548,32 +1548,22 @@ impl NodeBuilder {
...
-            Err(e) => {
-                fail!(from "NodeBuilder::create()",
-                    with NodeCreationFailure::InternalError,
-                    "Unable to create node since the global management segment could not be opened. [{e:?}]");
-            }
-        };
 
-        unsafe { self.__internal_create_with_custom_node_id(UniqueNodeId::new(node_counter)) }
-    }
+        let name = fatal_panic!(
```
elfenpiff, line 1558: To avoid this `fatal_panic!` I would recommend to adjust some code pieces.

1. The `Some(n)` path could be rewritten as `Some(n) => n.clone()` I think `n` is already of type node name and it should implement clone or even copy.
2. We might need an empty node name even in other places, therefore we could add this to the node name implementation:

```rust
impl NodeName {
  pub const fn new_unchecked(name: &str) {}

  const EMPTY: NodeName = unsafe { NodeName::new_unchecked(""); };
}
```

Then the whole code snippet could be rewritten as:

```rust
let name = match &self.name {
  Some(n) => *n,
  None => NodeName::EMPTY,
};
```

iceoryx2/src/unique_id_generator/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    config::Config,
+    node::node_name::NodeName,
+    port::port_name::PortName,
+    service::{self, service_name::ServiceName},
+};
+
+#[doc(hidden)]
```
elfenpiff, line 27: Since we only have one implementation, we do not need a recommended thing here. Just use the only thing available here.

The recommended approach makes sense, in my opinion, when we vary with implementations to choose the best fit for a target - a use case we do not have here.

iceoryx2/src/unique_id_generator/recommended.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) 2026 Contributors to the Eclipse Foundation
```
elfenpiff, line 1: Please remove this file.

iceoryx2/src/unique_id_generator/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+impl core::fmt::Display for UniqueIdGeneratorError {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(f, "UniqueIdGeneratorError::{self:?}")
+    }
+}
+
+impl core::error::Error for UniqueIdGeneratorError {}
+
+/// Generates [`UniqueId`]s whose [`UniqueId::unique_value()`]s are unique, at least within a single process.
+pub trait UniqueIdGenerator: From<UniqueId> {
```
elfenpiff, line 92: Why does the `UniqueIdGenerator` needs to implement `From<UniqueId>`. I thought it generates `UniqueId`s and is not a thing that is convertible from a `UniqueId` - maybe I am just a bit confused.

iceoryx2/src/identifiers.rs
```diff
@@ -145,6 +149,52 @@ impl UniquePortId {
...
+pub struct UniqueServiceId(pub(crate) UniqueId);
+
+impl core::fmt::Display for UniqueServiceId {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(f, "{:x}", self.0.value())
+    }
+}
+
+impl UniqueServiceId {
+    pub(crate) fn new<Service: crate::service::Service>(entity: Entity, config: &Config) -> Self {
```
elfenpiff, line 176: Would it be possible to reduce the error handling here and instead handle this with the signature. So we could introduce a method like:

```rust
impl UniqueServiceId {
  pub(crate) fn from_publish_subscribe_service(name: &ServiceName, config: &Config) -> Self {
    Service::UniqueId::generate::<Service>(Entity::PubSubService(name), config)
  }
}
```

So the user can call only `from_{***}_service` - hard to use wrong. We get rid of an error case which we also do not have to test and the implementation is simpler. What do you think?

iceoryx2/src/node/mod.rs
FerdinandSpitzschnueffler: I derived `Default` for `NodeName` - this creates an empty string and we need neither an unsafe method nor error handling.

iceoryx2/src/unique_id_generator/mod.rs
FerdinandSpitzschnueffler: I used this for the `pid()` and `creation_time()` methods of the `UniqueNodeId`. I refactored the related methods in the `UniqueIdGenerator` trait to get rid of the `From<UniqueId>`.

iceoryx2-bb/posix/src/unique_system_id.rs
```diff
@@ -72,9 +72,9 @@ enum_gen! { UniqueSystemIdCreationError
 )]
 #[repr(C)]
 pub struct UniqueSystemId {
-    pid: u32,
     seconds: u32,
     nanoseconds: u32,
+    pid: u32,
     counter: u32,
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: Changed the unique value calculation in `UniqueId` and added a test.

iceoryx2-ffi/c/src/api/unique_client_id.rs
```diff
@@ -21,7 +21,7 @@ use super::{AssertNonNullHandle, HandleToType};
 
 /// The system-wide unique id of a `iox2_client_t`.
 #[repr(C)]
-#[repr(align(4))] // core::mem::align_of::<UniqueClientId>()
+#[repr(align(8))] // core::mem::align_of::<UniqueClientId>()
 pub struct iox2_unique_client_id_storage_t {
     internal: [u8; 20], // core::mem::size_of::<Option<UniqueClientId>>()
 }
```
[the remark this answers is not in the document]
FerdinandSpitzschnueffler: The size changes were incorrect and were reversed.

iceoryx2/conformance-tests/src/unique_id_generator_trait.rs
```diff
@@ -0,0 +1,106 @@
...
+    impl UniqueIdGenerator for TestUniqueId {
+        fn generate<Service: service::Service>(
+            _entity: Entity,
+            _config: &Config,
+        ) -> Result<UniqueId, UniqueIdGeneratorError> {
+            Ok(unsafe { UniqueId::from_raw_id(TestUniqueId::new().value() as u128) })
+        }
+    }
+
+    #[test]
```
elfenpiff: All tests annotated with `#[test]` and not `#[conformance_test]` are normal unit tests. Those should be located under `iceoryx2/tests`, not here. Otherwise, we mix test variants and those tests are maybe never executed or always executed for different concepts (I do not know exactly how it behaves).

When they are moved we can merge this.

iceoryx2/src/unique_id_generator/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+
+    /// Returns the unique part of the ID.
+    pub fn unique_value(&self) -> u64 {
+        self.unique_value
+    }
+}
+
+/// Describes failures related to the [`UniqueIdGenerator`] trait.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum UniqueIdGeneratorError {
```
elfenpiff: Those should be two separate enums.

As it seems `UniqueIdGenerator::generate()` never returns `NotImplemented` and `pid` and `creation_time` never return `GenerationError`.
When they are combined together the user might suspect that both variants can always be returned by every method and then creates error handling and untestable code paths.

If we split the error enum up into `UniqueIdGeneratorGenerateError` for `generate` and `UniqueIdGeneratorDetailsError` for `pid` and `creation_time` we solve this issue.

iceoryx2/src/unique_id_generator/unique_system_id.rs
```diff
@@ -0,0 +1,60 @@
...
+    fn generate<Service: service::Service>(
+        entity: Entity,
+        config: &Config,
+    ) -> Result<UniqueId, UniqueIdGeneratorError> {
+        let id = match entity {
+            Entity::Node(_) => {
+                let node_counter = match GlobalManagementSegment::<Service>::open_or_create(config)
+                {
+                    Ok(mgmt) => mgmt.increment_node_counter(),
+                    Err(e) => {
```
elfenpiff, line 38: Just to be sure here. There is no interrupt signal error code in here?

Those errors always need to be forwarded to the end user API. If there is no such interrupt error signal case here, the code can stay as it is. If there is an interrupt signal error case, please introduce another enum value to propagate this to the end user.

This also needs to be propagated on all places where this class is used.

iceoryx2/tests-common/src/unique_system_id_tests.rs
```diff
@@ -0,0 +1,26 @@
...
+// https://www.apache.org/licenses/LICENSE-2.0, or the MIT license
+// which is available at https://opensource.org/licenses/MIT.
+//
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use iceoryx2::unique_id_generator::UniqueId;
+use iceoryx2_bb_posix::unique_system_id::UniqueSystemId;
+use iceoryx2_bb_testing::assert_that;
+use iceoryx2_bb_testing_macros::test;
+
```
elfenpiff: Here the normal unit test cases from the conformance test suite can be stored.

iceoryx2/src/identifiers.rs
```diff
@@ -169,9 +252,9 @@ impl core::fmt::Display for UniqueNodeId {
 }
 
 impl UniqueNodeId {
-    pub(crate) fn new(counter: u32) -> Self {
-        Self(fatal_panic!(from "UniqueNodeId::new",
-                when UniqueSystemId::from_counter(counter),
+    pub(crate) fn new<Service: crate::service::Service>(name: NodeName, config: &Config) -> Self {
+        Self(fatal_panic!(from "UniqueNodeId::new()",
+                when Service::UniqueId::generate::<Service>(Entity::Node(name), config),
```
elfenpiff, line 268: Please do not use `fatal_panic` in this function or in any other. Those `fatal_panic` or `expect` calls do call the end user application instantly and the end user application has no way of preventing this. There is no method to ask, if `pid` or `creation_time` is supported or not.

But we have a clear `Result` and error enum path where this is cleanly communicated.

`fatal_panic` should only be used if the next call would cause undefined behavior or any other guaranteed kind of malfunction.

`expect` should only be used if there is an algorithmic invariant that we can proof. But in this case it is absolutely supported to use a `UniqueId` that does not provide `pid`, `creation_time` or so but it would kill the process instantly.

iceoryx2/src/identifiers.rs
```diff
@@ -145,6 +150,84 @@ impl UniquePortId {
...
+    }
+}
+
+impl UniqueServiceId {
+    pub(crate) fn from_publish_subscribe_service<Service: crate::service::Service>(
+        name: &ServiceName,
+        config: &Config,
+    ) -> Self {
+        Self(
+            fatal_panic!(from "UniqueSystemId", when Service::UniqueId::generate::<Service>(
```
elfenpiff: Please do not use `fatal_panic` in any of those methods. Cleanly report and communicate the error with `Result` and an error enum.

elfenpiff (changes requested):

iceoryx2/src/unique_id_generator/unique_system_id.rs
FerdinandSpitzschnueffler: `GlobalManagementSegment::open_or_create()` does not return an interrupt signal error code, just forwards the error of the underlying `open_or_create()` call which comes from the dynamic storage - there I didn't find such an error code as well. So, either there is none or it's already lost in the layers below.

elfenpiff (dismissed):

elfenpiff (approved):

